PR #44, opened 2016-11-01
Jserrors

dlucas1:
- Fixes favicon.ico not found error with update to index.html for the "shortcut icon"
- Fixes javascript console errors which allows for full angular compatibility
- Since I verified that angular 1.4.14 (latest version) now works, I updated the bower.json to reflect this.

- Tested the CMS widget and list of products/developers/edition/version:
a. Adding Cert IDs populates CMS EHR Certification ID as expected and updates criteria
b. Adding multiple Cert IDs populates CMR EHR Certification ID as expected and updates criteria
c. Removing a single Cert ID from the widget updates the widget and criteria
d. Selecting 'Remove all products' in the widget removes all products and displays "No products selected", disables the 'Remove all products' button, and updates Base Criteria to 0%
e. The list of products/developer/edition/Version shows "No result found" if the API key is invalid in the search_options API call header
f. Compared all changes on local vs dev to verify that there are no behavior changes are found
g. Reviewed console throughout all testing to ensure that no javascript errors occur


app/index.html
```diff
@@ -24,6 +24,7 @@
     <link rel="stylesheet" href="product/product.css">
     <link rel="stylesheet" href="search/search.css">
     <link rel="stylesheet" href="search/certid.css">
+    <link rel="shortcut icon" href="">
```
andlar: I don't think we want to obscure the fact that we don't have a favicon. I think the better option is to actually get one, deploy it, and let the system pick it up. For that to work I don't want to band-aid over that issue

app/search/certid.js
```diff
@@ -35,9 +35,14 @@ var chplCertIdWidget = (function(){
...
-			prods.forEach(function(item,index) {
-				productIds.push(item.productId);
-			});
+			try{
+				prods.forEach(function(item,index) {
+					productIds.push(item.productId);
+				});
+			}
+			catch(e) {
+				// do nothing
```
andlar: You need to do _something_ with the error, even if it's just logging it

app/search/certid.js
```diff
@@ -137,7 +142,7 @@ var chplCertIdWidget = (function(){
 		updateButtonAndId: function (showIdRequested) {
 			var data = null;
 			var dataText = chplCertIdWidget.getCertificationIdData();
-			if ((null !== dataText) && (dataText.length > 0)) {
+			if ((dataText != "null") && (dataText.length > 0)) {
```
andlar: `"null" !== null`. One is a string, the other is null. As well, Javascript comparators should use the three-character comparators (`===` or `!==`)

app/search/certid.js
```diff
@@ -216,16 +221,12 @@ var chplCertIdWidget = (function(){
 		/////////////////////////////////////////////////////////////////////////////////////////////////////
 		getProductsInCart: function () {
 			var prods = [];
-			try {
 				var data = chplCertIdWidget.getCertificationIdData();
-				if (null === data || "undefined" === data || "" === data) {
-					//console.debug("getProductsInCart: No certification id data to retrieve.");
-				} else {
+				
+				if(data != "null"){
```
andlar: Not `"null"`, but `null`

app/search/searchCtrl.js
```diff
@@ -193,28 +193,30 @@
...
-                        vm.cqms = options.cqmCriterionNumbers;
-                        vm.editions = options.editions;
-                        vm.practices = options.practiceTypeNames;
-                        vm.certBodies = options.certBodyNames;
-                        vm.certificationStatuses = options.certificationStatuses;
-                        for (var i = 0; i < vm.certificationStatuses.length; i++) {
-                            if (vm.certificationStatuses[i].name === 'Pending') {
-                                vm.certificationStatuses.splice(i,1);
-                                break;
+                    	if(!options.hasOwnProperty('error')){
```
andlar: What happens if the service returns an error? Is there any message to the user displayed? If you're checking for something the else path needs to be followed too

bower.json
```diff
@@ -6,30 +6,30 @@
   "license": "https://github.com/chpladmin/chpl-api/blob/staging/LICENSE",
   "private": true,
   "dependencies": {
-    "angular": "1.4.7",
-    "angular-animate": "1.4.7",
+    "angular": "~1.4.8",
```
andlar: In the comments to the PR you said the code supported up to Angular 1.4.12. if so, why is this not requiring 1.4.12?

dlucas1:
I also noticed that the updated angular version causes other javascript errors which breaks some unit tests. These are mostly 'undefined' errors which are straightforward to resolve. These need to be resolved before accepting this PR.

dlucas1:
Closing this PR for now as other priorities popped up.

PR #56, opened 2016-12-01
OCD-1019: added main-content border to muu upload; stacked under CMS Download

dlucas1:


app/admin/components/cms.html
```diff
@@ -1,13 +1,15 @@
...
+  <div class="col-md-8 main-content">
     <h2>Download the latest CMS listing</h2>
     <div>
       <a type="button" class="btn btn-lg btn-ai-success" ng-disabled="!vm.isReady"
          ng-csv="vm.cmsArray" filename="{{ vm.filename }}"
          csv-header="vm.csvHeader" csv-column-order="vm.csvColumnOrder"><i class="fa fa-cloud-download"></i> Download</a>
     </div>
      </div>
-      <div class="col-md-6" ng-if="vm.isChplAdmin || vm.isOncStaff">
+     <div class="row">&nbsp;</div>
```
andlar: One thing I've tried _really_ hard to do is avoid adding empty elements just for the sake of spacing.  Part of that is that empty elements shouldn't exist, part is that it makes the responsiveness of the pages less simple, part of it is that empty elements imply there _should_ be something there, but there isn't. If something needs to be displayed differently it should be done via css. Give the row with the FileUploader in it a particular ID, and then give it a class that targets just that row, and adds some space above. But then check what happens when you resize the view, and pretend you're on a tablet or phone. I know there are other issues with small screen sizes, but we need to avoid adding new ones to that (long) list of challenges

andlar: Oh, and I'm going to approve this PR, because the comment above isn't really a blocker, more of a "best practices" kind of thing

PR #73, opened 2016-12-23
OCD-1049: Handle surveillance deletion

andlar:


andlar:
Running low on time today, so going to merge this and press on

PR #106, opened 2017-03-13
OCD-1251: Display ROLE, allow selection on CREATE, authorize on edit/delete

dlucas1:


RELEASE_NOTES.md
```diff
@@ -1,11 +1,12 @@
 # Release Notes
 
 ## Version TBD
-_Date TBD_
+_13 March 2017_
```
kekey1: I think this update actually goes into a new TBD section rather than added to the March 13 section deployed earlier today.

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -16,12 +16,16 @@
...
+        vm.isChplAdmin = authService.isChplAdmin;
+        vm.isAcbAdmin = authService.isAcbAdmin;
+        vm.isAcbStaff = authService.isAcbStaff;
 
         activate();
 
         ////////////////////////////////////////////////////////////////////
 
         function activate () {
+        	vm.authorities = ['ONC', 'ONC-ACB'];
```
andlar: If this is the list of possible authorities for the user to choose from when picking their role I think this is incorrect. I think the select box should show all of the ROLEs the user has, and require them to choose between them

RELEASE_NOTES.md
```diff
@@ -1,7 +1,13 @@
...
 
 ## Version TBD
-_Date TBD_
+_27 March 2017_
+
+### New features
+* Add "Surveillance initiated by" to Surveillance Activity
+
+## Version 7.3.2
+_13 March 2017_
```
andlar: The stuff you have under 7.3.2 is all of the stuff that should be in the Next Version section; Version 7.3.1 was released on 13 Mar, and the chunk above hasn't been released yet

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -32,6 +36,10 @@
             vm.disableValidation = vm.surveillance.errorMessages && vm.surveillance.errorMessages.length > 0;
             vm.workType = workType;
             vm.showFormErrors = false;
+            if(vm.surveillance.authority === 'ROLE_ADMIN' && !vm.isChplAdmin()){
+            	vm.disableValidation = false;
+            	vm.showFormErrors = true;
+            }
```
andlar: What's this for? Setting `vm.disableValidation` true is for when there were previous errors in the surveillance that we need to allow. Setting `vm.showFormErrors` to true is when the user is about to submit the form, and we want to show them the errors that are wrong. Whether the surveillance is ONC-created or not has nothing to do with that

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -147,48 +155,56 @@
...
-                        }
-                    },function (error) {
-                        if (error.data.errorMessages && error.data.errorMessages.length > 0) {
-                            vm.errorMessages = error.data.errorMessages;
-                        } else {
-                            vm.errorMessages = [error.statusText];
-                        }
-                    });
-            }
+        	if(!vm.surveillance.errorMessages.length > 0){
```
andlar: It is possible for a surveillance with legacy errorMessages to be saved. We can't restrict saving based on the presence of these messages

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -147,48 +155,56 @@
...
+                    $uibModalInstance.close(vm.surveillance);
+                } else if (vm.workType === 'initiate') {
+                    vm.surveillance.certifiedProduct.edition = vm.surveillance.certifiedProduct.certificationEdition.name;
+                    if(vm.surveillance.authority === 'ONC'){
+                    	vm.surveillance.authority = 'ROLE_ADMIN';
+                    }
+                    else if(vm.surveillance.authority === 'ONC-ACB'){
+                    	vm.surveillance.authority = 'ROLE_ACB_STAFF';
+                    }
+                    commonService.initiateSurveillance(vm.surveillance)
```
andlar: This `authority` value should be set in the SELECT box, not in the code

RELEASE_NOTES.md
```diff
@@ -1,7 +1,15 @@
...
-_Date TBD_
+_27 March 2017_
+
+### New features
+* Add "Surveillance initiated by" to Surveillance Activity
+
+---
+
+## Version 7.3.2
+_13 March 2017_
```
andlar: All of the things under Version 7.3.2 are in the "Next release" section. They have not been released. Version 7.3.1 is what was released today

src/app/admin/components/surveillance/editSurveillance.html
```diff
@@ -17,6 +17,14 @@ <h4 class="modal-title" id="editSurveillanceLabel"><span ng-if="vm.workType ===
       <div class="col-sm-3"><strong>Surveillance ID:</strong></div>
       <div class="col-sm-9">{{ vm.surveillance.friendlyId }}</div>
     </div>
+    <div class="row" ng-if="vm.workType !== 'initiate'">
+      <div class="col-sm-3""><strong>Surveillance initiated by:</strong></div>
```
andlar: Double quoted quotes on `col-sm-3`

src/app/admin/components/surveillance/editSurveillance.html
```diff
@@ -17,6 +17,14 @@ <h4 class="modal-title" id="editSurveillanceLabel"><span ng-if="vm.workType ===
...
+      <div class="col-sm-9">
+      	<span>{{ vm.surveillance.authority === 'ROLE_ADMIN' ? 'ONC' : 'ONC-ACB' }}</span>
+      	<span class="text-danger" ng-if="vm.surveillance.authority === 'ROLE_ADMIN' && !vm.isChplAdmin()"></span>
```
andlar: Why is this line here? Nothing is being displayed

src/app/admin/components/surveillance/editSurveillance.html
```diff
@@ -50,6 +58,15 @@ <h4 class="modal-title" id="editSurveillanceLabel"><span ng-if="vm.workType ===
         </select>
         <div class="text-danger" ng-if="(vm.showFormErrors || vm.editForm.surveillanceType.$touched) && vm.editForm.surveillanceType.$error.required">Field is required</div>
       </div>
+    </div> 
+    <div class="form-group" ng-if="vm.workType === 'initiate' && vm.isChplAdmin() && (vm.isAcbAdmin() || vm.isAcbStaff())">
+      <label for="authority" class="col-sm-3">Initiated by <span class="text-danger">*</span></label>
+      <div class="col-sm-9">
+        <select class="input-sm form-control" ng-model="vm.surveillance.authority" id="authority" name="authority"
+                ng-options="authority for authority in vm.authorities" required>
```
andlar: These options should be the options that are the user's role. If the user has `ROLE_ADMIN` and `ROLE_ACB_STAFF` then those two strings should be the options. If the user has `ROLE_ACB_ADMIN` and `ROLE_ACB_STAFF` then those should be the options. If they have all three then there should be three options

src/app/admin/components/surveillance/editSurveillance.html
```diff
@@ -112,14 +130,17 @@ <h4 class="modal-title" id="editSurveillanceLabel"><span ng-if="vm.workType ===
         <li ng-if="vm.editForm.surveillanceType.$error.required">Surveillance Type is required</li>
         <li ng-if="vm.editForm.sitesSurveilled.$error.required">Sites Surveilled is required</li>
         <li ng-if="vm.editForm.sitesSurveilled.$error.number">Sites Surveilled must be a number</li>
+        <li ng-if="vm.editForm.authority.$error.required">Initiated by is required</li>
         <li ng-if="!vm.surveillance.requirements || vm.surveillance.requirements.length === 0">At least one Requirement must be surveilled</li>
+        <li ng-if="vm.surveillance.authority === 'ROLE_ADMIN' && !vm.isChplAdmin()">User must have authority for ONC-Admin to edit surveillance created by ONC-Admin</li>
       </ul>
     </div>
-    <button ng-disabled="(vm.editForm.$invalid || !vm.surveillance.requirements || vm.surveillance.requirements.length ===0) && vm.showFormErrors && !vm.disableValidation"
+    <button ng-disabled="(vm.editForm.$invalid || !vm.surveillance.requirements || vm.surveillance.requirements.length === 0 || vm.surveillance.errorMessages.length > 0) && vm.showFormErrors && !vm.disableValidation"
```
andlar: If a product has errormessages we can't disable the button, because if we do there's no way for the user to fix the errors

src/app/admin/components/surveillance/editSurveillance.html
```diff
@@ -112,14 +130,17 @@ <h4 class="modal-title" id="editSurveillanceLabel"><span ng-if="vm.workType ===
...
             class="btn btn-ai-success pull-right"
             ng-mouseover="vm.showFormErrors = true"
             ng-click="vm.save()"><i class="fa fa-save"></i> Save</button>
-    <button class="btn btn-danger pull-left" ng-if="vm.workType === 'edit'"
+    <button ng-disabled="vm.surveillance.errorMessages.length > 0"
```
andlar: We can't disable a button for deleting when there are errorMessages, though we should if the user is not ROLE_ADMIN but the surveilllance was created by ROLE_ADMIN

src/app/admin/components/surveillance/surveillanceInspect.html
```diff
@@ -36,7 +36,9 @@ <h4 class="modal-title" id="inspectLabel">Inspect Surveillance Activity</h4>
         <strong>Start Date:</strong>{{ vm.surveillance.startDate | date : 'mediumDate' : 'UTC' }}<br />
         <strong>End Date:</strong>{{ vm.surveillance.endDate | date : 'mediumDate' : 'UTC' }}<br />
         <strong>Surveillance Type:</strong>{{ vm.surveillance.type.name }}<br />
-        <strong>Number of sites surveilled:</strong>{{ vm.surveillance.randomizedSitesUsed ? vm.surveillance.randomizedSitesUsed : 'N/A' }}<br />
+        <strong>Number of sites surveilled:</strong>{{ vm.surveillance.randomizedSitesUsed ? vm.surveillance.randomizedSitesUsed : 'N/A' }}
```
andlar: Needs a `<br />` at the end of the line

src/app/admin/components/surveillance/surveillanceInspect.html
```diff
@@ -36,7 +36,9 @@ <h4 class="modal-title" id="inspectLabel">Inspect Surveillance Activity</h4>
...
+        <strong>Surveillance initiated by:</strong>{{ vm.surveillance.authority === 'ROLE_ADMIN' ? 'ONC' : 'ONC-ACB' }}<br />
+        <div class="row" ng-if="vm.workType !== 'initiate'">
```
andlar: What is this div for?

src/app/components/surveillance/surveillance.html
```diff
@@ -34,6 +34,10 @@
               <a href uib-tooltip="Whether or not a non-conformity was found for the conducted surveillance." tooltip-trigger="mouseenter focus click">Surveillance Result</a></th>
             <td><span ng-repeat="res in vm.surveillanceResults(surv) | orderBy:vm.sortResults track by $index">{{ res }}<br /></span></td>
           </tr>
+          <tr>
+            <th scope="row">Surveillance initiated by</th>
```
andlar: This `<th>` probably needs a tooltip, though will need to get that text before it can be added. If everything else is fixed we can ignore this comment, but if we can get the text before then it should be added here.

dlucas1:
Note that I am still awaiting text for the tooltip on "Initiated by:"

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -14,6 +14,10 @@
         vm.deleteRequirement = deleteRequirement;
         vm.editRequirement = editRequirement;
         vm.inspectNonconformities = inspectNonconformities;
+        vm.isAcbAdmin = authService.isAcbAdmin;
+        vm.isAcbStaff = authService.isAcbStaff;
+        vm.isChplAdmin = authService.isChplAdmin;
+
```
andlar: Why the blank line?

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -149,46 +163,46 @@
...
+                            }
+                        },function (error) {
+                            if (error.data.errorMessages && error.data.errorMessages.length > 0) {
+                                vm.errorMessages = error.data.errorMessages;
+                            } else {
+                                vm.errorMessages = [error.statusText];
+                            }
+                        });
+                }
         }
```
andlar: Why all the indentation?

src/app/admin/components/surveillance/editSurveillance.html
```diff
@@ -50,6 +57,15 @@ <h4 class="modal-title" id="editSurveillanceLabel"><span ng-if="vm.workType ===
         </select>
         <div class="text-danger" ng-if="(vm.showFormErrors || vm.editForm.surveillanceType.$touched) && vm.editForm.surveillanceType.$error.required">Field is required</div>
       </div>
+    </div> 
+    <div class="form-group" ng-if="vm.workType === 'initiate' && vm.isChplAdmin() && (vm.isAcbAdmin() || vm.isAcbStaff())">
```
andlar: If someone is both ACB Admin and ACB Staff this select box won't appear, but we need it to appear.

src/app/admin/components/surveillance/editSurveillance.html
```diff
@@ -50,6 +57,15 @@ <h4 class="modal-title" id="editSurveillanceLabel"><span ng-if="vm.workType ===
         </select>
         <div class="text-danger" ng-if="(vm.showFormErrors || vm.editForm.surveillanceType.$touched) && vm.editForm.surveillanceType.$error.required">Field is required</div>
       </div>
+    </div> 
+    <div class="form-group" ng-if="vm.workType === 'initiate' && (vm.isChplAdmin() && (vm.isAcbAdmin() || vm.isAcbStaff()) || (vm.isAcbAdmin() && vm.isAcbStaff()))">
```
andlar: I think this might work, but it's really hard to parse. The *reason* to have this select box is if the user has more than one possible role. You know how many roles they have based on the `vm.authorities` array's length. Suggest replacing the entire `&&` portion after `'initiate'` with just `vm.authorities.length > 1`

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -161,12 +162,22 @@
 
         function missingEndDate () {
             var hasOpen = false;
-            for (var i = 0; i < vm.surveillance.requirements.length; i++) {
-                for (var j = 0; j < vm.surveillance.requirements[i].nonconformities.length; j++) {
-                    hasOpen = hasOpen || (vm.surveillance.requirements[i].nonconformities[j].status.name === 'Open');
+            if(vm.surveillance.requirements){
```
andlar: This line is good.

src/app/admin/components/surveillance/editSurveillance.controller.js
```diff
@@ -161,12 +162,22 @@
...
+                for (var i = 0; i < vm.surveillance.requirements.length; i++) {
+                    for (var j = 0; j < vm.surveillance.requirements[i].nonconformities.length; j++) {
+                        hasOpen = hasOpen || (vm.surveillance.requirements[i].nonconformities[j].status.name === 'Open');
+                    }
                 }
+                if(!hasOpen){
```
andlar: From here down is not. The final `return` call is where it should use the boolean that indicates whether there are open surveillances or not, combined with the existence of an end date, to return a response.

The last 'return false' that used to be 'return !hasOpen && !vm.surveillance.endDate` is where the logic happens, and that logic should be `!hasOpen && (!vm.surveillance.endDate || vm.surveillance.endDateObject)` with the English translation of: "this function tells you if you're missing an end date. You're missing an end date if you have surveillances, and at least one of them has an open nonconformity, and you have neither an endDate nor an endDateObject".

I think the current implementation would give you an error message about missing an end date, with all closed NCs, if you type a date in the endDate field, as opposed to using the date picker. That's not correct

dlucas1: The current implementation does not give an error message about missing an end date when initiating or editing a surveillance activity with all closed NCs and typing an end date into the field (also does not give an error when using the date picker). 

vm.surveillance.endDate is the value passed in from the API for an existing/edited surveillance activity, whereas vm.surveillance.endDateObject is the value in the endDate field of the editSurveillance form (which gets populated regardless of typing the date or using the date picker). The vm.surveillance.endDateObject is populated from the vm.surveillance.endDate during the editSurveillance's activate() function, thus we should only need to look at the value of the vm.surveillance.endDateObject in the editSurveillance.controller.missingEndDate(). 

Regarding the logic in missingEndDate(), here is my logic's English translation:
If there are no requirements, return false. If there are requirements, check to see if there are any open nonconformities, and if one or more exist then set hasOpen to true. 
Next, if there are only closed nonconformities, check to see if there is an endDate. If there is an endDate, then return false for missingEndDate(), otherwise return true. 

In comparison, here is the old logic: !hasOpen && !vm.surveillance.endDateObject. English translation: return true for missingEndDate() when there's no open nonconformities and no end date; return false when there's an open nonconformity and/or an endDate.

Although the old logic's code is more concise, I think the above logic is a lot easier to understand because there is only one condition to evaluate at a time. Ultimately, they appear to obtain the same result.

PR #115, opened 2017-03-23
OCD-1329: initiate/edit/add nonConformity requires CAPMustCompleteDate when entry for CAPApprovalDate

dlucas1:
Only deploying to development since staging is frozen until Wednesday 3/29.

### DIFF
--- a/src/app/admin/components/surveillance/editNonconformity.controller.spec.js
+++ b/src/app/admin/components/surveillance/editNonconformity.controller.spec.js
@@ -56,42 +56,5 @@
                 expect(mock.modalInstance.dismiss).toHaveBeenCalled();
             });
         });
-        
-        describe('business rules', function () {
-            beforeEach(function () {
-                vm.surveillance = angular.copy(Mock.surveillances[0]);
-            });
-
-            it('When creating a nonconformity, it should not require a Corrective Action Plan Must Complete Date when there is no entry for Corrective Action Plan Approval Date', function () {
-                expect(vm.isCapMustCompleteRequired()).toBe(false);
-            });
-            
-            it('When creating a nonconformity, it should require a Corrective Action Plan Must Complete Date when there is an entry for Corrective Action Plan Approval Date', function () {
-                vm.nonconformity.capApprovalDateObject = '03/22/2017';
-                expect(vm.isCapMustCompleteRequired()).toBe(true);
-            });
-            
-            it('When editing a nonconformity, it should not require a Corrective Action Plan Must Complete Date when there is no entry for Corrective Action Plan Approval Date', function () {
-                vm.workType = 'edit';
-                expect(vm.isCapMustCompleteRequired()).toBe(false);
-            });
-            
-            it('When editing a nonconformity, it should require a Corrective Action Plan Must Complete Date when there is an entry for Corrective Action Plan Approval Date', function () {
-                vm.workType = 'edit';
-                vm.nonconformity.capApprovalDateObject = '03/22/2017';
-                expect(vm.isCapMustCompleteRequired()).toBe(true);
-            });
-            
-            it('When adding a nonconformity, it should not require a Corrective Action Plan Must Complete Date when there is no entry for Corrective Action Plan Approval Date', function () {
-                vm.workType = 'add';
-                expect(vm.isCapMustCompleteRequired()).toBe(false);
-            });
-            
-            it('When adding a nonconformity, it should require a Corrective Action Plan Must Complete Date when there is an entry for Corrective Action Plan Approval Date', function () {
-                vm.workType = 'add';
-                vm.nonconformity.capApprovalDateObject = '03/22/2017';
-                expect(vm.isCapMustCompleteRequired()).toBe(true);
-            });
-        });
     });
 })();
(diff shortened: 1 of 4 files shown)


src/app/admin/components/surveillance/editNonconformity.controller.spec.js
```diff
@@ -11,21 +11,20 @@
...
-            module('chpl.admin', function ($provide) {
+            module('chpl.mock', 'chpl.admin', function ($provide) {
                 $provide.decorator('commonService', function ($delegate) {
                     $delegate.updateCP = jasmine.createSpy('updateCP');
                     return $delegate;
                 });
             });
 
-            inject(function ($controller, $rootScope, _$log_, _$q_, _commonService_) {
+            inject(function ($controller, $rootScope, _$log_, _$q_, _commonService_, _Mock_) {
```
andlar: `_$q_` isn't used, nor is `_commonService_`

src/app/admin/components/surveillance/editNonconformity.html
```diff
@@ -77,15 +77,16 @@ <h4 class="modal-title" id="editNonconformityLabel"><span ng-if="vm.workType ===
...
+          <input class="input-sm form-control" type="text" uib-datepicker-popup="MM/dd/yyyy" ng-model="vm.nonconformity.capMustCompleteDateObject" is-open="vm.nonconformity.capMustCompleteDatePicker" close-text="Close" id="capMustCompleteDate" name="capMustCompleteDate" ng-model-options="{ timezone: 'UTC' }" ng-required="vm.nonconformity.capApprovalDateObject">
           <span class="input-group-btn">
             <button type="button" class="btn btn-ai btn-sm" ng-click="vm.nonconformity.capMustCompleteDatePicker = true"><i class="fa fa-calendar"></i></button>
           </span>
         </div>
         <div class="text-danger" ng-if="(vm.showFormErrors || vm.editForm.capMustCompleteDate.$touched) && vm.editForm.capMustCompleteDate.$error.date">Improper format (mm/dd/yyyy)</div>
+        <div class="text-danger" ng-if="vm.showFormErrors && vm.isCapMustCompleteRequired()">Field is required</div>
```
andlar: This function was removed, right?

src/app/admin/components/surveillance/editNonconformity.html
```diff
@@ -202,7 +203,7 @@ <h4 class="modal-title" id="editNonconformityLabel"><span ng-if="vm.workType ===
     </div>
   </div>
   <div class="modal-footer">
-    <div class="text-danger text-left" ng-if="vm.editForm.$invalid && vm.showFormErrors">
+    <div class="text-danger text-left" ng-if="(vm.editForm.$invalid) && vm.showFormErrors">
```
andlar: Do not need parentheses

src/app/admin/components/surveillance/editNonconformity.html
```diff
@@ -227,7 +228,7 @@ <h4 class="modal-title" id="editNonconformityLabel"><span ng-if="vm.workType ===
         <li ng-if="vm.editForm.resolution.$error.required">Resolution is required</li>
       </ul>
     </div>
-    <button ng-disabled="vm.editForm.$invalid && vm.showFormErrors && vm.disableValidation"
+    <button ng-disabled="(vm.editForm.$invalid) && vm.showFormErrors && !vm.disableValidation"
```
andlar: Do not need parentheses.

Why did you `!` the `vm.disableValidation`?

dlucas1: If validation is disabled, then it wouldn't make sense to disable the save button since that's part of the validation process, right? If you look at editSurveillance.html line 141, its save button is also not disabled when validation is disabled.

andlar: That 'disableValidation' flag is used when there's something wrong with the surveillance but we have to allow it through anyway. I don't remember when that's true, but it comes up sometimes. I'm not sure why you flipped the value in the first place. While adding functionality to stuff we have to be very careful to not mess with stuff that's already been built, QA'd and deployed. Unless you track all the way back through the code to where this part originated, figure out why it's there, and then justify it, I'd be _super_ wary of flipping logic in an if statement

dlucas1: While I understand that I need to be _very_ careful with modifying existing functionality, I believe that the disabling of the save button has an existing bug due to the missing '!' in front of "vm.disableValidation". Since vm.disableValidation is hard coded to be "false" in editRequirement.controller.js line 49--and I verified that it is always false when adding a new nonconformity and editing an existing one--if I do not have the "!" in front of vm.disableValidation, then the disable functionality for the save button will not work when there are validation errors.

src/app/admin/components/surveillance/editNonconformity.controller.spec.js
```diff
@@ -11,21 +11,20 @@
         };
 
         beforeEach(function () {
-            module('chpl.admin', function ($provide) {
+            module('chpl.mock', 'chpl.admin', function ($provide) {
```
andlar: If you're not going to use the commonService/delegate thing, then this line should be `module('chpl.mock','chpl.admin');` and the decorator below should go away

PR #129, opened 2017-04-24
OCD-1415: Fixed bug related to merged developer status history

andlar:


dlucas1:
Where is $log being used?

PR #140, opened 2017-05-10
Text changes for Nonconformites & Decertified Products pages

andlar:
Has some e2e stuff in there that's in "work in progress" status

### DIFF
--- a/features/components/admin/admin.feature
+++ b/features/components/admin/admin.feature
@@ -1,0 +1,17 @@
+#features/components/admin/admin.feature
+@ignore
+Feature: Administration of the CHPL
+  As an authenticated user of the CHPL
+  I should be able to log in
+  in order to administer the CHPL
+
+  @wip
+  Scenario: Seeing the Administration Title
+    Given I am on the Administration page
+    Then the page title should be "CHPL Administration"
+
+  @wip
+  Scenario: Logging in as ROLE_ADMIN
+    Given I am on the Administration page
+    When I log in as ROLE_ADMIN
+    Then the login form should go away
--- a/features/components/admin/admin.login.steps.js.template
+++ b/features/components/admin/admin.login.steps.js.template
@@ -1,0 +1,10 @@
+//features/components/admin/admin.steps.js
+var page = require('./admin.page.js');
+var {defineSupportCode} = require('cucumber');
+
+defineSupportCode(({Given, When, Then}) => {
+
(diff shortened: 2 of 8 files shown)


features/components/admin/admin.feature
```diff
@@ -0,0 +1,17 @@
+#features/components/admin/admin.feature
+@ignore
+Feature: Administration of the CHPL
+  As an authenticated user of the CHPL
+  I should be able to log in
+  in order to administer the CHPL
+
+  @wip
+  Scenario: Seeing the Administration Title
```
dlucas1: Could we write scenarios such that each one is an independent clause/complete sentence? For example, maybe this one could be something like "The Administration page title shows appropriate text before login"

features/components/admin/admin.feature
```diff
@@ -0,0 +1,17 @@
...
+    Given I am on the Administration page
+    Then the page title should be "CHPL Administration"
+
+  @wip
+  Scenario: Logging in as ROLE_ADMIN
```
dlucas1: Same as above comment. Perhaps this could be something like, "Login form disappears upon successful login"

features/components/admin/admin.login.steps.js.template
```diff
@@ -0,0 +1,10 @@
+//features/components/admin/admin.steps.js
+var page = require('./admin.page.js');
+var {defineSupportCode} = require('cucumber');
+
+defineSupportCode(({Given, When, Then}) => {
+
+    When('I log in as ROLE_ADMIN', function () {
```
dlucas1: Do we want "When I log in as..." to be a WHEN, a GIVEN, or both? In this case, it's being used as a WHEN, but I suspect that future tests will not be testing this specifically, so it will be used as a GIVEN. While it's not necessary to add the GIVEN right now, just keep in mind that future tests that aren't specifically testing the login should use the GIVEN form for this functionality.

In addition, we'll probably want to parameterize the ROLE such that future tests can pass the role name in something like this: "GIVEN I log in as 'ROLE_ADMIN'" or "GIVEN I log in as "ACB_ADMIN". Again, it may not be necessary to update this "WHEN I log in as ROLE_ADMIN", but it's something to keep in mind once we start using the login as a prerequisite GIVEN for other tests.

andlar:
All of the stuff for e2e testing is still wip stuff, as I mentioned in the PR. There are many things wrong with what's in those files now, but that's not what this PR is about

PR #159, opened 2017-06-14
Surveillance and eye updates

andlar:
* Added Listing History explanatory text
* Updated summary description of surveillance result


src/app/product/history/history.html
```diff
@@ -26,4 +26,7 @@ <h2 class="modal-title" id="titleLabel">Product History</h2>
       </tbody>
     </table>
   </div>
+  <div class="modal-footer">
+    This module gives a basic overview of modifications made to the listing. For a more detailed history, please use the <code>/activity/certified_products/{id}</code> call. For more details, visit the CHPL API page here: <a href="https://chpl.healthit.gov/#/resources/chpl_api">https://chpl.healthit.gov/#/resources/chpl_api</a>
```
kekey1: Should the href be relative or from a property so it points to the API docs for the same system (dev, stg, prod) the user is currently using? Do you have the actual {id} and if so would it be useful to put that in the URL instead of "{id}"? If not, how else would the user know where to find the appropriate ID?

PR #207, opened 2017-09-29
Ocd 1776 sed tweaks

brianlindsy:


src/app/components/listing_details/details.html
```diff
@@ -7,7 +7,6 @@
       <input type="checkbox" ng-model="vm.viewAllCerts" id="viewAllCerts"></input>
       <label for="viewAllCerts">See all Certification Criteria / Clinical Quality Measures</label>
     </p>
-
```
andlar: Can you put these spaces back in, both here and all the way through? With the long HTML files it's sometimes easier to break up sections with a single new line, and that helps when scanning through the code.

src/app/components/listing_details/details.html
```diff
@@ -203,4 +196,4 @@
       </div>
     </div>
   </div>
-</div>
+</div>
```
andlar: I don't know why, but some editors take away the last newline on files; it's probably not bad to remove them, but then that does show up as a change in git

src/app/components/listing_details/sed/sed.html
```diff
@@ -32,7 +32,7 @@
           <tr ng-repeat="process in vm.ucdProcesses | orderBy: [vm.sortProcesses]">
             <td>
               <ul>
-                <li ng-repeat="crit in process.criteria | orderBy:vm.sortCert">{{ crit.number }}</li>
+                <li ng-repeat="crit in process.criteria | orderBy:vm.sortCert">{{ crit.number }} {{ crit.title }}</li>
```
andlar: The "certification criteria" accordion piece has the number / title as `number: title`, or `170.314 (a)(1): Computerized provider order entry`. The difference is the colon (`:`). I know it's minor, but consistency across the CHPL is something we should encourage

src/app/components/listing_details/details.html
andlar: Can you add the blank lines back in, both here and throughout the file? With the longer HTML files having some blank space at times can help with scanning through, and gives a sense of sections

src/app/components/listing_details/details.html
andlar: I don't know why some editors remove the trailing newline from the ends of files, and it's probably not a big deal, but it does show a meaningless difference in git, and I'd like to avoid that, as much as possible

PR #249, opened 2018-03-05
OCD-2091: UI: Add charts page

tmy1313:


src/app/charts/charts.controller.js
```diff
@@ -8,611 +8,45 @@
     function ChartsController ($log, networkService) {
         var vm = this;
 
-        vm.applyFilter = applyFilter;
-        vm.toggleSeries = toggleSeries;
-
-        activate();
+        prepareSedParticipantCountChart();
```
andlar: With controllers, the convention in this project is that the top section (before the first ////////// line) is sort of the interface. The "vm.something = something" are the "public" methods, sorted alphabetically, and the "activate()" call at the end is basically a call to the constructor. 

In this paradigm, "public" methods are usually called by the HTML, and occasionally invoked from other controllers & directives.

The next section is where the public methods are defined, also alphabetically. The final section (after the second /////) is optional, and is where helper functions (for the public functions) live. I usually name those with a prefixed underscore.

For this iteration of this page, I'd put the contents of your prepareSedParticipatnCountChart function in the "activate" one. We'll probably break it out as we add more charts, but not needed yet

src/app/charts/charts.controller.js
```diff
@@ -8,611 +8,45 @@
...
+            networkService.getSedParticipantStatisticsCount().then(function (data) {
+                vm.charts = {
+                    sedParticipantCounts: {
+                        type: "ColumnChart",
+                        data: {
+                            cols: [
+                                { label: 'Participants', type: 'number'},
+                                { label: 'SED Count', type: 'number'},
+                            ],
+                            rows: getSedParticipantCountDataInChartFormat(data),
```
andlar: This function could be a private one (_getSedParticipantCountDataInChartFormat)

src/app/charts/charts.controller.js
```diff
@@ -8,611 +8,45 @@
...
-                        calc: function () {
-                            return null;
+        function prepareSedParticipantCountChart() {
+            networkService.getSedParticipantStatisticsCount().then(function (data) {
+                vm.charts = {
+                    sedParticipantCounts: {
+                        type: "ColumnChart",
+                        data: {
+                            cols: [
+                                { label: 'Participants', type: 'number'},
```
andlar: We need to figure out what these axes / labels / titles are. I usually send screenshots to the the PM for her to check with the end customer as needed. It'll be tougher right now due to travel, but final acceptance depends on her saying it looks good :)

src/app/charts/charts.controller.js
```diff
@@ -8,611 +8,45 @@
...
-                        }
-                        if (testDate > vm.maxDate) {
-                            vm.maxDate = testDate;
-                        }
-                    }
-                    mapObj[data.date].values.push(data.value);
-                });
+        function getSedParticipantCountDataInChartFormat(data) {
+            data.sedParticipantStatisticsCounts.sort(function(a, b) {
+                return parseInt(a.participantCount) - parseInt(b.participantCount);
```
andlar: parseInt needs a radix, otherwise values like 9 will be converted to octal

src/app/charts/charts.controller.js
```diff
@@ -8,611 +8,45 @@
...
             });
-            vm.safeRows = [];
-            angular.forEach(mapObj, function (row) {
-                c = [{v: new Date(row.date)}];
-                c = c.concat(row.values.map(function (val) { return {v: val }; }))
-                vm.safeRows.push({c: c});
+            return data.sedParticipantStatisticsCounts.map(obj =>  {
```
andlar: We can't use arrow functions. At least, I don't think we can; something in our build process isn't that up to date

RELEASE_NOTES.md
```diff
@@ -1,5 +1,13 @@
 # Release Notes
 
+## Version 9.15.0
```
andlar: We generally go with "Version TBD, Date TBD" in this field. They'll get filled in when we actually push to PROD

andlar:
Have you run the gulp tests? I think those will fail at the current stage

andlar:
Outside of whatever conflict there is in the release notes I think this is good to go

PR #251, opened 2018-03-19
OCD-2084 - Generate charts for SED Participants

tmy1313:


src/app/charts/charts.controller.js
```diff
@@ -48,5 +56,209 @@
...
+                    },
+                    options: {
+                        title: 'Safety Enhanced Design Test Participants by Gender',
+                    },
+                }
+            });
+        }
+
+        function _getParticipantGenderCountDataInChartFormat (data) {
+            var genderData = [{c: [{ v: 'Male'},{v: data.maleCount}]}, {c: [{ v: 'Female'},{v: data.femaleCount}]}];
```
andlar: There _is_ a possibility of "Unknown" as a gender. Just remembered that. Might have impact elsewhere

src/app/charts/charts.controller.js
```diff
@@ -48,5 +56,209 @@
...
+            //console.log(data);
```
andlar: `$log.debug(data)` or `$log,info(data)` is the service that we use for logging

src/app/charts/charts.controller.js
```diff
@@ -48,5 +56,209 @@
...
+            });
+
+            //Sum participants based on years experience
+            //var experienceMap = new Map();
+            var experienceMap = {};
+            angular.forEach(data.participantExperienceStatistics, function (value) {
+                var count = value.participantCount;
+                //if (experienceMap.has(value.experienceYears)) {
+                if (value.experienceYears in experienceMap) {
+                    experienceMap
```
andlar: What does this line do?

src/app/charts/charts.html
```diff
@@ -7,7 +7,33 @@ <h1>Charts</h1>
   <div class="main-content" id="mainContent" tabindex="-1" ng-form="optionForm">
     <div class="row">
       <div class="col-sm-12">
-        <div google-chart chart="vm.charts.sedParticipantCounts" style="height:450px; width:100%;"></div>
+        <div google-chart chart="vm.sedParticipantCounts" style="height:450px; width:100%;"></div>
+      </div>
+    </div>
+    <div class="row">
+      <div class="col-sm-4">
+        <div google-chart chart="vm.participantGenderCounts" style="height:450px;"></div>
```
andlar: Do these not need widths?

src/app/charts/charts.controller.js
tmy1313: Thanks for the info.  I'm going to remove this line.  Where does $log write to?  The console?

src/app/charts/charts.html
tmy1313: The width is defined by the container div - "<div class="col-sm-4">".

src/app/charts/charts.controller.js
tmy1313: This is going to take some work.  Table, App, and UI chnage.  Should we include it in this sprint, or make it a new story?

andlar: It's part of the data, so it's part of this story.

src/app/charts/charts.controller.js
andlar: Mostly the console, but at different levels of verbosity, depending on your browser's "developer tools" settings. Specifically, I think, `debug` only shows up in "verbose" mode on Chrome. The other use of them is that the `debug` level gets output to the command line console during tests, so that can be used for debugging sometimes.

src/app/charts/charts.html
andlar: Not sure what you were trying to say here. I don't need widths here if they're not required, but all of the bar charts have them, so I was wondering why the pie-charts don't

src/app/charts/charts.controller.js
tmy1313: Thanks for catching this.  I'm not sure what this is - looks like I accidentally deleted part of this line maybe or pasted something in by accident.  I'm going to take a look at it and see what is going on.

PR #271, opened 2018-05-09
OCD-2210 - remove line that compiles html before returning function

brianlindsy:


PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,17 @@
+
+## Version 9.19.0
```
andlar: You're behind upstream. Version 9.19.0 was released on 7 May.

PR #282, opened 2018-05-22
Ocd 2260 fix corrective action plan activity

kekey1:


src/app/common/network.service.js
```diff
@@ -332,6 +333,11 @@
             }
         }
 
+        function getCorrectiveActionPlanActivity (activityRange) {
```
andlar: There should be a test for this method in the network.service.spec.js file

kekey1: Added! Can you check the PR again?

PR #287, opened 2018-05-30
Ocd 2252 -  Updated edit product and merge dev to support multiple error msgs

tmy1313:


src/app/admin/components/certifiedProduct/developer/merge.html
```diff
@@ -180,9 +180,9 @@ <h4>To form this one</h4>
         <li ng-repeat="error in vm.contactErrorMessages">{{ error }}</li>
       </ul>
     </div>
-    <div ng-if="vm.errorMessage && vm.errorMessage.length !== 0" class="row">
-      <div class="col-sm-12 text-left bg-danger">
-        Error in developer merge: {{ vm.errorMessage }}
+    <div ng-if="vm.isArray(vm.errorMessage) && vm.errorMessage.length !== 0" class="row">
```
andlar: I don't think you need this `isArray` check here; I think you could have left this line alone, which then means the `vm.isArray = angular.isArray` unnecessary in the controller

src/app/admin/components/certifiedProduct/developer/merge.html
```diff
@@ -180,9 +180,9 @@ <h4>To form this one</h4>
...
+      <div ng-repeat="msg in vm.errorMessage" class="col-sm-12 text-left bg-danger">
```
andlar: Recommend adding `| orderBy: 'name'` to the end of the repeat clause. Even if it's not perfect, it'll be better than no sorting

src/app/admin/components/certifiedProduct/developer/merge.controller.js
```diff
@@ -13,6 +13,7 @@
         vm.cancel = cancel;
         vm.hasDateMatches = hasDateMatches;
         vm.hasStatusMatches = hasStatusMatches;
+        vm.isArray = angular.isArray;
```
andlar: If removing the need for this in the view, can remove here.

src/app/admin/components/certifiedProduct/developer/merge.controller.js
```diff
@@ -119,7 +120,14 @@
                         vm.errorMessage = response.error;
                     }
                 },function (error) {
-                    vm.errorMessage = error.data.error;
+                    vm.errorMessage = [];
+                    if (error.data.error) {
+                        vm.errorMessage.push(error.data.error);
+                    } else if (error.data.errorMessages) {
+                        vm.errorMessage = error.data.errorMessages
```
andlar: Is it ever possible to have both an `error.data.error` and an `error.data.errorMessages`? If so, this would lose the first one

src/app/admin/components/certifiedProduct/management.html
```diff
@@ -515,7 +515,8 @@ <h3 ng-if="vm.inspectingCp">System developer information</h3>
                 <strong>Contact information</strong><br />
                 <ai-contact contact="vm.activeProduct.contact"></ai-contact>
               </span>
-              <p ng-if="vm.productMessage !== undefined" class="bg-danger">{{ vm.productMessage }}</p>
+              <p ng-if="vm.productMessage !== undefined && !vm.isArray(vm.productMessage)" class="bg-danger">{{ vm.productMessage }}</p>
```
andlar: For the first "if", if you just use `vm.productMessage && !vm.isArray...` then it will fail on productMessage as false, undefined, or null, and then the assignment to null in the controller isn't needed.

src/app/admin/components/certifiedProduct/management.directive.js
```diff
@@ -352,6 +353,7 @@
                     activeProduct: function () { return vm.activeProduct; },
                 },
             });
+            vm.productMessage = null;
```
andlar: Don't think is is necessary if the html if is updated

src/app/admin/components/certifiedProduct/product/edit.controller.js
```diff
@@ -84,7 +84,15 @@
                         $uibModalInstance.dismiss('An error occurred');
                     }
                 },function (error) {
-                    $uibModalInstance.dismiss(error.data.error);
+                    var errorMessage = '';
+                    if (error.data.error) {
+                        errorMessage = error.data.error;
+                    } else if (error.data.errorMessages) {
+                        errorMessage = error.data.errorMessages
```
andlar: Same question as above in the other "compilation" of these things.

PR #299, opened 2018-06-28
Modified the UX for multiselect on the Edit Criteria modal

tmy1313:
OCD-767


src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+
+    /////////////////////////////////////////////////////////////////
+
+    function activate () {
+        populateOptions();
+        populateSelectedItems();
+    }
+
+    function addItem () {
+        if (ctrl.selectedItem[ctrl.itemKey] === -99) {
```
andlar: checkstyle would call this a magic number. What's -99 for? Is there a better way of doing whatever this is trying to do? Some sort of boolean or something?

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+        var retValue = false;
+        angular.forEach(ctrl._selectedItems, function (arrayItem) {
+            if (arrayItem.item[ctrl.itemKey] === item[ctrl.itemKey]) {
+                retValue = true;
+            }
+        });
+        return retValue;
+    }
+
+    function getItemByKey (key) {
```
andlar: This and `getOnChangeObject` look like private functions, and should be below another row of `/////...///`s

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+addItemsPlaceholder -       Direction that appear in the text box
+onChange -                  Callback that is used when the user makes a change to the control
+*/
+
+angular.module('chpl').component('aiMultiselect', {
+    templateUrl: './app/components/util/multiselect.html',
+    controller: MultiselectController,
+    bindings: {
+        items: '<',
+        itemDisabled: '<',
```
andlar: The comment for this binding says it's a callback function, but the binding itself is a one-way binding, not a callback

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+        var itemToReturn = null;
+        angular.forEach(ctrl.items, function (item) {
+            if (item[ctrl.itemKey] === key) {
+                itemToReturn = item;
+            }
+        });
+        return itemToReturn;
+    }
+
+    function getOnChangeObject (action, selectedItem) {
```
andlar: Why does this exist? It doesn't add any logic, or any processing, but just asks the reader to go find out what it does

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+        return {'action': action, 'item': selectedItem};
+    }
+
+    function isItemDisabled (item) {
+        if (ctrl.itemDisabled) {
```
andlar: Is this supposed to check to see if `itemDisabled` exists? And is a function? It's super unclear what it's trying to do here. If that's the case, suggest making it explicit, with something like `if (typeof ctrl.itemDisabled === 'function') {`

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+        var index = ctrl._selectedItems.indexOf(item);
+        if (index > -1) {
+            ctrl._selectedItems.splice(index, 1);
+        }
+
+        ctrl.onChange(getOnChangeObject('Remove', item));
+    }
+}
+
+/*
```
andlar: I like the comments, but think they might better placed directly next to the bindings. Could go either way on this, but I'd prefer not to have to page up/down to understand things. Might also help highlight things like the odd comment regarding binding to the `itemDisabled` below

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+        additionalInputMaxLength: '@',
+        additionalInput2: '@',
+        additionalInput2Label: '@',
+        additionalInput2Required: '@',
+        additionalInput2MaxLength: '@',
+        addItems: '@',
+        addItemsOptionText: '@',
+        addItemsPlaceholder: '@',
+        onChange: '<',
+    },
```
andlar: Recommend alphabetizing these as well. Also, is there any way to combine the additionalInput fields into an array? I hate seeing `additionalInput2`, because that just screams out "what happens when we need three?" Could it be reworked to support an array of 0+ elements?

src/app/components/util/multiselect.html
```diff
@@ -0,0 +1,74 @@
+<div>
+    <div ng-repeat="option in $ctrl._selectedItems track by $index">
```
andlar: In the controller, `this` was referenced with `ctrl`, but in the HTML it's called out as `$ctrl`. Why?

src/app/components/util/multiselect.html
```diff
@@ -0,0 +1,74 @@
...
+        <div class="selected-item fade-in">
+            <span class="closebtn" ng-click="$ctrl.removeItem(option)">&times;</span>
```
andlar: Most of our icons use the font-awesome font; I think this would be `<i class="fa fa-times"></i>` instead of the `&times;`. Should also have a 508 friendly `<span class="sr-only">Remove item {{ option.item.name }}</span>` inside the span. I know we aren't required to 508 the admin section now, but we will be later on at least parts of it, so keeping in that habit now is a useful idea. Also, if this is a button, use HTML buttons; they react better to keyboard nav, touch screen, etc. Take a look at the CMS_Widget dropdown directive (or compare widget directive) for how those work

src/app/components/util/multiselect.html
```diff
@@ -0,0 +1,74 @@
+<div>
+    <div ng-repeat="option in $ctrl._selectedItems track by $index">
+        <div class="selected-item fade-in">
+            <span class="closebtn" ng-click="$ctrl.removeItem(option)">&times;</span> 
+            <strong>{{ option.item.name }}</strong>
+            <div ng-if="$ctrl.additionalInput" style="padding-top: 3px;" ng-form="{{ $ctrl.identifier }}_form">
```
andlar: Really don't like having styles as an attribute. I think it's a no-no in our `lint-html` checker as well. Recommend putting the ng-form on the main component tag as well; it's a more over-arching thing, and though the form elements are only in the additional information stuff, the "remove me" buttons are conceptually part of the fom

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
...
+        itemDisabled: '<',
+        selectedItemKeys: '<',
+        identifier: '@',
+        itemKey: '@',
+        itemText: '@',
+        placeholder: '@',
+        additionalInput: '@',
+        additionalInputLabel: '@',
+        additionalInputRequired: '@',
+        additionalInputMaxLength: '@',
```
andlar: Both inputRequired and inputMaxLength aren't used anywhere; do they need to be here?

src/app/components/util/multiselect.html
```diff
@@ -0,0 +1,74 @@
...
+                        <input type="text" 
+                            id="{{ $ctrl.identifier }}_additionalInput"
+                            name="{{ $ctrl.identifier }}_additionalInput"
+                            ng-model="option.additionalInputValue"
+                            ng-change="$ctrl.additionalInputChange(option)" 
+                            ng-maxlength="$ctrl.additionalInputMaxLength"
+                            ng-required="{{ $ctrl.additionalInputRequired }}">
+                    </div>
+                </div>
+                <p class="text-danger" ng-show="{{ $ctrl.identifier }}_form.{{ $ctrl.identifier }}_additionalInput.$error.maxlength">
```
andlar: Do these errors feed back up to the criteria controller to show in the "master list" of errors at the bottom of the modal? In a directive, I'd probably have a two-way binding to an array of errors; with a component it should probably be a callback that sends the "active" errors to the embedding controller to display the errors.

I'm specifically looking for the way we have errors displayed both at the site of the error, and a roll-up of all the errors on the page at the bottom, in the modal footer, with the various buttons.

Also, if a component input value is invalid, does it disable the embedding controllers' form's "save" button?

src/app/index.scss
```diff
@@ -643,6 +643,63 @@ option[disabled] {
     width: 100%;
     background: #ccc;
 }
+
+
+/* The alert message box */
+.selected-item {
```
andlar: Do these colors match the standard color swatch for CHPL? I don't recognize them, but I also haven't looked at color codes for a while. I generally try to avoid adding new CSS stuff if there's a bootstrap class that I can use instead (or a button, or link, or whatever)

src/app/index.scss
```diff
@@ -643,6 +643,63 @@ option[disabled] {
...
+}
+.additional-input:-ms-input-placeholder { /* IE 10+ */
+    color: #777777;
+}
+.additional-input:-moz-placeholder { /* Firefox 18- */
+    color: #777777;
+}
+
+/* The close button */
+.closebtn {
```
andlar: If you make this thing a button, is any of this required?

src/app/index.scss
```diff
@@ -643,6 +643,63 @@ option[disabled] {
...
+}
+
+/* When moving the mouse over the close button */
+.closebtn:hover {
+    color: black;
+}
+
+.fade-in {
+    animation:opac 0.3s}
+    @keyframes opac{from{opacity:0} to{opacity:1}
```
andlar: Something is weird here. It looks like the `@keyframes` isn't part of the `.fade-in` declaration; is that the case?

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -13,7 +13,12 @@
         vm.extendSelect = utilService.extendSelect;
         vm.isToolAvailable = isToolAvailable;
         vm.save = save;
-
+        vm.testFunctionalityOnChnage = testFunctionalityOnChnage;
+        vm.testProceduresOnChange = testProceduresOnChange;
+        vm.testToolsOnChange = testToolsOnChange;
+        vm.isTestToolDisabled = isTestToolDisabled;
+        vm.testDataOnChange = testDataOnChange;
+        vm.testStandardOnChange = testStandardOnChange;
```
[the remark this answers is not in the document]
tmy1313: Adcdressed

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -32,6 +37,13 @@
             vm.cert.metViaAdditionalSoftware = vm.cert.additionalSoftware && vm.cert.additionalSoftware.length > 0;
             vm.hasIcs = hasIcs;
             vm.resources = resources;
+
+            vm.selectedTestFunctionalityKeys = getSelectedTestFunctionalityKeys();
+            vm.selectedTestToolKeys = getSelectedTestToolKeys();
+            vm.selectedTestProcedureKeys = getSelectedTestProcedureKeys();
+            vm.selectedTestDataKeys = getSelectedTestDataKeys();
+            vm.selectedTestStandardKeys = getSelectedTestStandardKeys();
+
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -48,6 +60,51 @@
             $uibModalInstance.close();
         }
 
+        function testDataOnChange (action) {
+            if (action.action === 'Remove') {
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -60,5 +117,162 @@
...
+
+        function getSelectedTestToolKeys () {
+            var ttKeys = [];
+            angular.forEach(vm.cert.testToolsUsed, function (tt) {
+                ttKeys.push({'key': tt.testToolId, 'additionalInputValue': tt.testToolVersion});
+            });
+            return ttKeys;
+        }
+
+        function isTestToolDisabled (testTool) {
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/listing_details/criteriaModal.html
```diff
@@ -63,32 +63,21 @@ <h4 class="modal-title" id="modalLabel">Certification Criteria: {{ vm.cert.numbe
...
-                            ng-options="option.name as option.name for option in vm.resources.testStandards.data | orderBy: 'name'">
-                    </select>
-                    <div class="input-group-btn"><button class="btn btn-ai-success form-control" ng-click="vm.addNewValue(vm.cert.testStandards, vm.newTestStandard); vm.newTestStandard = {}"><i class="fa fa-plus"></i> <span class="sr-only">Add item</span></button></div>
-                  </div>
-                </div>
-                <label class="col-sm-6 control-label" for="newTestStandardExpandableName">Add new Test Standard to available options</label>
-                <div class="col-sm-6">
-                  <input type="text" class="form-control" ng-model="vm.newTestStandard.expandableName" id="newTestStandardExpandableName" ng-change="vm.extendSelect(vm.resources.testStandards.data, vm.newTestStandard.expandableName); vm.newTestStandard.testStandardName = vm.newTestStandard.expandableName"></input>
+                <label class="col-sm-6 control-label" for="testTool{{ item.id }}">Test Standard</label>
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/util/multiselect.component.js
```diff
@@ -0,0 +1,194 @@
+function MultiselectController () {
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/util/multiselect.component.js
tmy1313: Addressed

src/app/components/util/multiselect.component.js
tmy1313: This method is called several times in the component.  It allows us to change the structure of the object in a single place, rather than hunting through the code to replace all instances.

src/app/components/util/multiselect.component.js
tmy1313: Addressed

src/app/components/util/multiselect.component.js
tmy1313: Addressed

src/app/components/util/multiselect.html
tmy1313: Honestly, I don;t know.  That was how the component example I was looking at did it.  But I tried switching over to `ctrl` in the HTML and it doesn;t seem to work.  Not sure why...

src/app/components/util/multiselect.html
tmy1313: Addressed

src/app/components/util/multiselect.component.js
tmy1313: They are used in the HTML.

src/app/components/util/multiselect.html
tmy1313: Addressed

src/app/index.scss
tmy1313: Typo - fixed.

src/app/index.scss
tmy1313: Pulled the coloring out of the CSS.  The colors were based on the bootstrap alert-success colors.  Used the alert-success style instead.

src/app/components/listing_details/criteriaModal.html
```diff
@@ -187,51 +150,23 @@ <h4 class="modal-title" id="modalLabel">Certification Criteria: {{ vm.cert.numbe
...
-                    <div class="input-group-btn"><button class="btn btn-ai-success form-control" ng-click="vm.addNewValue(vm.cert.testToolsUsed, vm.newTestTool); vm.newTestTool = {}"><i class="fa fa-plus"></i> <span class="sr-only">Add item</span></button></div>
-                  </div>
-                </div>
-                <label class="col-sm-6 control-label" for="newTestToolVersion">Test Tool Version</label>
-                <div class="col-sm-6">
-                  <input type="text" class="form-control" ng-model="vm.newTestTool.testToolVersion" id="newTestToolVersion" id="newTestToolVersion" ng-maxlength="50"></input>
-                  <div class="text-danger" ng-if="vm.editForm.newTestToolVersion.$error.maxlength">Field is too long</div>
+                <label class="col-sm-6 control-label" for="testTool{{ item.id }}">Test Tool</label>
+                <div class="col-sm-6">
+                  <ai-multiselect
```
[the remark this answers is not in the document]
tmy1313: Name changed

src/app/components/util/multiselect.component.js
tmy1313: Addressed - it is not very intuitive how to pass parameters using the callback.
http://www.codelord.net/2016/05/13/understanding-angulars-and-binding/

src/app/components/listing_details/criteriaModal.html
```diff
@@ -187,51 +150,23 @@ <h4 class="modal-title" id="modalLabel">Certification Criteria: {{ vm.cert.numbe
...
+                    item-text="name" 
+                    items="vm.resources.testTools.data"
+                    item-disabled="vm.isTestToolDisabled(item)"
+                    selected-item-keys="vm.selectedTestToolKeys"
+                    placeholder="Select a Test Tool"
+                    on-change="vm.testToolsOnChange(action)"
+                    additional-input="true"
+                    additional-input-label="Version"
+                    additional-input-max-length="50"
+                    additional-input-required="true">
```
andlar: Does this error feed up to the parent component? If so, it should show in the `<footer>` section of the modal with the other errors on the criteria. If not, it should

src/app/components/util/expandableList.html
```diff
@@ -0,0 +1,77 @@
...
+            </button> 
+            <strong>{{ option.item.name }}</strong>
+            <div ng-if="$ctrl.additionalInput" class="additional-input">
+                <div class="row">
+                    <div class="col-sm-5">
+                        <label for="{{ $ctrl.identifier }}_additionalInput">{{ $ctrl.additionalInputLabel }}</label>
+                    </div>
+                    <div class="col-sm-5">
+                        <input type="text" 
+                            id="{{ $ctrl.identifier }}_additionalInput"
```
andlar: One of the style guides I added and am working towards is to use "dash case" for IDs. The command `gulp html-lint` (or `gulp lint-html`, as I can't recall which way it should be) shows those errors. Unfortunately, the `{{ something }}` always triggers as bad, but the goal is to migrate all of our IDs to something like `{{ $ctrl.identified }}-additional-input` in this case. Names (as in the row below) should stay camel case, as those can be used by AngularJS for form validation / errors / etc.

src/app/components/util/expandableList.html
```diff
@@ -0,0 +1,77 @@
...
+                            name="{{ $ctrl.identifier }}_additionalInput"
+                            ng-model="option.additionalInputValue"
+                            ng-change="$ctrl.additionalInputChange(option)" 
+                            ng-maxlength="$ctrl.additionalInputMaxLength"
+                            ng-required="{{ $ctrl.additionalInputRequired }}">
+                    </div>
+                </div>
+                <p class="text-danger" ng-show="{{ $ctrl.identifier }}_form.{{ $ctrl.identifier }}_additionalInput.$error.maxlength">
```
andlar: Almost exclusively we use 'ng-if' instead of 'ng-show'. There are a few specific instances of where the show is used, but it should be the exception, not the default. If you want to justify the show, feel free

src/app/index.scss
```diff
@@ -643,6 +643,64 @@ option[disabled] {
...
+    color: black;
+}
+
+.fade-in {
+    animation:opac 0.3s;
+    @keyframes opac{from{opacity:0} to{opacity:1}}
+}
+
+.additional-input {
+    padding-top: 3px;
```
andlar: I always get twitchy when I see pixels anywhere. My goal is to leverage the bootstrap styling in as many places as possible, and *not* use pixels at all (barring extenuating circumstances). Can whatever is styled with this have a different style applied that does something like what you want?

In any case, if you do think the pixels are important, please make sure they work correctly on a variety of screen sizes. It's basically a CYA; if we use bootstrap styles to do responsiveness, we can say "bootstrap did it". If we make our own adjustments to stylings (especially spacing) and responsiveness fails, whether or not we did it, then it's on us. I prefer to avoid that 😉

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -32,14 +36,23 @@
...
+            vm.selectedTestDataKeys = _getSelectedTestDataKeys();
+            vm.selectedTestStandardKeys = _getSelectedTestStandardKeys();
         }
 
         function cancel () {
             vm.cert = angular.copy(vm.certSave);
             $uibModalInstance.dismiss('cancelled');
         }
 
+        function isTestToolDisabled (testTool) {
```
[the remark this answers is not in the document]
tmy1313: From a philosophical standpoint (based on the Clean Code book) it should probably be written:
```
function isTestToolDisabled() {
    return isTestToolRetired();
}
```
This way we could easily add another reason why the Test Tool would be disabled: ` return isTestToolRetired() || isTestToolExpired();`

Just my opinion, though.

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -48,17 +61,236 @@
...
+
+        function _getSelectedTestToolKeys () {
+            var ttKeys = [];
+            angular.forEach(vm.cert.testToolsUsed, function (tt) {
+                ttKeys.push({'key': tt.testToolId, 'additionalInputValue': tt.testToolVersion});
+            });
+            return ttKeys;
+        }
+
+        function _testDataAddItem (testData) {
```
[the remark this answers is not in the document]
tmy1313: There are a lot of "add" and "remove" functions because there are a lot of the new controls on the page.  I'm not a fan of the individual controls sharing callback functions.  Seems like it could cause maintenance headaches.

src/app/components/listing_details/criteriaModal.html
```diff
@@ -241,46 +176,26 @@ <h4 class="modal-title" id="modalLabel">Certification Criteria: {{ vm.cert.numbe
...
+                    item-text="name" 
+                    items="vm.availableTestData"
+                    selected-item-keys="vm.selectedTestDataKeys"
+                    placeholder="Select Test Data"
+                    on-change="vm.testDataOnChange(action)"
+                    additional-input="true"
+                    additional-input-label="Version"
+                    additional-input-max-length="50"
+                    additional-input-required="true"
+                    additional-input2="true"
```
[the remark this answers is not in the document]
tmy1313: I think this is something that can be done as an additional feature in the future.  I agree that it should be done, but I think it is going to take me a while to figure that out.

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -7,13 +7,17 @@
     /** @ngInject */
     function EditCertificationCriteriaController ($log, $uibModal, $uibModalInstance, cert, hasIcs, resources, utilService) {
         var vm = this;
-
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -7,13 +7,17 @@
...
         vm.extendSelect = utilService.extendSelect;
+        vm.isTestToolDisabled = isTestToolDisabled;
         vm.isToolAvailable = isToolAvailable;
         vm.save = save;
-
+        vm.testFunctionalityOnChange = testFunctionalityOnChange;
+        vm.testProceduresOnChange = testProceduresOnChange;
+        vm.testToolsOnChange = testToolsOnChange;
+        vm.testDataOnChange = testDataOnChange;
+        vm.testStandardOnChange = testStandardOnChange;
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/util/expandableList.component.js
```diff
@@ -0,0 +1,184 @@
...
+
+    function ExapandableListController () {
+        var ctrl = this;
+        ctrl.selectOnChange = selectOnChange;
+        ctrl.addItemToListClick = addItemToListClick;
+        ctrl.additionalInputChange = additionalInputChange;
+        ctrl.additionalInput2Change = additionalInput2Change;
+        ctrl.cancelAddItemToListClick = cancelAddItemToListClick;
+        ctrl.isItemDisabled = isItemDisabled;
+        ctrl.removeItem = removeItem;
```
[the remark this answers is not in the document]
tmy1313: Alphabetized...  
The sample I saw for how `$onInit` works, used `this`.  Changed it to `ctrl`.

src/app/components/util/expandableList.component.js
```diff
@@ -0,0 +1,184 @@
...
+            var index = ctrl.selectedItems.indexOf(item);
+            if (index > -1) {
+                ctrl.selectedItems.splice(index, 1);
+            }
+            var onChangeObject = _getOnChangeObject('Remove', item)
+            ctrl.onChange({'action': onChangeObject});  //This is what makes the method binding work
+        }
+
+        function selectOnChange () {
+            if (ctrl.selectedItem[ctrl.itemKey] === -99) {
```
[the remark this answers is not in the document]
tmy1313: Addressed - used your `undefined` suggestion.

src/app/components/util/expandableList.component.js
```diff
@@ -0,0 +1,184 @@
...
+                }
+                ctrl.selectedItem = '';
+            }
+        }
+
+        /////////////////////////////////////////////////////////////////
+
+        function _createSelectedItem (item) {
+            if (ctrl.additionalInput) {
+                return {'item': item, 'additionalInputValue': ''};
```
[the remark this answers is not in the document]
tmy1313: Addressed - fixed.

src/app/components/util/expandableList.component.js
```diff
@@ -0,0 +1,184 @@
...
+            angular.forEach(ctrl._selectedItems, function (arrayItem) {
+                if (arrayItem.item[ctrl.itemKey] === item[ctrl.itemKey]) {
+                    retValue = true;
+                }
+            });
+            return retValue;
+        }
+
+        function _getItemByKey (key) {
+            //Should be replaced with array.filter...
```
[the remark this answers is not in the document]
tmy1313: Addressed

src/app/components/listing_details/criteriaModal.controller.js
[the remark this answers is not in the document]
andlar: From a TDD perspective, one should write only the minimum amount of code required to make the tests pass, and adding a new layer of functions is a step farther than necessary. I'd lean towards simply renaming the function now, and if we ever do come up with a reason to add a way to disable it then we should expand it in the way you suggest.

src/app/components/util/expandableList.html
```diff
@@ -0,0 +1,77 @@
+<div ng-form="{{ $ctrl.identifier }}_form">
```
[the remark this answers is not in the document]
tmy1313: I just stuck with '$ctrl' since it seemed to be the standard in the examples I was looking at.  I have no real feeling one way or another.

src/app/components/listing_details/criteriaModal.controller.js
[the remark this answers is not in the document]
andlar: I didn't mean to suggest that the controls share callback functions; each of the select-case-add things above should create a new instance of the unique class that the testThing represents, and then add that. This function, as written, is basically a constructor for a TestData class already

src/app/components/util/expandableList.html
```diff
@@ -0,0 +1,77 @@
...
+    <div ng-repeat="option in $ctrl.selectedItems track by $index">
+        <div class="selected-item alert-success fade-in">
+            <button class="closebtn" ng-click="$ctrl.removeItem(option)">
```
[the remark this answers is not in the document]
tmy1313: Changed to the Bootstrap class `close`.  This gives the proper upper right hand corner placement.

src/app/components/util/expandableList.component.js
```diff
@@ -0,0 +1,184 @@
...
+        templateUrl: './app/components/util/expandableList.html',
+        controller: ExapandableListController,
+        bindings: {
+            addItems: '@',                      //Boolean value indicating if the control has the functionality to add new items to the drop down.  When this is true, an extra item will appear in the list and when the user selects that item, they will be presented with a text box to type in the new value.
+            addItemsOptionText: '@',            //Text for the additional option
+            addItemsPlaceholder: '@',           //Direction that appear in the text box
+            additionalInput: '@',               //Boolean value indicating whether the control renders an additioanl text box for data associated with the selected item.
+            additionalInputLabel: '@',          //String that is the label for the additional input text box.
+            additionalInputRequired: '@',       //Boolean indicating if the addiational input is required.
+            additionalInputMaxLength: '@',      //Number indicating how many characters can be entered in the additional input text box
```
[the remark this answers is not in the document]
tmy1313: It is being set in criteriaModal.html, and being consumed in expandableList.html.

src/app/components/util/expandableList.html
tmy1313: Addressed

src/app/components/util/expandableList.html
tmy1313: Addressed

src/app/components/util/expandableList.component.js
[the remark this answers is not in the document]
andlar: Ah, okay. I couldn't find more than two `additionalInputMaxLength` but forgot about the `additional-input-max-length` that is needed in the HTML

src/app/components/listing_details/criteriaModal.html
tmy1313: The error does feed up.  The message does not.  Currently, on the modal, the errors do not display with any messages near the text box with the errors.  There is also no summary of error messages on the page.  The submit button does become disabled when there is an error in the component.

src/app/index.scss
tmy1313: I'll remove the class, and if we're not happy with the spacing, we can re-evaluate.

src/app/components/listing_details/criteriaModal.html
```diff
@@ -63,32 +63,21 @@ <h4 class="modal-title" id="modalLabel">Certification Criteria: {{ vm.cert.numbe
...
-                            ng-options="option.name as option.name for option in vm.resources.testStandards.data | orderBy: 'name'">
-                    </select>
-                    <div class="input-group-btn"><button class="btn btn-ai-success form-control" ng-click="vm.addNewValue(vm.cert.testStandards, vm.newTestStandard); vm.newTestStandard = {}"><i class="fa fa-plus"></i> <span class="sr-only">Add item</span></button></div>
-                  </div>
-                </div>
-                <label class="col-sm-6 control-label" for="newTestStandardExpandableName">Add new Test Standard to available options</label>
-                <div class="col-sm-6">
-                  <input type="text" class="form-control" ng-model="vm.newTestStandard.expandableName" id="newTestStandardExpandableName" ng-change="vm.extendSelect(vm.resources.testStandards.data, vm.newTestStandard.expandableName); vm.newTestStandard.testStandardName = vm.newTestStandard.expandableName"></input>
+                <label class="col-sm-6 control-label" for="testStandard">Test Standard</label>
```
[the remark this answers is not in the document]
tmy1313: In the component, the select box gets named/id'd based on the 'identifier' passed into the component.  Therefore, if the label's 'for' = component 'identifier', this should be correct.

src/app/components/listing_details/criteriaModal.controller.js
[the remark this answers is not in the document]
tmy1313: Changes made based on our Teams discussion.

src/app/components/util/expandableList.component.js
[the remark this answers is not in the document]
andlar: It looks like selectOnChange is the only one not alphabetized now

src/app/components/util/expandableList.component.spec.js
```diff
@@ -0,0 +1,245 @@
...
+            expect(ctrl.selectedItems.length).toBe(1);
+        });
+        it('should call the onChnage callback', function () {
+            expect(ctrl.onChange).toHaveBeenCalled();
+        });
+    }); 
+    
+    describe('on an item being selected', function () {
+        it('should should but the control in add mode', function () {
+
```
andlar: Missing something here

src/app/index.scss
```diff
@@ -643,6 +643,39 @@ option[disabled] {
     width: 100%;
     background: #ccc;
 }
+
+
+.selected-item {
+    padding: 5px;
+    margin-bottom: 5px;
+}
```
andlar: I thought you were going to remove the pixels?

src/app/index.scss
```diff
@@ -643,6 +643,39 @@ option[disabled] {
...
+    color: #777777;
+}
+.additional-input::-moz-placeholder { /* Firefox 19+ */
+    color: #777777;
+}
+.additional-input:-ms-input-placeholder { /* IE 10+ */
+    color: #777777;
+}
+.additional-input:-moz-placeholder { /* Firefox 18- */
+    color: #777777;
```
andlar: We had to do stuff with our placeholder things to make them 508 compatible. The default placeholder text doesn't have enough contrast to satisfy the requirements. I know this component is only in the back end now, but it'll likely be usable by the Developer pages at some point, and then it will have to be 508 compatible. I don't know if what you have here will pass or not, and can't check without loading it and running a tool of some sort

src/app/shared/certificationResultTestData.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
andlar: All of our other AngularJS pages are wrapped in an IIFE. I'm guessing we'll move them out when we go to Angular, but until then should keep the convention in place

src/app/index.scss
tmy1313: These ones were necessary to prevent the selected items from overlapping each other.

src/app/shared/certificationResultTestData.js
tmy1313: Addressed

src/app/index.scss
tmy1313: Worked through this.  Don't love the standard padding between the selected item, but it works.

src/app/components/util/expandableList.component.spec.js
tmy1313: addressed

PR #305, opened 2018-07-16
Ocd 2353 - nonconformity charts

brianlindsy:


src/app/charts/charts.controller.js
```diff
@@ -258,6 +260,139 @@
...
+                    type: 'ColumnChart',
+                    data: {
+                        cols: [
+                            { label: 'Certification Criteria and Program Requirements Surveilled', type: 'string'},
+                            { label: 'Number of Non-Conformities', type: 'number'},
+                        ],
+                        rows: _getNonconformityCountDataInChartFormat(data, 'all'),
+                    },
+                    options: {
+                            animation: {
```
andlar: Something is off with the indentation here

src/app/charts/charts.controller.js
```diff
@@ -258,6 +260,139 @@
...
+                        ],
+                        rows: _getNonconformityCountDataInChartFormat(data, 2014),
+                    },
+                    options: {
+                        animation: {
+                            duration: 1000,
+                            easing: 'inAndOut',
+                            startup: true,
+                        },
+                        title: 'Number of Non-Conformities by Certification Criteria and Program Requirements Surveilled',
```
andlar: Should have an indication in the Title or Column label that this is just 2014 criteria. Should do something similar for the 2015 and program versions

src/app/charts/charts.controller.js
```diff
@@ -258,6 +260,139 @@
...
+                        hAxis: {
+                            title: 'Certification Criteria and Program Requirements Surveilled',
+                            minValue: 0,
+                        },
+                        vAxis: {
+                            scaleType: 'mirrorLog',
```
andlar: Given the small values on this chart, and on the program one, suggest not using log scale for either

src/app/charts/charts.controller.js
```diff
@@ -258,6 +260,139 @@
...
+                        },
+                    },
+                },
+            }
+            });
+        }
+
+        function _getNonconformityCountDataInChartFormat (data, type) {
+            return data.nonconformityStatisticsResult.filter(function (obj) {
+                if (type === 2014) {
```
andlar: This many if-elses ought to be a select type { case 2014: } thing

src/app/charts/charts.controller.js
```diff
@@ -258,6 +260,139 @@
...
+                    return obj.nonconformityType.indexOf('170.314') >= 0;
+                } else if (type === 2015) {
+                    return obj.nonconformityType.indexOf('170.315') >= 0;
+                } else if (type === 'program') {
+                    return obj.nonconformityType.indexOf('170.523') >= 0;
+                } else {
+                    return obj.nonconformityType.indexOf('170.') >= 0;
```
andlar: Is this for the "all" one? What other values can show up here that would be filtered out? If none, should just return true

src/app/charts/charts.controller.spec.js
```diff
@@ -145,6 +173,10 @@
                 expect(vm.sedParticipantCounts.data.rows.length).toBe(mock.sedParticipantStatisticsCounts.length);
             });
 
+            it('should load the nonconformity count statistics', function () {
+                expect(networkService.getNonconformityStatisticsCount).toHaveBeenCalled();
+            });
+
```
andlar: Good start, and you have fake data, but there aren't any tests that check on either your filter or on your data transformation. Should have at least:
* it should filter to 2014 criteria (expect data.count.length to be whatever your mock data count of 2014 criteria is)
* it should filter to 2015 criteria
* it should filter to program criteria
* it should not filter on "all"
* it should format the data (after the filter, expect the first data point in the chart array to have the right format)

src/app/common/network.service.js
```diff
@@ -405,6 +406,10 @@
             return apiGET('/meaningful_use/accurate_as_of');
         }
 
+        function getNonconformityStatisticsCount () {
```
andlar: Needs a test for this one

src/app/charts/charts.html
```diff
@@ -94,5 +95,21 @@ <h1>Charts</h1>
         </div>
       </div>
     </div>
+    <div ng-if="vm.chartState.tab === 'nonconformity'">
+      <div class="row">
+        <div class="col-sm-12">
+          <ul class="nav nav-tabs">
+            <li role="presentation" ng-class="{'active': vm.chartState.nonconformityType === 'all'}"><a ng-click="vm.chartState.nonconformityType = 'all'">All Criteria</a></li>
```
andlar: The Feature was to have the charts in a single screen with a filter of some sort. Should be able to mimic the functionality at the bottom of the "Developer charts" area

PENDING_RELEASE_NOTES.md
```diff
@@ -5,5 +5,6 @@ _Date TBD_
 ### New Features
 * In the Edit Criteria modal, modified how the user selects multiple values to associate to a criteria
 * Removed 'Product wide Relied Upon Software' for new 2014 and 2015 listings
+* Add nonconformity charts to charts page
```
andlar: STG is frozen, so this new Feature needs to be in the PENDING_PENDING_RELEASE_NOTES file

src/app/charts/charts.controller.js
brianlindsy: There's another type of nonconformiy called "Other Nonconformity".

andlar: Okay. Did you check with the product owner to see if we want to display stats for that type of non-conformity too?

PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,9 @@
+
+## Version 9.24.0
```
andlar: This was already deployed. Should not be PENDING any more

src/app/charts/charts.html
```diff
@@ -94,5 +95,22 @@ <h1>Charts</h1>
         </div>
       </div>
     </div>
+    <div ng-if="vm.chartState.tab === 'nonconformity'">
+      <div class="row">
+        <div class="col-sm-12">
+          <div google-chart="" chart="vm.nonconformityCounts['' + vm.chartState.nonconformityCountType]" class="column-chart"></div>
```
andlar: Does this work? Specifically the part in the array of nonconformityCounts? It looks off with two single quotes on one side and nothing on the other

src/app/common/util.service.js
```diff
@@ -204,6 +205,13 @@
             return sortCert(type);
         }
 
+        function sortOtherNonconformityTypes (type) {
```
andlar: Should have a test for this function

src/app/charts/charts.html
brianlindsy: I took that line from lines 39 and 42 in the same file and it does work locally

PR #307, opened 2018-07-23
Ocd 2275 - Migrate Summary Statistics 

tmy1313:


src/app/admin/components/schedules/schedule.controller.js
```diff
@@ -47,6 +50,18 @@
                     vm.errorMessage = error.data.error;
                 });
         }
+        function onScheduleChange() {
```
andlar: Why does this function do the same thing as in the activate function, but doesn't use the same object? If we need to change configuration we'll need to do it in two places now, right?

tmy1313: Addressed.

PR #310, opened 2018-07-26
Ocd 2385 - update access to certain reports

brianlindsy:


PENDING_OOC_RELEASE_NOTES.md
```diff
@@ -0,0 +1,9 @@
+
+## Version TBD
```
kekey1: I think you need to specify an actual version number here

PR #318, opened 2018-08-24
Ocd 2218 - ui enhancement

brianlindsy:


src/app/charts/charts.html
```diff
@@ -37,6 +37,7 @@ <h1>Charts</h1>
           <div google-chart="" chart="chart" class="pie-chart"></div>
         </div>
       </div>
+      The 2011 edition has been retired and has no "Active" listings.
```
andlar: For proper spacing, formatting, and responsiveness, this text (and all other text in the file) should be inside a 
```
<div class="row">
  <div class="col-sm-12">
    <p> set of tags - text here </p>
  </div>
</div>
```
I don't know if I missed that on earlier PRs, or put them in incorrectly in the first place, but given you're in this file already they should be updated.

src/app/charts/charts.html
```diff
@@ -23,20 +23,27 @@ <h1>Charts</h1>
...
       </div>
     </div>
     <div ng-if="vm.chartState.tab === 'developer'">
-      Since the start of the ONC Health IT Certification Program, there have been three certification editions: 2011, 2014, and 2015. The charts below describe the transition between each edition by certified health IT developers. The regulations for each edition can be found <a href ="https://www.healthit.gov/topic/certification-ehrs/certification-standards-and-regulations">here</a>. 
       <div class="row">
-        <div class="col-sm-4" ng-repeat="chart in vm.incumbentDevelopersCounts">
-          <div google-chart="" chart="chart" class="pie-chart"></div>
+  		<div class="col-sm-12">
+      		<p>Since the start of the ONC Health IT Certification Program, there have been three certification editions: 2011, 2014, and 2015. The charts below describe the transition between each edition by certified health IT developers. The regulations for each edition can be found <a href ="https://www.healthit.gov/topic/certification-ehrs/certification-standards-and-regulations">here</a>.</p> 
+        	<div class="col-sm-4" ng-repeat="chart in vm.incumbentDevelopersCounts">
```
andlar: These should be in a row of their own. Each "class=row" should have "col-SIZE-columnNumbers"s where the sum of the numbers is 12. In this case, this row has 24 (12 with the paragraph at the top, then 3 sets of 4 for the charts)

PR #319, opened 2018-08-28
Ocd 794 - User/Contact name changes

andlar:


src/app/admin/components/schedules/schedules.directive.spec.js
```diff
@@ -7,7 +7,7 @@
         mock = {
             acbs: [{id: 1, name: 'fake'}],
             results: [{
-                name: 'tmy1313@gmail_com',
```
tmy1313: Thanks!

PR #321, opened 2018-08-28
Ocd 2409 Add Reason for developer ban

kekey1:


src/app/admin/components/certifiedProduct/developer/edit.controller.js
```diff
@@ -87,6 +88,16 @@
             return false;
         }
 
+        function isMissingReasonForBan () {
+            for (var i = 0; i < vm.developer.statusEvents.length; i++) {
+                if(vm.developer.statusEvents[i].status.status === 'Under certification ban by ONC' &&
```
tmy1313: Should this use and "id", rather than the string "Under certification ban by ONC"?

kekey1: Hmm.. I don't know, other places in this same controller use the status name strings so I was just going with it.

tmy1313: No problem... :)

PR #325, opened 2018-08-31
Ocd 1592 - add info messages to cms id widget

brianlindsy:


src/app/components/cms_widget/cms_widget_display.html
```diff
@@ -17,6 +17,22 @@
       <tr ng-if="widget.searchResult.year === '2014'"><th scope="row">Ambulatory CQMs</th><td class="text-right">{{ widget.searchResult.metPercentages.cqmsAmbulatory }}%</td></tr>
     </table>
   </div>
+  <div ng-if="widget.searchResult.missingAnd.length">
+    Please select a product or products that contain the following criteria:
+    <div>
```
andlar: Should be `<ul>` instead of `<div>`

src/app/components/cms_widget/cms_widget_display.html
```diff
@@ -17,6 +17,22 @@
...
+      <li ng-repeat="and in widget.searchResult.missingAnd | orderBy: vm.sortCert">
```
andlar: Would prefer not using the word `and` as a variable name. I know it's not reserved in javascript, but it is in other languages. Similar comment re: the `or` below

src/app/components/cms_widget/cms_widget_display.html
```diff
@@ -17,6 +17,22 @@
...
+        {{ and }}
+      </li>
+    </div>
+  </div>
+  <div ng-if="widget.searchResult.missingOr.length">
+    <span ng-if="widget.searchResult.missingAnd.length > 0">In addition, a</span><span ng-if="widget.searchResult.missingAnd.length === 0">Please select a</span> product with at least 1 criteria from the following categor<span ng-if="widget.searchResult.missingOr.length === 1">y</span><span ng-if="widget.searchResult.missingOr.length > 1">ies</span>:
```
andlar: Should this have "a product `or products`" with the or part added? I'd change it to "groups" instead of "categories" too, unless you explain what the categories are somewhere.

src/app/components/cms_widget/cms_widget_display.html
```diff
@@ -17,6 +17,22 @@
...
+    <div>
```
andlar: Also should be `<ul>` Could use `class="list-unstyled"` if you don't want the bullets to appear. These need to be in the `<ul>` elements for 508 reasons

src/app/components/cms_widget/cms_widget_display.html
```diff
@@ -17,6 +17,22 @@
...
+      <li ng-repeat="or in widget.searchResult.missingOr">
+        [<span ng-repeat="criteria in or">{{ criteria }}{{$last ? '' : ', '}}</span>]
```
andlar: Should be sorted. Do like the use of the `$last` ternary though :)

Could also do it with a `.join(", ")` but I think it'd have to be sorted first.

PR #330, opened 2018-09-06
Ocd 1592 - fix ui tests

brianlindsy:


src/app/components/cms_widget/cms_widget.directive.js
```diff
@@ -24,7 +24,7 @@ import 'jspdf-autotable';
         };
     }
     /** @ngInject */
-    function CmsWidgetController ($analytics, $localStorage, $log, utilService, WidgetService) {
+    function CmsWidgetController ($analytics, $localStorage, $log, WidgetService, utilService,) {
```
andlar: Shouldn't have trailing comma

src/app/components/cms_widget/cms_widget.directive.spec.js
```diff
@@ -67,6 +68,15 @@
             });
         });
 
+        describe('adding 2015 products to the list', function () {
+            fit('should show missing criteria', function () {
```
andlar: Shouldn't be `fit` as that stops all the other tests from running

PR #341, opened 2018-09-18
OCD-2431: update scheduler modal for weekly jobs

kekey1:


src/app/admin/components/schedules/schedule.controller.js
```diff
@@ -56,6 +56,7 @@
 
         function onScheduleChange () {
             vm.schConfig = _getScheduleConfig();
+            vm.trigger.cronSchedule = _getDefaultCron();
```
andlar: Does this reset the cron if someone changes the tab? I'm not sure about this. What happens if a user is on a DAILY trigger editing thing and switch to the WEEKLY tab? Will it set the cron schedule to the default DAILY value?

andlar: Oh, nevermind. This is when a user switches between types of Jobs. Definitely makes sense then

PR #342, opened 2018-09-18
OCD-1600 - add missingxOr info message for 2014 cms id wigdet

brianlindsy:


andlar:
Was this the PR where you were going to widen the widget box to reduce that "can't scroll" problem? 

Also, would recommend putting a space after the comma in the "join" command, as that might let it wrap correctly, but that's not a deal-breaker

PR #346, opened 2018-09-20
Remove password validation check; display strength values from API

andlar:
[#OCD-1310]

Needs to be deployed so QA can check the API strength stuff. Will be enhanced later with front-end password checking, hence no RELEASE NOTES

### DIFF
--- a/src/app/admin/components/login/login.directive.spec.js
+++ b/src/app/admin/components/login/login.directive.spec.js
@@ -2,11 +2,25 @@
     'use strict';
 
     describe('the Login', function () {
-        var $compile, $log, $q, Idle, Keepalive, authService, el, networkService, scope, vm;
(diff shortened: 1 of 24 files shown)


src/app/admin/components/login/login.directive.spec.js
```diff
@@ -179,6 +178,53 @@
...
+                        networkService.changePassword.and.returnValue($q.when({
+                            passwordUpdated: false,
+                            warning: 'a warning',
+                        }));
+                        vm.changePassword();
+                        scope.$digest();
+                        expect(vm.message).toBe('Your password was not changed. a warning');
+                    });
+
+                    it('wshould report a suggestion', function () {
```
tmy1313: Is 'wshould' a typo?  There are a couple occurrences.

andlar: Yep. Fixed it in current work.

PR #348, opened 2018-09-24
Ocd 2377-2  New PR after rebase

tmy1313:


src/app/registration/components/apiKey.html
```diff
@@ -39,6 +41,8 @@
             <td>{{ user.name }}</td>
             <td>{{ user.email }}</td>
             <td>{{ user.key }}</td>
+            <td>{{ user.lastUsedDate | date : 'MMM d, y h:mm:ss a' }}</td>
```
andlar: Is there a reason you're not using the `'medium'` localizable string? Like `... | date : 'medium' }}` ?

tmy1313: Didn't realize there were pre-defined formats.  Updated.

PR #353, opened 2018-10-03
Fix issue with tab titles after a user is logged out due to timeout

tmy1313:
OCD-1276

### DIFF
--- a/src/app/index.config.js
+++ b/src/app/index.config.js
@@ -8,7 +8,7 @@
         .config(config);
 
     /** @ngInject */
-    function config ($analyticsProvider, $locationProvider, $logProvider, stConfig) {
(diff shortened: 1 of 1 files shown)


src/app/index.config.js
```diff
@@ -8,7 +8,7 @@
         .config(config);
 
     /** @ngInject */
-    function config ($analyticsProvider, $locationProvider, $logProvider, stConfig) {
+    function config ($analyticsProvider, $locationProvider, $logProvider, TitleProvider, stConfig ) {
```
andlar: Is there an extra space after the "stConfig" item?

tmy1313: Yep - I'll fix.

PR #367, opened 2018-10-25
Fix error when username is too long

tmy1313:
OCD-2502


PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -4,5 +4,6 @@ _Date TBD_
 
 ### Bugs Fixed
 * Stop displaying QMS changes when there aren't any
+* Show proper error message when username is more than 25 characters and allow the user to proceed
```
andlar: Nothing in this PR has anything to do with the username. Maybe something like "correctly show errors on user creation, and allow user to proceed"?

src/app/registration/create.controller.js
```diff
@@ -60,14 +60,16 @@ window.zxcvbn = zxcvbn;
                     .then(function () {
                         vm.message.value = 'Your account has been created. Please check your email to confirm your account';
                         vm.userDetails = {user: {}};
-                        vm.createUserForm.$setPristine();
-                        vm.createUserForm.$setUntouched();
+                        vm.isCreateUser = false;
```
andlar: By setting this back to false does that then make the view go to the log in thing? Would suggest instead changing the entire state to be something about "success" and then hiding both the "log in" and "create user" form fields.

Probably should just change the entire thing to have a "current state" value that'd be one of "log in", "create user", "user created" at the end of this story

src/app/registration/create.controller.js
```diff
@@ -102,5 +104,16 @@ window.zxcvbn = zxcvbn;
             if (vm.userDetails.user.phoneNumber) { vals.push(vm.userDetails.user.phoneNumber); }
             vm.extras = vals;
         }
+
+        ////////////////////////////////////////////////////////////////////
+
+        function _isJSON (str) {
```
kekey1: Not sure how the code is all organized in the website but could/should this method go into some util-type file? Is there such a thing? Seems like it could be more universally useful.

andlar: It's been in the back of my mind, but I'm also a little concerned that the API is going to return something in a different format than the way the rest of the API does, or this method wouldn't be needed

src/app/registration/create-user.html
```diff
@@ -11,13 +11,16 @@ <h1>User Registration</h1>
           <p>Welcome to the Open Data CHPL. You've been invited to be an Administrator on the ONC's Open Data CHPL, which will allow you to manage certified product listings on the CHPL. Please use the form below to create an account, or log in to your existing account.</p>
         </div>
       </div>
-      <div ng-if="vm.message.value.length > 0" class="row">
-        <div class="col-md-6 col-md-offset-3">
-          <p class="lead" ng-class="vm.message.success ? 'bg-success' : 'bg-danger'">{{ vm.message.value }}</p>
+      <div ng-show="vm.isCreateAccountSuccessMode()" class="row">
```
andlar: Everywhere else in the UI we use ng-if in favor of ng-show, based in part on this article: https://www.codelord.net/2015/07/28/angular-performance-ng-show-vs-ng-if/ It's probably a negligible impact, but using ng-show does buck our conventions. Do you _need_ ng-show here for some reason?

src/app/registration/create-user.html
```diff
@@ -11,13 +11,16 @@ <h1>User Registration</h1>
...
+        <div class="col-md-8 col-md-offset-2">
+          <p class="alert alert-success" ng-if="vm.message.success">{{ vm.message.value }}</p>
```
andlar: By making the `<p>` only appear occasionally, you're leaving the parent `<div>` in the DOM in some cases, even though there's nothing in it. Suggest moving the if up one level. Maybe even two levels, because the parent's parent would also be empty then.

src/app/registration/create-user.html
andlar: Everywhere else in the UI we use ng-if in favor of ng-show, based in part on this article: https://www.codelord.net/2015/07/28/angular-performance-ng-show-vs-ng-if/ It's probably a negligible impact, but using ng-show does buck our conventions. Do you need ng-show here for some reason?

src/app/registration/create-user.html
```diff
@@ -11,13 +11,16 @@ <h1>User Registration</h1>
...
         </div>
       </div>
-      <div class="row">
-        <div class="col-md-12 form-horizontal" ng-form="vm.createUserForm" ng-show="vm.isCreateUser">
+      <div class="row" ng-show="vm.isCreateAccountMode()">
+        <div class="col-md-8 col-md-offset-2">
+            <p class="alert alert-danger" ng-if="!vm.message.success" ng-repeat="err in vm.message.value">{{ err }}</p>
```
andlar: Similar to above. Also, what you have here in the `vm.message.value` is a list of errors, right? Semantically, that should be in repeated `<li>` items, probably under an `<ul>` as the order doesn't matter. Should surround the entire list with a `<p>` and some text like: "There were errors with your submission"

src/app/registration/create-user.html
```diff
@@ -64,17 +67,19 @@ <h1>User Registration</h1>
           <button class="btn btn-default btn-block"
                   ng-mouseover="vm.showFormErrors = true; vm.setExtras()"
                   ng-click="vm.createUser()" ng-disabled="vm.createUserForm.$invalid && vm.showFormErrors">Create account</button>
-          <p>Or <a href="" ng-click="vm.isCreateUser = ! vm.isCreateUser">sign into your existing account</a></p>
+          <p>Or <a href="" ng-click="vm.changeDisplayMode('SIGN-IN')">sign into your existing account</a></p>
           <p>If you require accessibility assistance, please contact Dustin Charles at <a href="mailto:onc_chpl@hhs.gov">onc_chpl@hhs.gov</a></p>
         </div>
-        <div class="col-md-12" ng-form="vm.authorizeUserForm" ng-show="!vm.isCreateUser">
+      </div>
+      <div class="row" ng-show="vm.isSignInMode()">
```
andlar: Why the ng-show here?

src/app/registration/create-user.html
```diff
@@ -64,17 +67,19 @@ <h1>User Registration</h1>
           <button class="btn btn-default btn-block"
                   ng-mouseover="vm.showFormErrors = true; vm.setExtras()"
                   ng-click="vm.createUser()" ng-disabled="vm.createUserForm.$invalid && vm.showFormErrors">Create account</button>
-          <p>Or <a href="" ng-click="vm.isCreateUser = ! vm.isCreateUser">sign into your existing account</a></p>
+          <p>Or <a href="" ng-click="vm.changeDisplayMode('SIGN-IN')">sign into your existing account</a></p>
```
andlar: I've been trying to convert all `<a>` links that don't actually act as links to buttons wherever I run across them. You can have it still look like a link with something like `<button class="btn btn-link">`. Not a required change

PR #371, opened 2018-11-02
Apply maxlength validator to username

tmy1313:
OCD-2502


tmy1313:
This is going to STG.  This was the two part ticket.  The first part of the ticket is already on STG.

PR #373, opened 2018-11-05
OCD-2544: update ONC contact information

kekey1:


src/app/charts/charts.html
```diff
@@ -7,7 +7,7 @@ <h1>Charts</h1>
   <div class="main-content" id="main-content" tabindex="-1">
     <div class="row">
       <div class="col-sm-12">
-        <p>This is a beta release for charts in the CHPL. We will be working on iterating these charts over the coming months. Please send any questions or comments to <a href="mailto:ONC_CHPL@hhs.gov">ONC_CHPL@hhs.gov</a></p>
+        <p>This is a beta release for charts in the CHPL. We will be working on iterating these charts over the coming months. Please use the <a href="https://www.healthit.gov/form/healthit-feedback-form" target="_blank">Health IT Feedback Form</a> for questions or comments.</p>
```
andlar: Is there a requirement that this open in a new window? Nowhere else in the code do we open a new window on links

kekey1: No.. I thought it was pretty much an accepted standard that if you link someplace other than in your own website you'd do it in a new window (or new tab, as per your browser's configuration... mine opens in a new tab).

andlar: I really hope that it's not a standard. I can't stand it when links open in new tabs without me telling them to. I've done some googling, and there are a few design articles as well as some w3.org 508 accessibility guidelines that suggest not using the target=_blank unless absolutely necessary, and, even when doing so, informing the user before they go

kekey1: Ha that's funny because I don't mind opening new tabs but I really hate when a link navigates me away from a site in the same tab. I'll take out the target=_blanks.

PR #375, opened 2018-11-05
Display username too long error at bottom of page

tmy1313:
OCD-2502

### DIFF
--- a/src/app/registration/create-user.html
+++ b/src/app/registration/create-user.html
@@ -59,7 +59,7 @@
             The following error(s) must be fixed:
             <ul>
               <li ng-if="vm.createUserForm.username.$error.required">Username is required</li>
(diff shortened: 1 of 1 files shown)


src/app/registration/create-user.html
```diff
@@ -59,6 +59,7 @@ <h1>User Registration</h1>
             The following error(s) must be fixed:
             <ul>
               <li ng-if="vm.createUserForm.username.$error.required">Username is required</li>
+              <li ng-if="vm.createUserForm.username.$error.maxlength">Field is too long</li>
```
andlar: By using the phrase "Field is too long" the user doesn't know what field it is

tmy1313: Updated

PR #376, opened 2018-11-05
Allow DELETE calls with body data

andlar:
[#OCD-2277]


andlar:
Thanks for the catch. Yes, this is for STG

PR #378, opened 2018-11-12
Speed up UI by caching some network calls for a short time

andlar:
* Cache three network calls for varying timeouts
  * /collections/certified_products for five minutes
  * /certified_products/{productId}/details for fifteen minutes
  * /data/search_options for five minutes


src/app/services/network.service.js
```diff
@@ -300,7 +322,14 @@ export class NetworkService {
     }
 
     getProduct (productId) {
-        return this.apiGET('/certified_products/' + productId + '/details');
+        const EXPIRATION_TIME = 15; // in minutes
```
kekey1: Does the UI have the equivalent of a properties file for numbers like this that we might want to make configurable? Or is this already the best practice way of doing it?

andlar: There is a place where there are some initialization values, but it only has a couple in them now, and given that I'm not positive this change is going to stick, and that I still want to re-do the way these are "pseudo-memoized" anyway, I thought it made more sense for all three of these values to be in the single file where they're used for now

PR #379, opened 2018-11-14
2015 functionality tested should be restricted by criteria

tmy1313:
OCD-2532


src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -44,6 +44,8 @@
             vm.selectedTestProcedureKeys = _getSelectedTestProcedureKeys();
             vm.selectedTestStandardKeys = _getSelectedTestStandardKeys();
             vm.selectedTestToolKeys = _getSelectedTestToolKeys();
+
+            _sortTestFunctionalities();
```
andlar: One of the things we're trying to do with the conversion to components / prep for Angular move is work to make each of our current controllers/directives pure-ish, as in, data passed to them shouldn't be changed in the eyes of whatever passed it in. Calling this function sorts some incoming data in place, which changes it in the eye of the calling object

src/app/components/listing_details/criteriaModal.controller.js
```diff
@@ -187,6 +189,18 @@
             return ttKeys;
         }
 
+        function _sortTestFunctionalities () {
+            vm.cert.allowedTestFunctionalities.sort(function (a, b) {
```
andlar: When a sorting function is this simple, I'd rather it either be in the view, using something like
```
ng-options="allowedTestFunctionalities.name as ... | orderBy : 'name'
```
I looked at the view here and see we're handing it off to that multi-pick directive. It'd be nice to enhance that to allow sorting, but if that's not the case then the data could be sorted in one line in the controller with something like
```
vm.cert.sortedTestFunctionalities = $filter('orderby')('name')
```
which would also avoid mutating the passed in testfunctionalities

PENDING_RELEASE_NOTES.md
```diff
@@ -9,5 +9,6 @@ _Date TBD_
   * /certified_products/{productId}/details for fifteen minutes
   * /data/search_options for five minutes
 * Add validation to surveillance "sites passed" and "total sites" for randomized surveillance activities
+* Add sorting to functionality tested
```
andlar: This will need to be in PENDING_PENDING now

PR #387, opened 2018-11-28
Ocd 2575 - API Documentation data

andlar:


src/app/navigation/navigation-bottom.html
```diff
@@ -27,6 +27,7 @@
             <li ng-if="vm.isChplAdmin() || vm.isAcbAdmin()"><a href="#/admin/scheduledJobs">Scheduled Jobs</a></li>
             <li ng-if="vm.isChplAdmin()"><a href="#/admin/apiKeyManagement">API Key Management</a></li>
             <li ng-if="vm.isChplAdmin()"><a href="#/admin/announcementsManagement">Announcements Management</a></li>
+            <li ng-if="vm.isChplAdmin() || vm.isOncStaff()"><a href="#/admin/apiDocumentationManagement">API Documentation Management</a></li>
```
kekey1: I think the API is only allowing ROLE_ADMIN, which role(s) are right?

andlar: The ticket does say just ROLE_ADMIN. I left ONC_STAFF in there because they're probably going to be able to do it too. Probably should take them out of this

PR #402, opened 2019-01-30
Ocd 2634

tmy1313:


src/app/admin/components/schedules/schedules.html
```diff
@@ -43,7 +43,10 @@ <h3>Types of Jobs</h3>
           <td>{{ job.name }}</td>
           <td>{{ job.description }}</td>
           <td>{{ job.jobDataMap.acbSpecific ? 'Yes' : 'No' }}</td>
-          <td><button class="btn btn-ai-success" ng-click="vm.editJob(job)" ng-if="job.jobDataMap.editableJobFields"><i class="fa fa-pencil-square-o"></i><span class="sr-only">Edit job for {{ job.name }}</span></button></td>
+          <td>
+            <button class="btn btn-ai-success" ng-click="vm.editJob(job)" ng-if="job.jobDataMap.editableJobFields"><i class="fa fa-pencil-square-o"></i><span class="sr-only">Edit job for {{ job.name }}</span></button>
+            <button class="btn btn-ai-success" ng-click="vm.createSimpleTrigger(job)" ng-if="job.group == 'systemJobs'"><i class="fa fa-bolt"></i><span class="sr-only">Create trigger for {{ job.name }}</span></button>
```
andlar:
```suggestion
            <button class="btn btn-ai-success" ng-click="vm.createSimpleTrigger(job)" ng-if="job.group === 'systemJobs'"><i class="fa fa-bolt"></i><span class="sr-only">Create trigger for {{ job.name }}</span></button>
```

Don't remember why javascript prefers `===` to `==`, but I know there's some reason for it

src/app/admin/components/schedules/systemJob.controller.js
```diff
@@ -0,0 +1,56 @@
...
+    angular.module('chpl.admin')
+        .controller('SystemJobController', SystemJobController);
+
+    /** @ngInject */
+    function SystemJobController ($interval, $log, $uibModalInstance, SPLIT_PRIMARY, job, networkService) {
+        var vm = this;
+
+        vm.cancel = cancel;
+        vm.save = save;
+        vm.datePickerOpen = false;
```
andlar: This should be in the `activate` function. These assignments are the "public API" of this controller, and indicate what methods are available to be called. Oh, it's also probably not needed, as the datePickerOpen will default to `undefined`, which counts as `false`

src/app/admin/components/schedules/systemJob.controller.js
```diff
@@ -0,0 +1,56 @@
...
+
+        activate();
+
+        ////////////////////////////////////////////////////////////////////
+
+        function activate () {
+            vm.job = angular.copy(job);
+            vm.selectedDateTime = new Date();
```
andlar: Have you tried typing the date / time into the text field directly, without using the date/time pickers? We have to support that, and what usually happens is that someone doing that typing will end up having our dateObject have a string value instead of a Date object in it at that point. That's why a lot of the date editing things have a long that comes from the system that gets converted to an object, and is reconverted back to a long on save. If what you have works without that, then great, but if not please check on it

src/app/services/network.service.js
```diff
@@ -72,6 +72,10 @@ export class NetworkService {
         return this.apiPOST('/users/create', contactDetails);
     }
 
+    createScheduleOneTimeTrigger (trigger) {
+        return this.apiPOST('/schedules/triggers/one_time', trigger);
```
andlar: This should have a test associated with it. Tests for the network services are really easy, but should be there

src/app/admin/components/schedules/systemJob.controller.js
tmy1313: This appears to work correctly when editing the text field directly...  Maybe I'm missing something?

src/app/services/network.service.spec.js
```diff
@@ -209,6 +209,14 @@
             $httpBackend.flush();
         });
 
+        ('should createScheduleOneTimeTrigger', function () {
```
andlar: Did you run this test? This looks like a compilation failure

tmy1313: hmm, ran `yarn test` and didn;t see any issues.  Let me try again.

tmy1313: Though, it does have the wrong URL.

andlar: Comment out the new function you added to the network.service.js file and run the tests. If this test fails then you're in good shape. If nothing fails then your test is broken

tmy1313: Ha!   Commenting out the function didn't cause a failure, but I did find the error - `it`...

PR #403, opened 2019-01-31
fix: display the correct surveillance authority

tmy1313:
OCD-2655

### DIFF
--- a/src/app/admin/components/surveillance/edit.html
+++ b/src/app/admin/components/surveillance/edit.html
@@ -20,7 +20,7 @@
     <div class="row" ng-if="vm.workType !== 'initiate'">
       <div class="col-sm-3"><strong>Surveillance initiated by:</strong></div>
       <div class="col-sm-9">
-        <span>{{ vm.surveillance.authority === 'ROLE_ADMIN' ? 'ONC' : 'ONC-ACB' }}</span>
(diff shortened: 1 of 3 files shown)


src/app/admin/components/surveillance/edit.html
```diff
@@ -20,7 +20,7 @@ <h4 class="modal-title" id="edit-surveillance-label"><span ng-if="vm.workType ==
     <div class="row" ng-if="vm.workType !== 'initiate'">
       <div class="col-sm-3"><strong>Surveillance initiated by:</strong></div>
       <div class="col-sm-9">
-        <span>{{ vm.surveillance.authority === 'ROLE_ADMIN' ? 'ONC' : 'ONC-ACB' }}</span>
+        <span>{{ vm.surveillance.authority === 'ROLE_ADMIN' || vm.surveillance.authority === 'ROLE_ONC' ? 'ONC' : 'ONC-ACB' }}</span>
```
andlar: The surveillance authority can never be "ROLE_ADMIN", right? As long as that's the case I think it should be removed from the conditional
```suggestion
        <span>{{ vm.surveillance.authority === 'ROLE_ONC' ? 'ONC' : 'ONC-ACB' }}</span>
```

tmy1313: I left it there, because I wasn't sure why it was there to begin with.  I'll remove it.

PR #435, opened 2019-03-26
Ocd 2757 - Developer edit

andlar:


src/app/components/developer/developer.component.js
```diff
@@ -45,10 +45,10 @@ export const DeveloperComponent = {
                 this.canEdit = angular.copy(changes.canEdit.currentValue);
             }
             if (changes.canMerge) {
-                this.canMerge = angular.copy(changes.canMerge.currentValue);
+                this.canMerge = angular.copy(changes.canMerge.currentValue) && this.developer.status.status === 'Active';
             }
             if (changes.canSplit) {
-                this.canSplit = angular.copy(changes.canSplit.currentValue);
+                this.canSplit = angular.copy(changes.canSplit.currentValue) && this.developer.status.status === 'Active';
```
kekey1: I have this comment in the API code
// if the user is an ACB then the developer must be Active otherwise the split is not allowed.
// ADMIN and ONC can perform a split no matter the developer's status

andlar: The comment might be there, but I don't think the API actually restricts it. The old UI code doesn't allow splitting of anything under a Developer if the Developer is non-active

PR #436, opened 2019-03-28
OCD 2756 - Developer editing

andlar:


src/app/navigation/navigation.module.js
```diff
@@ -34,6 +34,10 @@
                 if (response.config.url.indexOf(API) === 0) {
                     response.data = parseToken(response.data);
                 }
+                if (response.data.error === 'Invalid authentication token.' && authService.hasAnyRole()) {
```
kekey1: I noticed the API is sending back a success code 200 with this error response text. This is not really for your PR but I think it would be nice to send back an error code instead (4xx?) so you don't have to key off of this error message.

andlar: If this doesn't work, or even if it does, when we get to that ticket in the backlog we can certainly try to make this better

PR #438, opened 2019-03-29
Ocd 2777 - Improve Reports responsiveness - page load - Developer Activity

tmy1313:


src/app/admin/components/reports/reports.developer.component.js
```diff
@@ -0,0 +1,264 @@
...
+            };
+            this.activityRange.startDate.setDate(this.activityRange.endDate.getDate() - this.activityRange.range + 1); // offset to account for inclusion of endDate in range
+            this.filename = 'Reports_' + new Date().getTime() + '.csv';
+        }
+
+        $onChanges () {
+            this.search();
+        }
+
+        compareArray (prev, curr, keys, root, nested, altRoot) {
```
andlar: I know this function is in the reports.listing.component.js file. My plan was, as I moved parts out of the main reports file, to extract any duplicate functions and put them in the ReportService. Could you do that? With this function and any others that are duplicates?

src/app/admin/components/reports/reports.developer.component.js
```diff
@@ -0,0 +1,264 @@
...
+            let that = this;
+            this.networkService.getActivityMetadata('developers', this.dateAdjust(this.activityRange))
+                .then(results => {
+                    that.results = results;
+                    that.prepare(that.results);
+                });
+        }
+
+        validDates () {
+            if (this.isValidDate(this.activityRange.endDate) && this.isValidDate(this.activityRange.startDate)) {
```
andlar: If there was a reason to add this check, can you add it to the reports.listing.component.js file? And/or extract this function to the ReportService?

src/app/admin/components/reports/reports.developer.html
```diff
@@ -0,0 +1,71 @@
...
+    <tbody>
+      <tr ng-if="$ctrl.displayed.length === 0">
+        <td colspan="5">No activity found</td>
+      </tr>
+      <tr ng-repeat="activity in $ctrl.displayed">
+        <td>{{ activity.developerName }}</td>
+        <td>{{ activity.developerCode }}</td>
+        <td>{{ activity.responsibleUser.fullName }}</td>
+        <td>{{ activity.date | date : 'MMM d, y H:mm:ss Z' : 'UTC' }}</td>
+        <td>
```
andlar: This is a fifth column, but there are only four header columns. In the Listing page the link to expand for details is in the same column as the date of the activity. I don't like that for 508 purposes, but it's what the UX person wanted, so I tried to get to that state for now. We're in the admin section, so 508 isn't technically required

src/app/admin/components/reports/reports.html
```diff
@@ -3,74 +3,9 @@
...
-                  <li ng-repeat="str in activity.details | orderBy : 'toString()' track by $index" ng-bind-html="str"></li>
-                </ul>
-              </td>
-              <td>{{ activity.date | date : 'MMM d, y H:mm:ss Z' : 'UTC' }}</td>
-            </tr>
-          </tbody>
-        </table>
-      </div>
+    <div ng-if="$ctrl.workType === 'developers'">
+      <chpl-reports-developers product-id="$ctrl.productId"></chpl-reports-developers>
```
andlar: The product-id is only used to get a Listing's full history, and you don't have a binding to productId in this component anyway

src/app/admin/components/reports/reports.developer.component.js
tmy1313: I did this because the front end was logging errors while the user was editing the activity dates, because dates were not valid as they entered the date.  I'll move it to the ReportService and update reports.listing.component.js.

PR #448, opened 2019-04-08
OCD 2822 - more stuff in the eye

andlar:


src/app/pages/listing/history/history.component.js
```diff
@@ -0,0 +1,315 @@
...
+            this.$location.path('/resources/chpl-api');
+            this.cancel();
+        }
+
+        _interpretActivity (activity) {
+            var curr, prev;
+            activity.change = [];
+            prev = activity.originalData;
+            curr = activity.newData;
+            if (activity.description.startsWith('Updated certified product')) {
```
kekey1: Is any of this activity interpretation stuff duplicated code that's also done on the report pages? If so, could it be moved to a common place?

andlar: Some of it is similar, but I don't know if any of it is really duplicated. Like, for the criteria, on the eye all they want is "added or removed", but in the reports it also shows all of the sub-elements and their changes. 

The history of non-Listing stuff isn't reported on in the same way in the admin reports section either; there they're under the general object, not tied to "which one is applicable for the Listing" as the eye has it.

Things like the MUU & status history doesn't even use the activity reports anymore either, as those explicitly already have saved history.

With surveillance I might be able to use the added/removed checks, but those are wrapped up in all of the surveillance checks, including the details under surveillance

PR #449, opened 2019-04-09
Ocd 2821 - Incorrect permissions for edit/split/merge of developer/product/version

tmy1313:


tmy1313:
> I can't tell from these changes if the logic is correct, but I don't see any reason why they're not. That said, is the "isDeveloperMergeable" function still in use? If not, it should be removed

`isDeveloperMergeable` is still used, so it was left alone.

PR #461, opened 2019-04-22
OCD-2842: Improve banned developer page

kekey1:


src/app/pages/collections/collections.service.js
```diff
@@ -70,12 +70,11 @@
                 dev = {
                     acb: [],
                     decertificationDate: array[i].decertificationDate,
-                    developer: array[i].developer.name,
-                    mainSearch: array[i].developer.name,
-                    status: array[i].developer.status.status,
+                    developer: array[i].developerName,
+                    mainSearch: array[i].developerName,
```
andlar: If the structure of these objects changed, then the tests that reference them will probably need to change. In collections.service.spec.js, the lines from 35 to 44 reference the banned developers that are probably in the src/app/something.mock.js file. That data will probably need to change, which will then cause the tests to fail

PR #465, opened 2019-04-23
Ocd 2755 - Improve Reports responsiveness - save filter

tmy1313:


PENDING_RELEASE_NOTES.md
```diff
@@ -2,8 +2,10 @@
 ## Version 10.18.0
 _Date TBD_
 
-### New feature
+### New features
+* Shrink main header navigation on page scroll
 * Show "history popup" on all Listings
 * Updated security for edit/split/merge of developers, products, and versions
+* Add "Save Search" functionality to admin reports
```
andlar: Is the "Save Search" functionality hidden behind a flag? If so, this should be in the UAT_RELEASE_NOTES.md

src/app/components/smart-table/chplSearch.directive.js
```diff
@@ -0,0 +1,31 @@
+angular.module('chpl.components')
+    .directive('chplSearch', ['$timeout', 'stConfig', function ($timeout, stConfig) {
```
andlar: You should be able to Inject these using either `/** @ngInject */` or `'ngInject'` somehow

andlar:
Specific comments are resolved, but still need to see this without all of the adminNav commits

PENDING_RELEASE_NOTES.md
```diff
@@ -2,13 +2,16 @@
 ## Version 10.18.0
 _Date TBD_
 
-### New Features
+<<<<<<< HEAD
```
andlar: Merge issue

src/app/admin/components/reports/filter.component.js
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        $onInit () {
+            this.filterName = '';
+            this.refreshFilterList();
+        }
+
+        refreshFilterList () {
+            let that = this;
+            this.networkService.getFilters(this.filterTypeId)
```
andlar: If the `filterTypeId` is optional, then you should check that it exists before loading the list `onInit`

src/app/admin/components/reports/filter.component.js
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        saveFilter () {
+            let that = this;
+            let filter = {};
+            filter.filterType = {};
+            filter.filterType.id = this.filterTypeId;
+            filter.filter = JSON.stringify(this.getFilterData());
+            filter.name = this.filterName;
+            this.$log.info(filter);
```
andlar: Debugging? Does it need to be here?

src/app/admin/components/reports/filter.component.js
```diff
@@ -0,0 +1,61 @@
...
+        deleteFilter (filterId) {
+            let that = this;
+            this.networkService.deleteFilter(filterId)
+                .then(() => that.refreshFilterList());
+        }
+    },
+}
+
+angular.module('chpl.admin')
+    .component('aiFilter', FilterComponent);
```
andlar: Trying to move towards using `chpl` as the prefix for components/directives. Not required to change it, but requested

src/app/admin/components/reports/filter.component.spec.js
```diff
@@ -0,0 +1,74 @@
...
+                Mock = _Mock_;
+                networkService = _networkService_;
+
+                networkService.getFilters.and.returnValue($q.when(Mock.developerReportsFilter))
+                networkService.createFilter.and.returnValue($q.when({}));
+                networkService.deleteFilter.and.returnValue($q.when({}));
+
+                scope = $rootScope.$new()
+
+                el = angular.element('<ai-filter filter-type-id="2" on-apply-filter="onApplyFilter(filter)" get-filter-data="createFilterDataObject()"></ai-filter>');
```
andlar: The two functions here should be created as spies, so tests can use them. Something like: `scope.onApplyFilter = jasmine.createSpy('onApplyFilter');` should work. Then, you can assert that the spies are called during tests with something like `expect(scope.onApplyFilter).toHaveBeenCalledWith(some object)`

src/app/index.route.js
```diff
@@ -16,6 +16,11 @@
 
     function routeConfig ($stateProvider, $urlRouterProvider) {
         $stateProvider
+            /*.state('admin', {
+                abstract: true,
+                url: '/admin',
```
andlar: Shouldn't be here

src/app/index.route.js
```diff
@@ -35,6 +40,24 @@
                 controllerAs: 'vm',
                 data: { title: 'Password Reset' },
             })
+            .state('admin.reports', {
+                url: '/reports',
+                component: 'ReportsListingsComponent',
```
andlar: What's this for? The new navigation has the new Reports component for Listings. We shouldn't be adding states to the /admin route now, only removing them as we can

src/app/pages/reports/developers.html
```diff
@@ -1,35 +1,39 @@
...
+        <h3>Developer</h3>
+      </div>
   </div>
+
   <div class="row">
-    <label for="activity-start-upload" class="col-sm-2 control-label">Start Date</label>
-    <div class="col-sm-3">
-      <input type="date" class="form-control" ng-model="$ctrl.activityRange.startDate" name="activityStartUpload" id="activity-start-upload" placeholder="yyyy-MM-dd">
+    <div class="col-sm-6">
+      <ai-filter feature-flag="adminNav" filter-type-id="1" on-apply-filter="$ctrl.onApplyFilter(filter)" get-filter-data="$ctrl.createFilterDataObject()"></ai-filter>
```
andlar: The `filter-type-id` should be a value that the component retrieves from the networkService, not hard coded as a magic number

src/app/pages/reports/listings.html
```diff
@@ -1,42 +1,51 @@
...
   <div class="row">
-    <label for="activity-start-upload" class="col-sm-2 control-label">Start Date</label>
-    <div class="col-sm-3">
-      <input type="date" class="form-control" ng-model="$ctrl.activityRange.startDate" name="activityStartUpload" id="activity-start-upload" placeholder="yyyy-MM-dd">
-    </div>
-    <label for="activity-end-upload" class="col-sm-2 control-label">End Date</label>
-    <div class="col-sm-3">
-      <input type="date" class="form-control" ng-model="$ctrl.activityRange.endDate" name="activityEndUpload" id="activity-end-upload" placeholder="yyyy-MM-dd">
+    <div class="col-sm-6">
+      <ai-filter feature-flag="adminNav" filter-type-id="2" on-apply-filter="$ctrl.onApplyFilter(filter)" get-filter-data="$ctrl.createFilterDataObject()"></ai-filter>
```
andlar: Also shouldn't be hard-coded

src/app/pages/reports/listings.html
```diff
@@ -52,7 +61,7 @@ <h4>{{ acb }} ({{ $ctrl.acbCount(acb) }})</h4>
             <th class="search-header" st-sort="edition" st-skip-natural="true">Certification Edition</th>
             <th class="search-header" st-sort="chplProductNumber" st-skip-natural="true">CHPL Product Number</th>
             <th class="search-header" st-sort="certificationDate" st-skip-natural="true">Certification Date</th>
-            <th class="search-header" st-sort="date" st-skip-natural="true" st-sort-default="reverse">Activity Date</th>
+            <th class="search-header" st-sort="date" st-skip-natural="true" >Activity Date</th>
```
andlar: Why remove the default sort order?

src/app/services/network.service.js
```diff
@@ -59,6 +59,10 @@ export class NetworkService {
         return this.apiPOST('/certification_ids?ids=' + ids.join(','), {});
     }
 
+    createFilter (filter) {
+        return this.apiPOST('/filters', filter);
```
andlar: All three of these should have corresponding tests in the network.service.spec.js file

src/app/admin/components/reports/filter.component.js
```diff
@@ -15,8 +15,14 @@ export const FilterComponent = {
         }
 
         $onInit () {
+            let that = this;
             this.filterName = '';
-            this.refreshFilterList();
+            this.networkService.getFilterTypes()
+                .then(response => {
+                    that.filterTypeId = response.data.find(item => item.name === that.filterTypeName).id;
+                    this.$log.info(that.filterTypeId);
```
andlar: Does this log statement need to stay?

src/app/admin/components/reports/filter.component.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+        describe('view', () => {
+            it('should be compiled', () => {
+                expect(el.html()).not.toEqual(null);
+            });
+        });
+
+        describe('controller', () => {
+            it('should be initialized with filters', () => {
+                expect(ctrl.availableFilters.length).toBe(3);
```
andlar: `filterTypeId` went away, and `filterTypeName` showed up. This test should have failed without changing the element definition in the first `beforeEach`

PR #471, opened 2019-05-01
Ocd 2821

tmy1313:


src/app/admin/components/certifiedProduct/management.html
```diff
@@ -450,10 +450,12 @@ <h3 ng-if="vm.inspectingCp">System developer information</h3>
             <div class="panel-heading">
               <span>Active Product</span>
               <span class="pull-right">
-                <span ng-if="vm.workType !== 'merge' && vm.isDeveloperMergeable(vm.activeDeveloper)">
+                <span ng-if="vm.workType !== 'merge' && vm.isDeveloperEditable(vm.activeDeveloper)">
```
andlar: Is the "isDeveloperMergeable" function still used anywhere?

tmy1313: It is used by the "isProductEditable" function.

PR #477, opened 2019-05-08
Ocd 2885  (OCD-2820 - Resolve phantom logged in issue)

tmy1313:


PENDING_RELEASE_NOTES.md
```diff
@@ -9,6 +9,7 @@ _Date TBD_
 * Updated text and column headings on the Decertified Products and Inactive Certificates pages.
 * Updated language on Charts page.
 * Shrink main header navigation on page scroll
+* Add link to FF4j admin console for ROLE_ADMIN
```
andlar: Needs update from development branch; there shouldn't be anything in PENDING except this line

PENDING_RELEASE_NOTES.md
```diff
@@ -4,6 +4,7 @@ _Date TBD_
 
 ### New features
 * Use new endpoints for better ONC-ACB and ONC-ATL report activity
+* Add link to FF4j admin console for ROLE_ADMIN
```
andlar: Sorry, now I wonder if this should be included at all. the `vm.DEVELOPER_MODE` is false in STG/PROD, which would mean this isn't a feature that is available to an end user. I think I could see it either way, but if we think of the RELEASE_NOTES as showing things that an end user could see, then it maybe should be here. Going to approve this anyway, but up for a conversation

PR #481, opened 2019-05-10
Ocd 2871 - Scheduled Jobs list gets out of sync when deleting

tmy1313:


andlar:
@tmy1313 , if you put this behind a flag you could deploy it immediately, instead of having it waiting for QA to be ready...

PR #489, opened 2019-05-16
OCD-2775- Improve Reports responsiveness - save filter

tmy1313:
fix: update listings report to switch between filters with productIds
fix: refresh filter list when user stops impersonating

### DIFF
--- a/src/app/pages/reports/listings.component.js
+++ b/src/app/pages/reports/listings.component.js
@@ -2,6 +2,7 @@ export const ReportsListingsComponent = {
     templateUrl: 'chpl.reports/listings.html',
     bindings: {
         productId: '<?',
+        filterToApply: '<?',
     },
     controller: class ReportsListings {
         constructor ($filter, $log, $state, $uibModal, ReportService, networkService, utilService) {
@@ -29,6 +30,11 @@ export const ReportsListingsComponent = {
             if (changes.productId && changes.productId.currentValue) {
                 this.productId = angular.copy(changes.productId.currentValue);
             }
+            if (changes.filterToApply && changes.filterToApply.currentValue && changes.filterToApply.currentValue !== changes.filterToApply.previousValue) {
+                this.doFilter(changes.filterToApply.currentValue);
+                return;
+            }
+
             this.search();
         }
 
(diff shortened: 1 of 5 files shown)


src/app/pages/reports/listings.component.js
```diff
@@ -29,6 +30,11 @@ export const ReportsListingsComponent = {
             if (changes.productId && changes.productId.currentValue) {
                 this.productId = angular.copy(changes.productId.currentValue);
             }
+            if (changes.filterToApply && changes.filterToApply.currentValue && changes.filterToApply.currentValue !== changes.filterToApply.previousValue) {
```
andlar: if "filterToApply" is in the changes object, then that means it changed. Also, if filterToApply is an object, you shouldn't use `===` or `!==` to test for equality

src/app/pages/reports/reports.state.js
```diff
@@ -22,9 +22,11 @@ function reportsStateConfig ($stateProvider) {
             component: 'chplReportsListings',
             params: {
                 productId: {squash: true, value: null},
+                filterToApply: {squash: true, value: null},
```
andlar: I don't think this should be here. The things in params are part of the URL

src/app/pages/reports/reports.state.js
```diff
@@ -22,9 +22,11 @@ function reportsStateConfig ($stateProvider) {
...
             },
             resolve: {
                 productId: $transition$ => $transition$.params().productId,
+                filterToApply: $transition$ => $transition$.params().filterToApply,
```
andlar: Can you try this without this line? I'm not sure if it's needed

src/app/pages/reports/reports.state.js
tmy1313: I tried removing each of the references to `filterToApply`, but if they are both not present, the `filterToApply` value is not in the `changes` object.  It seems they re both needed.

src/app/pages/reports/reports.state.js
tmy1313: I tried removing each of the references to `filterToApply`, but if they are both not present, the `filterToApply` value is not in the `changes` object.  It seems they re both needed.

PR #498, opened 2019-05-30
Ocd 2820 - Resolve phantom logged in issue

tmy1313:
Remove ocd2820 flag


src/app/index.run.js
```diff
@@ -42,18 +42,8 @@ import { states as surveillanceStates } from './pages/surveillance/surveillance.
 
         if (authService.hasAnyRole()) {
             networkService.keepalive()
-                .then(response => {
-                    if (featureFlags.isOn('ocd2820')) {
-                        angular.noop;
```
andlar: Does this work without the "then"? I've been trying to research whether promises can exist without the then, but haven't found anything specific. If it works with just the catch, then that's fine, but if the then is required then putting the `angular.noop;` back in might make sense. It might make sense to do that anyway so that we're explicitly saying we're not doing anything on a success

PR #515, opened 2019-06-13
test: fix brokern tests and only run tests wrt to current development

tmy1313:
OCD-2912

### DIFF
--- a/src/app/pages/reports/users/users.component.spec.js
+++ b/src/app/pages/reports/users/users.component.spec.js
@@ -3,10 +3,10 @@
 
     fdescribe('the Reports.Users component', () => {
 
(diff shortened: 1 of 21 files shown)


src/app/pages/reports/users/users.component.spec.js
```diff
@@ -25,6 +26,8 @@
                 networkService.getActivityMetadata.and.returnValue($q.when(Mock.productReportsMetadata));
                 networkService.getActivityById.and.returnValue($q.when(Mock.listingActivity));
 
+                $httpBackend.whenGET(/feature-flags/).respond({});
```
andlar: What is in the users component that's calling the $http service? If it's calling through the network service then that call should be mocked. Individual components shouldn't have a dependency on the $http service in almost all cases

PR #519, opened 2019-06-18
Ocd 2913 - Associate a listing to a complaint

tmy1313:


src/app/components/surveillance/complaint/complaint.component.js
```diff
@@ -22,7 +23,7 @@ export const SurveillanceComplaintComponent = {
                 ADD: 'add',
             }
             this.currentMode = '';
-
+            this.certifiedProduct;
```
andlar: What does this do?

src/app/components/surveillance/complaint/complaint.component.js
```diff
@@ -71,6 +76,39 @@ export const SurveillanceComplaintComponent = {
...
+            }
+            this.complaint.listings.push({
+                listingId: $item.id,
+                chplProductNumber: $item.chplProductNumber,
+            });
+            this.listing = '';
+        }
+
+        removeListing (chplProductNumber) {
+            this.complaint.listings = this.complaint.listings.filter(listing => listing !== chplProductNumber);
```
andlar: Is "chplProductNumber" a string? If so, then this shouldn't work, as "complaint.listings" is an object. If it's not a string then the variable name ought to indicate that somehow

src/app/components/surveillance/complaint/complaint.component.js
```diff
@@ -71,6 +76,39 @@ export const SurveillanceComplaintComponent = {
...
+        startsWith (valueToCheck, viewValue) {
+            return valueToCheck.substr(0, viewValue.length).toLowerCase() === viewValue.toLowerCase();
+        }
+
+        changeAcb () {
+            this.filterListingsBasedOnSelectedAcb();
+        }
+
+        filterListingsBasedOnSelectedAcb () {
+            let that = this;
```
andlar: I don't think you need a "that" here. Could be wrong, but I think I've only needed them when the "this" was referenced in the resolution of a promise, and in this method there aren't any promises

src/app/pages/surveillance/complaints/complaints.component.js
```diff
@@ -123,6 +124,14 @@ export const SurveillanceComplaintsComponent = {
             //get all acbs that the user has edit capability of
             this.networkService.getAcbs(true).then(response => {
                 that.certificationBodies = response.acbs;
+                that.refreshListings();
```
andlar: Why wait to get listings until after getting ACBs? Is one dependent on the other?

src/app/pages/surveillance/complaints/complaints.component.js
```diff
@@ -123,6 +124,14 @@ export const SurveillanceComplaintsComponent = {
...
+            });
+        }
+
+        refreshListings () {
+            let that = this;
+            this.networkService.getAll().then(response => {
```
andlar: Not sure if this is the best call. One of the (or a new) getCollection calls might be better. This call returns all of the data needed on the main search page, but you're only using one or two of those fields. That's a lot of network traffic that could be reduced considerably

PR #535, opened 2019-06-26
Ocd 2912 - Create a basic complaint

tmy1313:
I'm not 100% sure that the date fix is going to work on DEV.

### DIFF
--- a/src/app/pages/surveillance/complaints/complaints.component.js
+++ b/src/app/pages/surveillance/complaints/complaints.component.js
@@ -43,7 +43,6 @@ export const SurveillanceComplaintsComponent = {
         saveComplaint (complaint) {
             if (complaint.formattedReceivedDate) {
                 complaint.receivedDate = this.convertToUTC(complaint.formattedReceivedDate).valueOf();
-                this.$log.info(complaint.receivedDate);
             }
             if (complaint.id) {
                 this.updateComplaint(complaint);
--- a/src/app/pages/surveillance/complaints/complaints.component.js
+++ b/src/app/pages/surveillance/complaints/complaints.component.js
@@ -42,7 +42,8 @@ export const SurveillanceComplaintsComponent = {
 
         saveComplaint (complaint) {
             if (complaint.formattedReceivedDate) {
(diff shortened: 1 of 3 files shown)


src/app/pages/surveillance/complaints/complaints.component.js
```diff
@@ -99,14 +99,18 @@ export const SurveillanceComplaintsComponent = {
                 that.complaints = response.results;
                 that.complaints.forEach(complaint => {
                     if (complaint.receivedDate) {
-                        complaint.formattedReceivedDate = new Date(complaint.receivedDate);
+                        complaint.formattedReceivedDate = this.convertToUTC(new Date(complaint.receivedDate));
```
andlar: Dates coming from the API are in UTC already, right? I usually do something like "complaint.receievedDateObject = new Date(complaint.receivedDate)" and then use that as the ng-model. If the ng-model-options is set as UTC, then it should all stay that way. On save I usually call "receivedDateObject.getTime()" to send back the long as the "receivedDate".

tmy1313: That is how it used to be, where it worked locally but not on DEV.  Part of the problem that makes this a little different (maybe) is that receivedDate is stored in the DB as a date (not datetime) since there is no need to store the trim.  So, the received date always come back with a time of "0000".  Then when it gets converted to a js Date it is the previous day.  I wonder if we would have this problem in other places in the app if the date was updated between midnight and 4 am?

andlar: Why is receivedDate stored as a Date and not a Timestamp? It'd be reasonable to assume that having the time a complaint was received could be information that might be useful.

And while we shouldn't change the API or DB to suit the UI, JavaScript doesn't know what a Date is without a time, so if we do have some DB Dates to deal with we'll need to handle them consistently across the application, and I don't know that we're doing so now

tmy1313: So, are you suggesting?
* Change the DB type to timestamp
* Change the date picker to a datetime picker
  * do we have an example how we want to enter time (Scheduled Jobs?)


-

PR #555, opened 2019-07-10
fix: remove space between main header and admin header

tmy1313:


kekey1 (approved):
I'm fine with this if it works. Just wondering if it was tested on/how it affects other screen resolutions or on mobile.

tmy1313:
> I'm fine with this if it works. Just wondering if it was tested on/how it affects other screen resolutions or on mobile.

I think it works better than what we have right now when the resolution changes and on mobile.  I want Heather to look at it and see what she thinks.  If she's not happy, it can easily be reverted.

PR #568, opened 2019-07-17
OCD-2754 - Improve Reports responsiveness - Announcements

tmy1313:
feat: update how we are handling checking the Public flag

### DIFF
--- a/src/app/pages/reports/announcements/announcements.component.js
+++ b/src/app/pages/reports/announcements/announcements.component.js
@@ -97,23 +97,40 @@ export const ReportsAnnouncementsComponent = {
 
         getUpdateActivity (detail) {
             let action = '<ul>';
-            action += this.getActionDescriptionIfChanged(detail, 'title', 'Title');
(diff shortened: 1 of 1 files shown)


src/app/pages/reports/announcements/announcements.component.js
```diff
@@ -116,6 +115,24 @@ export const ReportsAnnouncementsComponent = {
...
+            if (detailObject.originalData && detailObject.newData) {
+                if (detailObject.originalData[key] && !detailObject.newData[key]) {
+                    //changed to false
+                    return display + ' changed from true to false';
+                } else if (!detailObject.originalData[key] && detailObject.newData[key]) {
+                    //chnaged to true
+                    return display + ' changed from false to true';
+                }
+            }
+            return '';
```
andlar: I'm not entirely comfortable with returning an empty string. `false` or `undefined` would make me happier, as those are actually indications that nothing changed, while an empty string (though I know it's === false) feels like an actual thing; like, it could be concatenated with stuff reasonably

Not enough to request changes, but leaving this comment before approving them

PR #579, opened 2019-07-19
OCD-2912 - Create a basic complaint

tmy1313:
UI changes requested by Heather

### DIFF
--- a/src/app/components/surveillance/complaint/complaint.component.js
+++ b/src/app/components/surveillance/complaint/complaint.component.js
@@ -56,7 +56,7 @@ export const SurveillanceComplaintComponent = {
             if (changes.certificationBodies) {
                 this.certificationBodies = angular.copy(changes.certificationBodies.currentValue);
                 this.certificationBodies.forEach(acb => {
-                    acb.displayValue = acb.name + (acb.retired ? ' [Retired]' : '');
+                    acb.displayValue = acb.name + (acb.retired ? ' (Retired)' : '');
                 });
             }
             if (changes.errorMessages) {
--- a/src/app/components/surveillance/complaint/complaint.component.js
+++ b/src/app/components/surveillance/complaint/complaint.component.js
@@ -17,8 +17,9 @@ export const SurveillanceComplaintComponent = {
         surveillances: '<',
     },
     controller: class SurveillanceComplaintComponent {
-        constructor ($filter, $log, authService, featureFlags, toaster, utilService) {
+        constructor ($anchorScroll, $filter, $log, authService, featureFlags, toaster, utilService) {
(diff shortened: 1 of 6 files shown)


src/app/components/surveillance/complaint/complaint.component.js
```diff
@@ -55,6 +55,9 @@ export const SurveillanceComplaintComponent = {
             }
             if (changes.certificationBodies) {
                 this.certificationBodies = angular.copy(changes.certificationBodies.currentValue);
+                this.certificationBodies.forEach(acb => {
+                    acb.displayValue = acb.name + (acb.retired ? ' [Retired]' : '');
```
andlar: If this is going to be displayed on the UI, the ACB-management screen displays "Retired" in a span with class "badge". If it's in a dropdown of a select box, the other places where we do retired things use parentheses, not brackets. Looking for consistency

src/app/components/surveillance/complaint/complaint.html
```diff
@@ -131,12 +131,11 @@ <h2 ng-if="$ctrl.currentMode === $ctrl.modes.ADD">Add Complaint</h2>
   <div class="form-group">
     <label class="control-label" for="listings">Selected Certified Products</label>
     <div ng-if="!$ctrl.complaint.listings || $ctrl.complaint.listings.length === 0">
-      No Certified Products Selected
+      <i>No Certified Products Selected</i>
```
andlar: I'm not sure about the `<i>` tag. We don't use it anywhere else in the site, except as a placeholder for the font-awesome icons. The HTML5 spec gives that tag a semantic meaning, but I don't think that meaning is appropriate here. I _think_ this should be emphasized using the `<em>` tag, though I could be persuaded to not change it

src/app/components/surveillance/complaint/complaint.scss
```diff
@@ -25,6 +25,16 @@
         padding-left: 0;
     }
 
+    .breadcrumb a {
+        text-decoration: none;
```
andlar: If it's a link to another page, by 508 requirements it's _required_ to have an underline

src/app/components/surveillance/complaint/complaint.html
tmy1313: I made it italic because that is what Heather specifically requested.

andlar: `<em>` and `<i>` both look the same, but have different semantic meanings

`<em>` is used to indicate that text should be emphasized, which makes sense here

`<i>` is for "... to indicate a technical term, a phrase from another language, a thought, or a ship name, etc."

PR #584, opened 2019-07-24
[#OCD-2749] feat: enable flag for table view of users

andlar:


src/app/components/user/users.html
```diff
@@ -64,4 +64,43 @@ <h2 class="panel-title">Invite a User</h2>
       </div>
     </div>
   </div>
+  <div class="row table-responsive" feature-flag="ocd2749-table" st-table="$ctrl.displayedUsers" st-safe-src="$ctrl.filteredUsers">
+    <div class="col-md-12">
+      <table class="table table-condensed">
+        <thead>
+          <tr>
+            <th scope="col" class="search-header" st-sort="fullName" st-skip-natural="true" st-sort-default="true">Full Name</th>
+            <th scope="col" class="search-header" st-sort="friendlyName" st-skip-natural="true">Friendly<br />name</th>
```
kekey1: Why is "Name" capitalized in "Full Name" but not "Friendly name"?

andlar: Because I changed it in my code but forgot to save before committing. I'll fix it later; I'm sure there will be other things that need to change before this could get pushed anywhere

PR #603, opened 2019-08-01
qa: update Complaints UI based on ONC feedback

tmy1313:
OCD-2912


src/app/components/surveillance/complaint/complaint.html
```diff
@@ -114,22 +114,54 @@ <h2 ng-if="$ctrl.currentMode === $ctrl.modes.ADD">Add Complaint</h2>
...
     </div>
   </div>
+  <div class="form-group">
+    <label class="control-label" for="criteria">Associated Criteria</label><br />
+    <div ng-if="!$ctrl.complaint.criteria || $ctrl.complaint.criteria.length === 0">
+      <em>No Criteria Selected</em>
+    </div>
+    <div class="selected-items-container" ng-if="$ctrl.complaint.criteria && $ctrl.complaint.criteria.length !== 0">
+      <span ng-repeat="criterion in $ctrl.complaint.criteria">
+        <button class="btn btn-primary btn-sm" ng-click="$ctrl.removeCriterion(criterion)">
```
andlar: Can you add an id to these buttons? It'll help with AQA testing if there's an ID, even if it ends up being something like `id="remove-criterion-{{ criterion.number}}"`

PR #614, opened 2019-08-07
fix: Limit access to pages / features with explicit ROLE lists

andlar:


src/app/components/developer/developer.html
```diff
@@ -17,7 +17,7 @@ <h2>Developer</h2>
     <p ng-if="$ctrl.developer.status.status !== 'Active'"><strong>Developer Status:</strong> <i class="fa status-bad" ng-class="$ctrl.developer.status.status === 'Suspended by ONC' ? 'fa-exclamation-circle' : 'fa-ban'"></i> {{ $ctrl.developer.status.status }}</p>
     <span ng-if="$ctrl.showFull">
       <p ng-if="$ctrl.developer.website"><a ai-a href="{{ $ctrl.developer.website }}">{{ $ctrl.developer.website }}</a></p>
-      <table ng-if="$ctrl.developer.statusEvents && $ctrl.developer.statusEvents.length > 0 && $ctrl.hasAnyRole()" class="table table-condensed">
+      <table ng-if="$ctrl.developer.statusEvents && $ctrl.developer.statusEvents.length > 0 && $ctrl.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC', 'ROLE_ACB')" class="table table-condensed">
```
kekey1: Just wondering about the roles specified here. In the old code it said hasAnyRole() and in another file you changed that to include the three roles here plus 'ROLE_ATL', 'ROLE_CMS_STAFF'. Was that a mistake in the old code to allow atl and cms staff access or should they be included here?

src/app/pages/administration/administration.state.js
```diff
@@ -100,7 +100,7 @@ function administrationStateConfig ($stateProvider) {
             resolve: {
                 fuzzyTypes: (authService, networkService) => {
                     'ngInject'
-                    if (authService.hasAnyRole()) {
+                    if (authService.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC'])) {
```
kekey1: Same question as above for this file. It looks like the roles specified were on purpose but I am just checking.

src/app/components/developer/developer.html
andlar: The API returns, for any given Developer, all of the status events. In this case I made the judgement call that ROLE_ATL and ROLE_CMS_STAFF were not likely to be looking for those statuses, and, moving forward, wouldn't expect a ROLE_DEVELOPER to see status histories of other Developers. Given the number of ROLE_ATL and ROLE_CMS_STAFF members I guessed there wouldn't be much outcry, but am up for discussion

src/app/pages/administration/administration.state.js
andlar: Here, I specified the ROLEs that were able to see/manage the fuzzy stuff. Previously, if a non (one of these) users had gone to the page (without it being in their navigation) they would've just seen empty pages, now, they should see the same thing, but we won't even bother with making a network request for stuff they can't get

src/app/components/developer/developer.html
kekey1: Okay, I'm fine with it if that's how it's supposed to work. Was just commenting in case it was an oversight!

PR #629, opened 2019-08-21
feat: Disable next on developer inspect product page without save or valid system data

drbgfc:
* Disable next button on developer inspect product page when there is no valid system data loaded, either via default, or via the save button itself
* Fix attestation not being displayed for the system developer even if saved in the system

[#OCD-3002]


src/app/admin/components/certifiedProduct/inspect/inspect.controller.js
```diff
@@ -213,7 +217,8 @@
         function isDisabled () {
             switch (vm.stage) {
             case 'dev':
-                return (vm.developerChoice === 'choose' && !vm.cp.developer.developerId);
+                return (vm.developerChoice === 'choose' && !vm.cp.developer.developerId)
+                  || (!isSystemDevContactInfoValid() && !vm.isDevUpdated);
```
andlar: I think the vm.isDevUpdated variable doesn't need to exist. If the system Developer information is valid, whether or not the developer was updated, then the "next" button shouldn't be disabled. the `!isSystemDevContactInfoValid()` method checks that

src/app/admin/components/certifiedProduct/inspect/inspect.controller.spec.js
```diff
@@ -140,5 +142,125 @@
...
+                        fullName: 'The Dude',
+                        email: 'abc@abcdefg.com',
+                        phoneNumber: '999-999-9999',
+                    },
+                    transparencyAttestations: [{
+                        acbId: 3,
+                        acbName: 'Drummond Group',
+                        attestation: 'Affirmative',
+                    },
+                    ],
```
andlar: This is odd indentation. I think the ending characters of the transparencyAttestations block don't match the beginning ones

src/app/services/util.service.js
```diff
@@ -31,6 +31,7 @@
             sortRequirements: sortRequirements,
             statusFont: statusFont,
             ternaryFilter: ternaryFilter,
+            isBlank: isBlank,
```
andlar: These "table of contents" sections are sorted alphabetically, by convention

src/app/services/util.service.spec.js
```diff
@@ -494,5 +494,33 @@
                 expect(util.rangeCol(undefined)).toBe('col-sm-12');
             });
         });
+
+        fdescribe('when providing a string', () => {
+            const shouldReturnTrue = 'should return true';
+
+            it('which has no characters ' + shouldReturnTrue, () => {
```
andlar: I'd also test this against `null`, `undefined`, and possibly empty arrays / objects. At least the first two are often in our data; the second two are more of a completionism thing

src/app/admin/components/certifiedProduct/inspect/inspect.controller.spec.js
drbgfc: I agree it is odd and did not want it to be this way. Originally I had it like this (see ahead) but the linter refused to allow the indentation on the object within the array. It required it to be the same indentation as the array (as was committed). Is there a different format you would suggest that separates properties by line and passes the linter? Or maybe we can modify the linter for this scenario?
transparencyAttestations: [{
		acbId: 3,
		acbName: 'Drummond Group',
		attestation: 'Affirmative',
	},
],

src/app/services/util.service.spec.js
drbgfc: Great idea to check for null and undefined. As for empty arrays/objects I could/probably should do that but should also then change the name of the parmeter in the isBlank function from str to, well, something else.

src/app/services/util.service.js
drbgfc: Gotcha, thanks.

src/app/admin/components/certifiedProduct/inspect/inspect.controller.js
drbgfc: Thanks for the feedback, what your saying certainly makes sense in theory. Good catch. Not having isDisabled will break my current next button test since the frontend itself doesn't do the overwrite. Either it will need to be more complex or removed. But, besides that, as you stated, it should not be necessary for the actual fix due to the data update provided by the backend upon pressing save. I will look into this, thanks.

src/app/admin/components/certifiedProduct/inspect/inspect.controller.spec.js
andlar: The way this structure looks in most other places is:
```
transparencyAttestations: [{
		acbId: 3,
		acbName: 'Drummond Group',
		attestation: 'Affirmative',
	}],
```

src/app/services/util.service.spec.js
andlar: Let's just do the null/undefined then. Those other options aren't really needed

src/app/admin/components/certifiedProduct/inspect/inspect.controller.spec.js
drbgfc: Thanks. Just a note, the start of the properties and below will have to be tabbed back one like this to pass the linter, and, as you have already done, remove the trailing object comma.
```
transparencyAttestations: [{
	acbId: 3,
	acbName: 'Drummond Group',
	attestation: 'Affirmative',
}],
```

src/app/services/util.service.spec.js
drbgfc: Already covered them anyway before reading the reply which is fine by me

src/app/admin/components/certifiedProduct/inspect/inspect.controller.spec.js
andlar: I was copy/pasting to get it to that state. I tend to treat code suggestions / samples in these kinds of comments as suggestive, rather then prescriptive :)

src/app/admin/components/certifiedProduct/inspect/inspect.controller.js
drbgfc: This is now implemented along with a fix for it no longer allowing next with this logic due to an existing bug causing attestation to not be populated for the system developer in the UI (as discussed in the meeting today). Thanks.

src/app/admin/components/certifiedProduct/inspect/inspect.controller.js
```diff
@@ -223,6 +225,25 @@
             }
         }
 
+        function isSystemDevContactInfoValid () {
+            return (vm.developer && !vm.isBlank(vm.developer.name) && !vm.isBlank(vm.developer.website))
```
Plow74: Suggestion - these large conditionals take a lot of brain power to decipher. Consider wrapping this in a descriptive method to help the reader understand what you are checking.

Plow74: Talked to Dan, he is already doing what I suggested. Thanks, Dan!

PR #630, opened 2019-08-21
qa: update UI elements for entering surveillance data

andlar:


src/app/components/surveillance/reporting/relevant-listing.component.js
```diff
@@ -50,6 +50,25 @@ export const SurveillanceReportRelevantListingComponent = {
             this.activeSurveillance = undefined;
             this.onCancel();
         }
+
+        calculateCompletion (surveillance) {
+            surveillance.completed = Math.round((
+                (surveillance.surveillanceOutcome ? 1 : 0 ) +
```
kekey1: One of these fields has an "other" choice and if the user picks other I think an additional description is desired... do you want to incorporate that into the calculation?

andlar: Two of them do, but on those you're required to enter the "Description" before saving, which covers the requirement

PR #656, opened 2019-09-25
ui: change wording on user creation workflow

andlar:
[#OCD-2158]

### DIFF
--- a/src/app/pages/registration/create-user.html
+++ b/src/app/pages/registration/create-user.html
@@ -8,7 +8,7 @@
     <div class="col-md-8 col-md-offset-2">
       <div class="row">
         <div class="col-md-12">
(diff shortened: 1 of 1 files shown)


src/app/pages/registration/create-user.html
```diff
@@ -8,7 +8,7 @@ <h1>User Registration</h1>
     <div class="col-md-8 col-md-offset-2">
       <div class="row">
         <div class="col-md-12">
-          <p>Welcome to ONC’s Certified Health IT Product List (CHPL). You have been invited to be an Administrator, which will allow you to manage your organization’s information on the CHPL. Please use the form below to create an account utilizing the ‘create a new account’ link, or sign in to your existing account to add any permissions and/or organizations.</p>
+          <p>Welcome to ONC’s Certified Health IT Product List (CHPL). You have been invited to be an Administrator, which will allow you to manage your organization’s information on the CHPL. Please sign in to your existing account to add any permissions and/or organizations, or create a new account by completing the form and selecting the ‘create a new account’ link below.</p>
```
kekey1: I'm fine with this if it matches what was asked for. However, users cannot add additional permissions (only organizations) so it seems odd to suggest that they may. Also I think the "create a new account" in quotes should match exactly with the link itself, just meaning that I think the "c" should be capitalized.

PR #663, opened 2019-10-10
fix: Add missing Product activity data to downloadable file

Plow74:
OCD-3129


Plow74:
> I don't see anything to comment on with this code. I would like to request that the PR title, when it's merged, follow the format that works with the RELEASE_NOTES, where the title should read (ignoring the "fix:" prefix) as something like: If applied, this Pull Request will
> 
> So, for this one, it might be something like: fix: Add missing Product activity data to downloadable file

heh, I was including the "fix:" prefix as part of the title, as in applying this will 'fix Product activity report missing activity details'.

PR #666, opened 2019-10-17
 fix: Enable next for new developer and add error reporting

drbgfc:
OCD-3141

### DIFF
--- a/src/app/components/listing/inspect/inspect.controller.js
+++ b/src/app/components/listing/inspect/inspect.controller.js
@@ -227,6 +227,9 @@
         }
 
         function isSystemDevContactInfoValid () {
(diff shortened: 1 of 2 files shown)


src/app/components/listing/inspect/inspect.controller.js
```diff
@@ -69,7 +71,7 @@
         }
 
         function loadDev () {
-            if (vm.cp.developer.developerId) {
+            if (vm.cp.developer && vm.cp.developer.developerId) {
```
drbgfc: Note: This line is fixing an exiting, unrelated bug I found in the console where the data was not yet available when the check was made.

PR #667, opened 2019-10-18
feat: Allow jobs to display text, date, and select for additional job data

tmy1313:
OCD-3132

### DIFF
--- a/src/app/pages/administration/jobs/scheduled/trigger.html
+++ b/src/app/pages/administration/jobs/scheduled/trigger.html
@@ -12,18 +12,18 @@
       <label for="{{ param.key }}" >{{ param.label }}</label>
     </div>
     <div class="col-md-6">
-      <input class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required ng-if="!param.type || param.type == 'text'">
-      <textarea class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required ng-if="param.type == 'textarea'"></textarea>
-      <div class="input-group" ng-if="param.type == 'date'">
+      <input class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required ng-if="!param.type || param.type === 'text'">
+      <textarea class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required ng-if="param.type === 'textarea'"></textarea>
+      <div class="input-group" ng-if="param.type === 'date'">
         <input class="input-sm form-control" id="{{ param.key }}" name="{{ param.key }}" type="text" uib-datepicker-popup="MM/dd/yyyy"
-            ng-model="$ctrl.trigger.job.jobDataMap[param.key]" is-open="$ctrl.PickerOpen"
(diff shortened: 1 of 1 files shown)


src/app/pages/administration/jobs/scheduled/trigger.html
```diff
@@ -12,7 +12,20 @@
       <label for="{{ param.key }}" >{{ param.label }}</label>
     </div>
     <div class="col-md-6">
-      <input class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required>
+      <input class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required ng-if="!param.type || param.type == 'text'">
```
andlar:
```suggestion
      <input class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required ng-if="!param.type || param.type === 'text'">
```

I know there's a difference in javascript for `==` vs `===` and `!=` vs `!==`, and think it has something to with types of equality, though I can't remember specifically what it was. That said, the best practices suggest using the three character versions almost exclusively

Similar comments to lines below

src/app/pages/administration/jobs/scheduled/trigger.html
```diff
@@ -12,7 +12,20 @@
...
+      <textarea class="input-sm form-control" type="text" id="{{ param.key }}" name="{{ param.key }}" ng-model="$ctrl.trigger.job.jobDataMap[param.key]" required ng-if="param.type == 'textarea'"></textarea>
+      <div class="input-group" ng-if="param.type == 'date'">
+        <input class="input-sm form-control" id="{{ param.key }}" name="{{ param.key }}" type="text" uib-datepicker-popup="MM/dd/yyyy"
+            ng-model="$ctrl.trigger.job.jobDataMap[param.key]" is-open="$ctrl.PickerOpen"
```
andlar: Should have the is-open object be specific to this parameter; if we get to a point where we have to pick two different dates for job data they'll both open and close simultaneously without doing that. Something like `is-open="$ctrl.PickerOpen[param.key]"` should work. Oh, and prefer the lowercase "P" in "PickerOpen"

src/app/pages/administration/jobs/scheduled/trigger.html
```diff
@@ -12,7 +12,20 @@
...
+            close-text="Close" ng-model-options="{ timezone: 'UTC' }" required>
+        <span class="input-group-btn">
+          <button type="button" class="btn btn-ai btn-sm" ng-click="$ctrl.PickerOpen = true"><i class="fa fa-calendar"></i></button>
+        </span>
+      </div>
+      <select ng-model="$ctrl.trigger.job.jobDataMap[param.key]" id="{{ param.key }}" name="{{ param.key }}" class="form-control"
+        ng-options="item.name for item in param.items" required ng-if="param.type == 'select'">
```
andlar: Where does "param.items" come from? Is that data going to be in the jobs.xml file for this Trigger? If so, I think that's fine, at least for now. Might revisit after letting that sit for a bit

tmy1313: Yes - param.items is in the jobs.xml file.  There may be other solution, but I didn't have time to work on any more elegant solutions.

src/app/pages/administration/jobs/scheduled/trigger.html
tmy1313: I think "===" also checks that the data type is the same
1 == '1'  -> true
1 === '1' -> false

PR #670, opened 2019-10-24
feat: Display the API Key Management Report is in the new style

Plow74:
OCD-2960


src/app/pages/reports/api-keys/api-keys.component.js
```diff
@@ -2,12 +2,14 @@ export const ReportsApiKeysComponent = {
     templateUrl: 'chpl.reports/api-keys/api-keys.html',
     bindings: { },
     controller: class ReportsApiKeyComponent {
-        constructor ($log, ReportService, networkService, utilService) {
-            'ngInject'
+        constructor ($filter, $log, ReportService, networkService, utilService) {
+            'ngInject';
```
andlar: The semi-colon is not standard

src/app/pages/reports/api-keys/api-keys.component.js
```diff
@@ -28,7 +49,10 @@ export const ReportsApiKeysComponent = {
         }
 
         prepare (results) {
-            this.displayed = results.map(item => item);
+            this.displayed = results.forEach(item => {
```
andlar: `forEach` acts on the given array in place and updates it. So if you want to do a change on the existing array you don't need to assign it to anything. `map` operates on the array and returns a new array with the changes you've made. In a `forEach` you don't need to return the item; I think they're actually discarded. `map` requires it

src/app/pages/reports/api-keys/api-keys.html
```diff
@@ -4,7 +4,20 @@
       <h3>Api Key Management</h3>
     </div>
   </div>
-
+  <div class="row">
+    <div class="col-sm-6">
+      <chpl-filter filter-type-name="API Key Management Report"
+                   on-apply-filter="$ctrl.onApplyFilter(filter)" get-filter-data="$ctrl.createFilterDataObject()"
+                   on-clear-filter="$ctrl.onClearFilter()" ></chpl-filter>
```
andlar:
```suggestion
                   on-clear-filter="$ctrl.onClearFilter()"></chpl-filter>
```

Tiny thing

src/app/services/network.service.js
```diff
@@ -217,7 +217,7 @@ export class NetworkService {
     }
 
     getApiUserActivity (activityRange) {
-        var call = '/activity/api_keys';
+        var call = '/activity/metadata/api-keys';
```
andlar: We have 100% test coverage of this file. If you changed this line a test should have failed. If it didn't it's because it's being ignored, which is unfortunate, but either way needs to be addressed. There's probably a test labeled something like `it('should getApiUserActivity`, => {`

src/app/pages/reports/api-keys/api-keys.component.js
```diff
@@ -17,32 +19,63 @@ export const ReportsApiKeysComponent = {
...
-            this.displayed = results.map(item => item);
-        }
-
         search () {
-            let that = this;
             this.networkService.getApiUserActivity(this.dateAdjust(this.activityRange))
                 .then(results => {
-                    that.results = results;
-                    that.prepare(that.results);
+                    this.results = results;
```
andlar:
```suggestion
                    this.results = results.map(item => {
                        item.friendlyCreationDate = this.$filter('date')(item.date, 'MMM d, y H:mm:ss');
                        return item;
                    });
                    // and remove the entire "this.displayed" section
```

I feel like we've talked about this, and I know you were going to change it around. I thought you did, but somehow it doesn't look like what I expected. Sorry, I probably approved it earlier without catching that

Plow74: Hmm, it looks strange in this code snippet. But, I added a this.results variable to accommodate `<div st-table="$ctrl.displayed" st-safe-src="$ctrl.results">`
For some reason, in the above code snippet, it looks like results is being assigned, and then right below being mutated with the results of the map. Anyway, this pattern is what I've seen in other controllers - providing the table with the necessary 'displayed' collection and original dataset.

andlar: I was looking at the full file to make sure I saw it correctly. The problem is that `this.displayed` should _not_ be set in the controller/component. The smart-table directive handles the data flow from `this.results` to the `this.displayed` variable. In some of the other reports it's not done correctly, but it is in the search page. The `this.displayed` variable should not exist in the controller/component at all

src/app/pages/reports/api-keys/api-keys.html
```diff
@@ -46,7 +46,7 @@ <h3>Api Key Management</h3>
           </tr>
           <tr ng-repeat="activity in $ctrl.displayed">
             <td>{{ activity.description }}</td>
-            <td>{{ activity.date | date : 'MMM d, y H:mm:ss Z' : 'UTC' }}</td>
+            <td>{{ activity.date | date : 'MMM d, y H:mm:ss' }}</td>
```
andlar: Why was this removed? At least two of the other reports use
```
          {{ activity.date | date : 'MMM d, y H:mm:ss Z' : 'UTC' }}
``` 
for this column; why is this one different?

Plow74: It was caught as a bug since the screen was showing UTC but the report was showing EST. Looking at the other API report, it does not use UTC. If it's supposed to use UTC then I can put it back - but then is API KEY MANAGEMENT not using UTC a bug? This is from 
https://github.com/chpladmin/chpl-website/blob/staging/src/app/pages/reports/api-key-usage/api-key-usage.html#L90
`<td>{{ activity.creationDate | date : 'MMM d, y H:mm:ss' }}</td>`

andlar: I looked at the Developers and Listings reports, and both of them are using UTC. So is Versions and ONC-ACBs. Feel free to check the others. If API Key management is the outlier then it seems that's the incorrect one. Not sure if it rises to bug level; maybe could just be handled in this PR, but probably worth some sort of conversation

src/app/pages/reports/api-keys/api-keys.html
```diff
@@ -30,6 +30,15 @@ <h3>Api Key Management</h3>
   <div class="row form-group" ng-if="!$ctrl.validDates()">
     <div class="col-sm-12 text-danger">Start date must be before end date, and no more than {{ $ctrl.activityRange.range }} days apart</div>
   </div>
+  <div class="row">
+    <div class="col-sm-5">
+      <div class="form-group">
+        <label for="data-filter" class="control-label">Filter</label>
+        <input chpl-search="filterText" name="dataFilter" id="data-filter" class="form-control" type="search"
+               ng-model="$ctrl.filterText" placeholder="Filter on API Key">
```
andlar: Is the description "just" the API Key? It looked like the filter was filtering on the description, but the placeholder doesn't say that. If someone typed "create" into the filter, would it filter down to just the creation activities? If so, the placeholder doesn't seem right

src/app/pages/reports/api-keys/api-keys.component.js
```diff
@@ -2,13 +2,14 @@ export const ReportsApiKeysComponent = {
     templateUrl: 'chpl.reports/api-keys/api-keys.html',
     bindings: { },
     controller: class ReportsApiKeyComponent {
-        constructor ($filter, $log, ReportService, networkService, utilService) {
+        constructor ($filter, $log, ReportService, networkService, utilService, authService) {
```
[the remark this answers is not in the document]
Plow74: Hmm, I'm not seeing this error in my linter - it's picking up others, but not anything about being alphabetized.

src/app/pages/reports/api-keys/api-keys.html
```diff
@@ -53,3 +53,11 @@ <h3>Api Key Management</h3>
     </div>
   </div>
 </div>
+<div class="main-content row splash-row" id="main-content" tabindex="-1" ng-if="!$ctrl.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC'])">
```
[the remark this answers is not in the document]
Plow74: Yeah, removed the id since it's not being targeted by anything. Also removed that splash-row.

PR #681, opened 2019-11-21
feat!: Support custom parameters for recurring triggers

tmy1313:
OCD-3119


src/app/components/util/string-to-number.js
```diff
@@ -0,0 +1,18 @@
+angular.module('chpl.components').run(function ($rootScope) {
+    $rootScope.typeOf = function (value) {
```
andlar: I don't see this being used anywhere, which is good, because things on `$rootScope` worry me

src/app/components/util/string-to-number.js
```diff
@@ -0,0 +1,18 @@
...
+        return typeof value;
+    };
+})
+.directive('stringToNumber', function () {
+    return {
+        require: 'ngModel',
+        link: function (scope, element, attrs, ngModel) {
```
andlar: What does this directive do? If the form element is of type number, isn't the value going to be a number? Why convert it to a string, or vice versa?

andlar (approved):

kekey1 (approved):

PR #685, opened 2019-12-04
feat-flag!: Restrict ACBs from viewing Transparency Attestation near or in editable scenarios

drbgfc:
[#OCD-3164]


src/app/admin/components/certifiedProduct/developer/edit.controller.js
```diff
@@ -5,22 +5,24 @@
...
 
     /** @ngInject */
-    function EditDeveloperController ($filter, $log, $uibModalInstance, activeAcbs, activeDeveloper, authService, networkService, utilService) {
+    function EditDeveloperController ($filter, $log, $uibModalInstance, activeAcbs, activeDeveloper, authService, featureFlags, networkService, utilService) {
         var vm = this;
 
         vm.addPreviousStatus = addPreviousStatus;
         vm.addressRequired = addressRequired;
+        vm.cancel = cancel;
+        vm.featureFlags = featureFlags;
```
andlar: We don't use this construct anywhere else. Anywhere the "isOn" function is used, this section of the code would just attach the `featureFlags.isOn` function to the `vm`, and then use it later as `vm.isOn('a-flag')`

src/app/admin/components/certifiedProduct/developer/edit.controller.js
```diff
@@ -40,6 +42,7 @@
 
             vm.showFormErrors = false;
             vm.loadedAsInactiveByOnc = (vm.developer.status.status === 'Suspended by ONC' || vm.developer.status.status === 'Under certification ban by ONC');
+            vm.isAcbAdmin = vm.hasAnyRole(['ROLE_ACB']);
```
andlar: It's not a good idea to fix a variable on the `vm` at the time of creation. When someone gets logged out due to inactivity they'll still be "true" on "isAcbAdmin", but wouldn't be if the page was calling the `authService.hasAnyRole` function

Similar things show up if they go to this page unauthenticated, then log in. They've already been set at `vm.isAcbAdmin is false`, and it won't be updated on login

src/app/admin/components/certifiedProduct/management.directive.js
```diff
@@ -31,10 +31,13 @@
         vm.editDeveloper = editDeveloper;
         vm.editProduct = editProduct;
         vm.editVersion = editVersion;
+        vm.featureFlags = featureFlags;
         vm.hasAnyRole = authService.hasAnyRole;
+        vm.isAcbAdmin = vm.hasAnyRole(['ROLE_ACB']);
         vm.isDeveloperEditable = isDeveloperEditable;
         vm.isDeveloperMergeable = isDeveloperMergeable;
         vm.isProductEditable = isProductEditable;
+        vm.isTransparencyAttestationViewable = isTransparencyAttestationViewable;
```
andlar: Why "Viewable"? Is there a requirement about hiding the state of TA? The attestations exist, even if they're not viewable

src/app/components/developer/developer.component.js
```diff
@@ -86,6 +87,26 @@ export const DeveloperComponent = {
...
+            }
+            // Allows any non-ACB to view if flag is off
+            return true;
+        }
+
+        isTransparencyAttestationEditable () {
+            if (this.isEffectiveRuleDatePlusOneWeekOn()) {
+                return !this.isAcbAdmin;
+            }
+            // Enforces existing contract in this case prior to flag requiring ACB to edit, whereas after flag, only non-ACBs can edit
```
andlar: There are more letters in this comment than there are in the function... 😁

src/app/admin/components/certifiedProduct/management.directive.js
drbgfc: In general they all make things not viewable as after discussing with Heather, that is what she decided would be best vs just making disabled. However, sometimes I called it ...viewable() vs ...editable to give it a distinction in that - In some cases, there was an existing contract prior to the flag requiring ACB to edit, whereas after flag, only non-ACBs can edit (the comment you noted). This requires slightly different logic. And, to be consistent with that logic/naming convention from the developer coponent version of these changes, I kept that convention accross the board. 
Here is the logic difference as tests:
(need to grab this from another computer and replace this comment...)

src/app/admin/components/certifiedProduct/developer/edit.controller.js
drbgfc: The existing admin controller uses that construct. I can update that along with mine though to all be consistent.

src/app/admin/components/certifiedProduct/developer/edit.controller.js
drbgfc: Thanks!

src/app/admin/components/certifiedProduct/management.html
```diff
@@ -154,10 +154,12 @@ <h3 ng-if="vm.inspectingCp">System developer information</h3>
                 Zip: {{ vm.activeDeveloper.address.zipcode }}<br />
                 Country: {{ vm.activeDeveloper.address.country }}<br />
                 Website: <span ng-if="vm.activeDeveloper.website"><a ai-a href="{{vm.activeDeveloper.website}}">{{ vm.activeDeveloper.website }}</a></span><span ng-if="!vm.activeDeveloper.website">None</span><br />
-                Transparency Attestation:<br />
-                <ul>
-                  <li ng-repeat="trans in vm.activeDeveloper.transparencyAttestations">{{ trans.acbName }}: {{ trans.attestation ? trans.attestation : 'None' }}</li>
-                </ul>
+                <span ng-if="vm.isTransparencyAttestationViewable()">
```
andlar: I don't see anything in the description, test cases, or comments that suggest the entirety of the transparency attestation fields shouldn't be shown at all

src/app/components/developer/developer.component.js
```diff
@@ -86,6 +86,25 @@ export const DeveloperComponent = {
             }
         }
 
+        isEffectiveRuleDatePlusOneWeekOn () {
+            return this.isOn('effective-rule-date-plus-one-week');
```
andlar: While I appreciate the decomposition to smaller functions, I, personally, don't see anything more readable about this method than just using the "isOn(flag-name)" model. If anything, by having this as a method, when I'm looking at places where the method is called I'll likely wonder if this method does anything other than just check the flag

Using this also makes it a little harder to remove the flag check later, after the flag is on permanently; there will be more places in the code we need to look for it

That said, this is a matter of preference, so I don't think I'm going to ask that it be changed, just that you think about it

src/app/components/listing/inspect/inspect.html
```diff
@@ -63,7 +63,7 @@ <h3>System developer information</h3>
               Zip: {{ vm.developer.address.zipcode }}<br />
               Country: {{ vm.developer.address.country }}<br />
               Website: {{ vm.developer.website }}<br />
-              Transparency Attestation: {{ vm.getAttestationStringForCurrentSystemDeveloper() }}<br />
+              <span ng-if="vm.isTransparencyAttestationViewable()">Transparency Attestation: {{ vm.getAttestationStringForCurrentSystemDeveloper() }}<br /></span>
```
andlar: Not sure this shouldn't be shown either

src/app/admin/components/certifiedProduct/management.html
drbgfc: Was discussed extensively with Heather - that was her final decision - vs disabling - in all cases

src/app/components/listing/inspect/inspect.html
drbgfc: Again, discussed with Heather. In this case though, there was not much of one, as she was 100% sure not to show it. I don't mind brigning it all to her attention and rediscussing though for each case to see if that changes things if you'd like.

src/app/components/developer/developer.component.js
drbgfc: Thanks for the review again. I'll tell you why I did it: It references the same string (the rule name) and so I thought this way it's less prone to mistakes. Since the string being incorrect would not be a syntax error but a bug. A constant could have been used instead, but that is rarely done in JS due to it not really being, so.
As to the second point, which is a good one, I would retort that we recently discussed to code as if it is the final version. Considering that, being concise to what it does now seems the most relevant. Still though, all valid points. As you said, preference. I think in this case I prefer to leave as is though if your ok with it. It's only used in two places, and they very small functions, in a order, in the same file. We know it won't get out of hand/used extensively, as it will be removed anyway at some point.

src/app/admin/components/certifiedProduct/management.html
andlar: Whatever discussions were had, we have to write the code to match what's in the ticket. If you had discussions about this, and they ended up with removing the display of these things, then that's fine, but that needs to be reflected in the description and test cases for the ticket. It also raises more questions for me. Like, if we're hiding this on the screens around where editing is happening, why are we showing it on the screens where the public can see it?

drbgfc: That makes sense, I will update the description. I only have it in comments now, wasn't sure on if I should be the one changing that or if that should be the PO. However, simply disabling editing is still valid, and was done, from the backend perspective. It's the front end that is just removed from view. Even if the items were there, they still wouldn't be editable. This is all addressed in the test cases as they have been updated precisely to match any changes in realtime. The other stuff is all addressed in the comments on the ticket, but certainly needs to be more official (the description). I will take care of it.
"Like, if we're hiding this on the screens around where editing is happening, why are we showing it on the screens where the public can see it?"
Not really sure the answer to that. It's a question for Heather I suppose or something to discuss.

drbgfc: Thanks. I have updated the description to be specific about the requirements and how they differ from the frontend and the backend, on the ticket.

PR #691, opened 2020-01-03
OCD-3124: full listing reports UI

andlar:


src/app/components/smart-table/aiCustomFilter.js
```diff
@@ -4,7 +4,7 @@
         .filter('customFilter', CustomFilter);
 
     /** @ngInject */
-    function CustomFilter ($filter, $log, cfpLoadingBar) {
+    function CustomFilter ($filter/*, $log, cfpLoadingBar*/) {
```
Plow74: Are these commented-out lines supposed to be here?

andlar: I left them there because I wasn't sure if we were going to keep the changes in place, and touch this code so infrequently that I didn't want to lose where it was to actually remove it

PR #693, opened 2020-01-07
OCD-3174: Hide surveillance edit links for 2014 listing + ACB user

kekey1:


src/app/pages/surveillance/surveillance/components/view.component.js
```diff
@@ -5,10 +5,11 @@ export const SurveillanceManagementViewComponent = {
         takeAction: '&',
     },
     controller: class SurveillanceManagementViewComponent {
-        constructor ($log, utilService) {
+        constructor ($log, authService, utilService) {
             'ngInject'
             this.$log = $log;
             this.certificationStatus = utilService.certificationStatus;
+            this.canEdit = authService.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC']);
```
andlar: This sets the boolean value of "canEdit" to a fixed value at the time that the surveillance is loaded. If the user is not logged in at that time, but logs in later, it won't change that value to allow them to edit. Usually, in this case, I'd make this line 
```suggestion
            this.hasAnyRole = authService.hasAnyRole;
```
and add a function below like

```
canEdit () {
return this.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC']) ||
(this.hasAnyRole(['ROLE_ACB']) && this.listing.certificationEdition.name !== '2014');
}
```

Then, in the HTML, the code would be
```
allow-editing="$ctrl.canEdit()"
```

src/app/pages/surveillance/surveillance/components/view.html
```diff
@@ -45,6 +45,6 @@
     </tbody>
   </table>
   <h3>Surveillance</h3>
-  <ai-surveillance allow-editing="true"
+  <ai-surveillance allow-editing="$ctrl.canEdit && $ctrl.listing.certificationEdition.name === '2014'"
```
andlar: Does this allow anyone to edit 2015 listings?

src/app/pages/surveillance/surveillance/components/view.component.js
```diff
@@ -5,10 +5,12 @@ export const SurveillanceManagementViewComponent = {
         takeAction: '&',
     },
     controller: class SurveillanceManagementViewComponent {
-        constructor ($log, utilService) {
+        constructor ($log, authService, featureFlags, utilService) {
             'ngInject'
             this.$log = $log;
             this.certificationStatus = utilService.certificationStatus;
+            this.hasAnyRole = authService.hasAnyRole;
+            this.featureFlags = featureFlags;
```
andlar: Most of the time we just attach the `isOn` function to the component:
```suggestion
            this.isOn= featureFlags.isOn;
```
similarly to the `hasAnyRole` function from the authService. It's not required, but it is convention

PR #698, opened 2020-01-16
[#OCD-3234] Add Self-Developer element to Developer information

andlar:


andlar:
There already was an XSD_MODIFICATION file, but I added to that

kekey1:
> There already was an XSD_MODIFICATION file, but I added to that

Yep. Since we don't know if this feature will go out in the same release that the existing xsd modification is a part of I think it was important to have this specific change stated in the file (and make sure to keep it if the one on STG gets removed after a prod push)

PR #700, opened 2020-01-21
feat!: Add view of all system triggers on Scheduled Jobs page

drbgfc:
[#OCD-3159]


src/app/pages/administration/administration.state.js
```diff
@@ -195,6 +195,13 @@ let states = {
                     }
                     return [];
                 },
+                scheduledSystemJobs: (authService, networkService) => {
+                    'ngInject'
+                    if (authService.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC', 'ROLE_ACB'])) {
```
andlar: ROLE_ACB can't see any of these jobs, so they shouldn't be in this If

src/app/services/network.service.js
```diff
@@ -393,6 +393,10 @@ export class NetworkService {
         return this.apiGET('/schedules/triggers');
     }
 
+    getScheduledSystemJobs () {
```
andlar: Should add a test to the network.service.spec.js file

src/app/services/network.service.js
```diff
@@ -393,6 +393,10 @@ export class NetworkService {
...
+        return this.apiGET('/schedules/triggers/?jobType=system');
```
andlar: I don't think the last `/` should be there. The `/`s imply a sub-level of stuff, while the `?` query is a filter

src/app/pages/administration/administration.state.js
tmy1313: Just want to confirm that an ACB can see this data?

src/app/services/network.service.js
drbgfc: Indeed - fat-fingered but surprisingly it works fine - will remove

src/app/pages/administration/administration.state.js
drbgfc: removed, thanks

drbgfc: No, they can't, thanks. I considered in front-end but copy-pasting caused this gloss-over here

src/app/services/network.service.js
drbgfc: added

src/app/pages/administration/jobs/scheduled/triggers.html
```diff
@@ -24,3 +24,27 @@ <h3>Currently Scheduled Jobs</h3>
...
+        <th scope="col">Description</th>
+        <th scope="col" class="search-header" st-sort="nextRunDate" st-skip-natural="true" st-sort-default="true">Next Run Date</th>
+        <th scope="col" class="search-header" st-sort="triggerScheduleType" st-skip-natural="true">Trigger Schedule Type</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr ng-repeat="scheduledSystemJob in $ctrl.displayedScheduledSystemJobs track by $index">
+        <td>{{ scheduledSystemJob.name }}</td>
+        <td>{{ scheduledSystemJob.description }}</td>
+        <td>{{ scheduledSystemJob.nextRunDate | date:'yyyy-MM-dd HH:mm:ss Z' }}</td>
```
andlar:
```suggestion
        <td>{{ scheduledSystemJob.nextRunDate  ? (scheduledSystemJob.nextRunDate | date:'yyyy-MM-dd HH:mm:ss Z') : 'In Progress' }}</td>
```

PR #701, opened 2020-01-27
OCD-3187 - re-add file that indicates there was an XSD modification

tmy1313:
OCD-3187


andlar (approved):
I've been putting stuff in the file itself to indicate what ticket caused the need for the file. It does make merge conflicts more likely, but also helps with traceability. Not sure if it's necessary though. Approving, but if you want to update feel free

kekey1 (approved):

PR #703, opened 2020-01-31
Ocd 3218 - Update transparency attestations to work in accordance to defined functionality

tmy1313:


src/app/components/developer/developer.component.js
```diff
@@ -91,18 +91,10 @@ export const DeveloperComponent = {
...
-            return isAcbAdmin;
-        }
-
-        isTransparencyAttestationRemoved () {
-            if (this.developer.transparencyAttestations || this.developer.transparencyAttestations.length === 0) {
-                return this.isEffectiveRuleDateOn();
+                return this.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC']);
             } else {
-                return this.developer.transparencyAttestations.reduce((acc, cur) => (cur.attestation ? cur.attestation.removed : false) || acc, false);
+                return this.isEffectiveRuleDateOn();
```
andlar: If both ERD and ERD+1week are off, are TAs editable?

src/app/components/developer/developer.html
```diff
@@ -23,21 +23,11 @@ <h3>Address</h3>
             <chpl-address address="$ctrl.developer.address"></chpl-address>
           </div>
           <div class="col-sm-4">
-            <div class="product-element" id="developer-transparency-attestation" ng-if="!$ctrl.isTransparencyAttestationRemoved()">
+            <div class="product-element" id="developer-transparency-attestation" feature-flag="effective-rule-date" feature-flag-hide>
               <h3>Transparency Attestations</h3>
               <p ng-repeat="att in $ctrl.developer.transparencyAttestations | orderBy : 'acbName'" ng-if="att.attestation">
                 <strong>{{ att.acbName }}:</strong> {{ att.attestation.transparencyAttestation }}</p>
             </div>
-            <div class="product-element removed" id="developer-transparency-attestation" ng-if="$ctrl.isTransparencyAttestationRemoved()">
```
andlar: There were some classes in the `src/app/index.scss` file that were added specifically to handle this "removed" class. I don't know if they're still needed, but if they aren't they should be removed

src/app/index.scss
```diff
@@ -599,10 +599,6 @@ option[disabled] {
     border-radius: 5px;
     background-image: linear-gradient(to bottom, #cccccc 0%, #ddd 51%, #cccccc 100%);
 
-    p {
```
andlar: The `.product-heading` class from a few lines down is also now dead code

src/app/components/developer/developer.component.js
```diff
@@ -91,18 +91,16 @@ export const DeveloperComponent = {
...
-            if (this.developer.transparencyAttestations || this.developer.transparencyAttestations.length === 0) {
-                return this.isEffectiveRuleDateOn();
+            if (this.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC'])) {
+                return true;
             } else {
-                return this.developer.transparencyAttestations.reduce((acc, cur) => (cur.attestation ? cur.attestation.removed : false) || acc, false);
+                if (this.isEffectiveRuleDateOn()) {
+                    return true;
+                } else if (this.isEffectiveRuleDatePlusOneWeekOn()) {
+                    return false;
```
andlar: We'll never get to this case. If the ERD+1Week is on, then the ERD flag would be on too, which would mean that line 98 would return `true` before ever getting to this check

PR #705, opened 2020-02-10
Ocd 3218 - Fix display of TAs on the listing management page

tmy1313:
OCD-3218

### DIFF
--- a/src/app/admin/components/certifiedProduct/management.html
+++ b/src/app/admin/components/certifiedProduct/management.html
@@ -3,8 +3,7 @@
     <div class="row admin-label-spaced lead text-center" ng-if="vm.workType==='merge'">
       <ul class="list-inline">
         <li><strong>What would you like to merge?</strong></li>
-        <li><label><input type="radio" name="merge-type" ng-model="vm.mergeType" value="developer"> Developers</label>
-        </li>
+        <li><label><input type="radio" name="merge-type" ng-model="vm.mergeType" value="developer"> Developers</label></li>
         <li><label><input type="radio" name="merge-type" ng-model="vm.mergeType" value="product"> Products</label></li>
         <li><label><input type="radio" name="merge-type" ng-model="vm.mergeType" value="version"> Versions</label></li>
       </ul>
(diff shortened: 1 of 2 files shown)


src/app/admin/components/certifiedProduct/management.html
```diff
@@ -34,7 +34,7 @@
 
     <div class="row" ng-if="vm.workType === 'manageSurveillance'">
       <div class="col-sm-12">
-        <chpl-surveillance-management allowed-acbs="vm.surveillanceAllowedAcbs" listings="vm.surveillanceListings"></chpl-surveillance-management>
+        <chpl-surveillance-management allowed-acbs="vm.allowedAcbs" listings="vm.surveillanceListings"></chpl-surveillance-management>
```
andlar: I don't think this variable exists in the directive

src/app/admin/components/certifiedProduct/management.html
```diff
@@ -155,28 +155,13 @@ <h3 ng-if="vm.inspectingCp">System developer information</h3>
                 Country: {{ vm.activeDeveloper.address.country }}<br />
                 Website: <span ng-if="vm.activeDeveloper.website"><a ai-a href="{{vm.activeDeveloper.website}}">{{ vm.activeDeveloper.website }}</a></span><span ng-if="!vm.activeDeveloper.website">None</span><br />
                 Self-developer: {{ vm.activeDeveloper.selfDeveloper ? 'Yes' : 'No' }}<br />
-                <span ng-if="vm.isTransparencyAttestationViewable()">
```
andlar: This method has stuff in it about ROLE_ACB and ERD+1week, but that doesn't match what it's being replaced with. I don't know what the right logic _should_ be, but something is off.

Also, the method "isTransparencyAttestationVIewable" isn't used anywhere else in this HTML or in the relevant directive (nor is "isTransparencyAttestationRemoved"), so should be removed if the logic is no longer needed.

tmy1313: The req is: As of ERD do not display TAs for any role.
Updated the code to remove the isTransparencyAttestationViewable() for the directive.

PR #719, opened 2020-03-11
feat: indicate status of system wrt ERD

andlar:
[#cures]

Temporary debugging stuff to help users know the state of the system wrt ERD date. Will be removed before actual ERD

### DIFF
--- a/src/app/navigation/navigation-bottom.html
+++ b/src/app/navigation/navigation-bottom.html
@@ -16,10 +16,10 @@
           {{ announcement.title }}<span ng-if="announcement.text">: {{ announcement.text }}</span>
         </p>
         <p class="announcement nav-text spaced-out badge" ng-if="vm.isOn('effective-rule-date') && !vm.isOn('effective-rule-date-plus-one-week')">
-          Effective Rule Date has passed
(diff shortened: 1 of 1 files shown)


src/app/navigation/navigation-bottom.html
```diff
@@ -15,6 +15,12 @@
         <p class="announcement nav-text spaced-out badge" ng-repeat="announcement in vm.announcements | orderBy:'startDate'">
           {{ announcement.title }}<span ng-if="announcement.text">: {{ announcement.text }}</span>
         </p>
+        <p class="announcement nav-text spaced-out badge" ng-if="vm.isOn('effective-rule-date') && !vm.isOn('effective-rule-date-plus-one-week')">
+          Effective Rule Date has passed
+        </p>
+        <p class="announcement nav-text spaced-out badge" ng-if="vm.isOn('effective-rule-date-plus-one-week')">
+          Effective Rule Date is at least one week ago
```
kekey1:
```suggestion
          Effective Rule Date was at least one week ago
```

andlar: For those coming later, the wording on these notices needs to come from the customer; we're pushing this to DEV to demo it and will update that text when we have final words

PR #723, opened 2020-03-19
[#OCD-3295] feat!: Only allow ROLE_DEVELOPER to withdraw a Change Request

drbgfc:
[#OCD-3295]

### DIFF
--- a/src/app/components/change-request/change-requests.component.js
+++ b/src/app/components/change-request/change-requests.component.js
@@ -83,7 +83,7 @@ export const ChangeRequestsComponent = {
             case 'edit':
                 return this.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC', 'ROLE_ACB', 'ROLE_DEVELOPER']);
             case 'withdraw':
-                return this.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC', 'ROLE_ACB', 'ROLE_DEVELOPER']);
+                return this.hasAnyRole(['ROLE_DEVELOPER']);
                 // no default
             }
         }
--- a/src/app/services/login.service.js
+++ b/src/app/services/login.service.js
@@ -79,7 +79,7 @@
                 let userRole = parseJwt(token).Authority;
                 if (roles) {
                     if (userRole) {
-                        return roles.reduce((ret, role) => ret || userRole === role, false); // true if user has a role in the required list
+                        return roles.reduce((ret, role) => ret || userRole === role, false); // true iff user has a role in the required list
                     }
                     return false; // logged in, role(s) required, user has no role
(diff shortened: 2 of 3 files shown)


src/app/components/change-request/change-requests.component.js
```diff
@@ -83,7 +83,7 @@ export const ChangeRequestsComponent = {
             case 'edit':
                 return this.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC', 'ROLE_ACB', 'ROLE_DEVELOPER']);
             case 'withdraw':
-                return this.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC', 'ROLE_ACB', 'ROLE_DEVELOPER']);
+                return this.hasAnyRole(['ROLE_DEVELOPER']);
```
andlar: I know I wrote tests that covered this line, and I had tried to write them in such a way that different ROLEs would be handled in different ways. Do all the tests pass with this change? Is there still 100% coverage of this file? If so, that's a problem. You don't necessarily need to fix the tests on this PR, but would appreciate taking a look at them

src/app/services/login.service.js
```diff
@@ -79,7 +79,7 @@
                 let userRole = parseJwt(token).Authority;
                 if (roles) {
                     if (userRole) {
-                        return roles.reduce((ret, role) => ret || userRole === role, false); // true iff user has a role in the required list
+                        return roles.reduce((ret, role) => ret || userRole === role, false); // true if user has a role in the required list
```
andlar: I think this is wrong. `iff` = `if and only if`, not just `if`. If this is a change because you think it's a typo, please consider it with the understanding that `iff` has a defined meaning

src/app/components/change-request/change-requests.component.js
drbgfc: They did all pass. But, I will add new tests to reflect the changes. Thanks.

src/app/services/login.service.js
drbgfc: I will revert, thanks.

src/app/components/change-request/change-requests.component.js
andlar: If they all passed then the tests are wrong. If we have 100% coverage of a file, and change something, but not tests fail, then our tests aren't doing what they should be

drbgfc: IMO they didn't fail because they didn't cover the previous implementation as far as roles in all cases of withdrawls and the specific case I needed to modify in the code-base, not because the tests were wrong. I saw to it to confirm this before adding new ones. That's why I added new ones. Let me know what you think.

PR #726, opened 2020-04-01
OCD-3290 - Change the order of display of revised criteria so "Cures Update" is first

tmy1313:
OCD-3290


src/app/services/util.service.js
```diff
@@ -412,5 +394,234 @@
...
+            //If we don't find the item in the referece array, put it at the end
+            let index = certificationResultSortOrder().findIndex(item => item === valueToFind);
+            if (index === -1) {
+                index = 10000;
+            }
+            return index;
+        }
+
+        function certificationResultSortOrder () {
+            return ['170.302 (a)',
```
kekey1, line 433: Same concern here with the list of all the criteria... in this case we have services that return them from the API. Maybe you could adjust one of those services to use the sorting code you wrote in the API? Is a list guaranteed to remain sorted once it goes from api json into browser javascript? I just wonder if there is any other way to accomplish it. If not, I will get over it. :)

src/app/services/util.service.js
```diff
@@ -412,5 +394,234 @@
...
+                    valueToFind += '(Cures Update)';
+                }
+            } else {
+                valueToFind = certResult;
+            }
+
+            //If we don't find the item in the referece array, put it at the end
+            let index = certificationResultSortOrder().findIndex(item => item === valueToFind);
+            if (index === -1) {
+                index = 10000;
```
kekey1, line 427: is there a MAX_VALUE type of constant in javascript?

src/app/services/util.service.js
```diff
@@ -412,5 +394,234 @@
...
+            return certificationResultSortIndex(valueToFindA) - certificationResultSortIndex(valueToFindB);
+        }
+
+        function certificationResultSortIndex (certResult) {
+            //Handle both criteria numbers and certificationResult objects
+            var valueToFind;
+            if (certResult.number) {
+                valueToFind = certResult.number;
+                if (isCertResultForCuresUpdateCriterion(certResult)) {
+                    valueToFind += '(Cures Update)';
```
kekey1, line 418: Do we have a list of constants anywhere that this "(Cures Update)" string may be able to be added to? I feel like it's probably in a lot of places.

kekey1 (changes requested):

src/app/services/util.service.spec.js
```diff
@@ -118,31 +118,9 @@
                     expect(util.sortCert('170.302 (a)')).toBeLessThan(util.sortCert('170.314 (a)(10)'));
                 });
 
-                it('should be able to sort objects by name', () => {
-                    expect(util.sortCert({name: '170.314 (a)(2)'})).toBeLessThan(util.sortCert({name: '170.314 (a)(10)'}));
```
andlar, line 122: This test was added because there were some lists of criteria that didn't have "number" as fields, but instead had "name". Unless you can confirm that there are no criteria that come back from the API like that any more then I think the test needs to be reinstated

src/app/services/util.service.spec.js
```diff
@@ -118,31 +118,9 @@
...
                     expect(util.sortCert({number: '170.314 (a)(2)'})).toBeLessThan(util.sortCert({number: '170.314 (a)(10)'}));
                 });
-
-                it('should sort strings that have identical numbers by title', () => {
-                    let a = '170.315 (b)(2): A title';
-                    let b = '170.315 (b)(2): A title (with stuff)';
-                    expect(util.sortCert(a)).toBeLessThan(util.sortCert(b));
-                });
-
-                it('should sort objects that have identical numbers by title', () => {
```
andlar, line 135: While the description of the test isn't great, the way to make it better would be to update the "title" values below to have "(Cures Update)" in some of them, and update the test to expect the Cures Update ones to come first

src/app/services/util.service.spec.js
```diff
@@ -196,35 +174,35 @@
 
             it('should be able to sort nonconformity types', () => {
                 var criteria2014_g_4 = { number: '170.314 (g)(4)' };
-                var criteria2014_g_10 = { number: '170.314 (g)(10)' };
+                var criteria2014_h_1 = { number: '170.314 (h)(1)' };
```
andlar, line 177: Why did this number change? I don't see why the functionality in this ticket would necessitate changing this particular test

src/app/services/util.service.spec.js
```diff
@@ -196,35 +174,35 @@
...
                 var criteria2015_d_1 = { number: '170.315 (d)(1)' };
                 var criteria2015_e_1 = { number: '170.315 (e)(1)' };
-                var criteria2015_e_1_removed = { number: '170.315 (e)(1) (Removed)' };
+                var criteria2015_e_1_cures = { number: '170.315 (e)(1)(Cures Update)' };
```
andlar, line 180: Non-conformity types are a bunch of strings, and some of them have the string "Removed" in them. This variable / test below were set up to ensure that the "removed" version of a criteria stayed in the same location as it had before

src/app/services/util.service.spec.js
```diff
@@ -196,35 +174,35 @@
...
                 var criteria2015_g_10 = { number: '170.315 (g)(10)' };
                 var transparency_k_2 = { number: '170.523 (k)(2)' };
                 var other = { number: 'Other Non-Conformity' };
-                expect(util.sortNonconformityTypes(criteria2014_g_4)).toBeLessThan(util.sortNonconformityTypes(criteria2014_g_10));
-                expect(util.sortNonconformityTypes(criteria2014_g_10)).toBeLessThan(util.sortNonconformityTypes(criteria2015_d_1));
+                expect(util.sortNonconformityTypes(criteria2014_g_4)).toBeLessThan(util.sortNonconformityTypes(criteria2014_h_1));
+                expect(util.sortNonconformityTypes(criteria2014_h_1)).toBeLessThan(util.sortNonconformityTypes(criteria2015_d_1));
                 expect(util.sortNonconformityTypes(criteria2015_d_1)).toBeLessThan(util.sortNonconformityTypes(criteria2015_e_1));
                 expect(util.sortNonconformityTypes(criteria2015_e_1)).toBeLessThan(util.sortNonconformityTypes(criteria2015_g_4));
-                expect(util.sortNonconformityTypes(criteria2015_e_1)).toBe(util.sortNonconformityTypes(criteria2015_e_1_removed));
```
andlar, line 211: This is sorting non-conformity types. I tried to get to close to 100% coverage of the utility class; if functionality changed to sort criteria then there should be some new tests added to handle the new functionality, and maybe modifications made to this test, but now it looks like the "removed" criteria sorts aren't being tested

andlar (changes requested):

src/app/services/util.service.spec.js
tmy1313: 170.314 (g)(10) is not a valid criteria, therefore it is not in the list to be sorted.

src/app/services/util.service.spec.js
```diff
@@ -126,22 +126,22 @@
                     expect(util.sortCert({number: '170.314 (a)(2)'})).toBeLessThan(util.sortCert({number: '170.314 (a)(10)'}));
                 });
 
-                it('should sort strings that have identical numbers by title', () => {
+                it('should sort objects that are Cures Update before the original criteria', () => {
```
andlar, line 129:
```suggestion
                it('should sort strings that are Cures Update before the original criteria', () => {
```

andlar (dismissed):

kekey1 (dismissed):

src/app/pages/charts/surveillance/surveillance.component.js
```diff
@@ -168,7 +171,7 @@ export const ChartsSurveillanceComponent = {
                     default: false;
                     }
                 })
-                .sort((a, b) => that.utilService.sortOtherNonconformityTypes(a.nonconformityType) - that.utilService.sortOtherNonconformityTypes(b.nonconformityType))
+                .sort((a, b) => that.utilService.sortCertActual(a, b))
```
andlar, line 174: If you're not calling the old sort function anymore, can you check to see if it's called anywhere at all? if it's not used any more I'd rather just remove it entirely

src/app/pages/charts/charts.component.js
```diff
@@ -3,10 +3,11 @@ export const ChartsComponent = {
     bindings: {
     },
     controller: class ChartsComponent {
-        constructor ($log, networkService) {
+        constructor ($log, networkService, utilService) {
             'ngInject'
             this.$log = $log;
             this.networkService = networkService;
+            this.utilService = utilService;
```
andlar, line 10: Is this being used anywhere?

kekey1 (approved):

andlar (approved):

PR #731, opened 2020-04-07
OCD-2162: Support Developer Change Request for Developer Details

andlar:


src/app/components/change-request/change-requests.html
```diff
@@ -13,6 +13,18 @@
...
+        <div uib-dropdown auto-close="outsideClick" class="btn-group">
+          <button class="btn btn-default dropdown-toggle" type="button" id="filter-type-button" uib-dropdown-toggle aria-haspopup="true" aria-expanded="true">
+            <i class="fa fa-filter"></i> Request Type
+          </button>
+          <ul class="dropdown-menu" aria-labeledby="filter-type-button" uib-dropdown-menu>
+            <st-list-multiple
+              collection="$ctrl.changeRequests"
+              predicate="requestType"
+              fixed-items="$ctrl.filterItems.types"
+              ></st-list-multiple>
```
tmy1313, line 25: Is this the formatting you intended?

andlar: Yep. It's similar to the trailing comma thing. If the "close" line was on the same line as the fixed-items line, then the next time I need to add a new attribute it'd require the fixed-items line to be changed, according to git, when it isn't functionally changed

PR #744, opened 2020-05-08
OCD-3117:  feat!: Add filters to Complaints pages 

andlar:


src/app/components/filter/filter.component.js
```diff
@@ -0,0 +1,82 @@
...
+          let filterData = {};
+          filterData.dataFilter = this.filterText;
+          filterData.tableState = this.tableController.tableState();
+          return filterData;
+          }
+
+          tableStateListener (tableController) {
+          this.tableController = tableController;
+          }
+        */
```
tmy1313, line 77: Is there a purpose for leaving in the commented code?

src/app/index.scss
```diff
@@ -813,12 +815,17 @@ option[label^="Removed"] {
...
     }
+
+    &.rotate {
+        flex-direction: column;
+    }
 }
 
 .flex-item {
-    min-width: 150px;
+    //min-width: 150px;
```
tmy1313, line 825: Commented out styles.  Is this intentional?

src/app/components/filter/filter.component.js
andlar: Forgot to remove that one. Had hoped to make the filter component handle all of this stuff, but couldn't figure that out

src/app/index.scss
andlar: Everything in the styles is pending review / update from UX team. I want to keep some of the style stuff around because it was added for a reason, but I'm not sure if the reason still exists. Before it gets really deployed / UX finalized I'd expect the commented out stuff to be removed / cleaned up.

PR #763, opened 2020-06-18
OCD-2944 - Update ACB specific jobs to use ACB ids rather than names

tmy1313:


andlar (dismissed):

kekey1 (dismissed):

kekey1 (dismissed):

src/app/index.constants.js
```diff
@@ -7,5 +7,6 @@
         .constant('MINUTES_BETWEEN_KEEPALIVE', 5)
         .constant('RELOAD_TIMEOUT', 3000)
         .constant('SPLIT_PRIMARY', '☺')
-        .constant('SPLIT_SECONDARY', '☹');
+        .constant('SPLIT_SECONDARY', '☹')
+        .constant('SPLIT_COMMA', ',');
```
andlar: I don't think this is necessary. In all of our other splits & merges in the UI project we just use the actual character that is being split or merged on. The SPLIT_PRIMARY and SPLIT_SECONDARY constants are there for two reasons. The first is that when we use those we're splitting/merging on fields that have user entered data, so we needed to find characters that were unlikely to be in the user generated information, and if/when that ever broke we could update them to a "more obscure" character as needed. The second reason is that the two characters we use in those SPLIT_* constants can't actually be typed on a keyboard, so by making them constants we're able to avoid having to do a copy/paste when needing to use them

I can see two paths forward:
1) Remove the constant entirely. This has the advantage of removing all of the places where the constant needs to be injected, shrinks the lines that use it from needing 15-ish characters to three, and makes it explicit what characters we're using. This is also consistent with the other few dozen places in the UI where we do splits/joins on data that we know has a structured format
2) Keep the constant. If we do this, for consistency's sake I think we'd need to go through all of the other split/joins and replace them with new constants. In a quick search of our code I think we use all of the characters in: `. ; - ( ) : # $` as well as the regex `/ /`. Also, if we do keep the constant, I think the name needs to change. If, for some reason, we need to change the split character to a semi-colon, or slash, `SPLIT_COMMA` would then be a name that didn't match what the character was

src/app/pages/administration/jobs/scheduled/triggers.component.js
```diff
@@ -38,6 +42,14 @@ export const JobsScheduledTriggersComponent = {
                 data: trigger,
             });
         }
+
+        _getAcb (acbId) {
+            return this.acbs.find(acb => acb.id === acbId)
+        }
+
+        _getAcbNamesCommaDelimited (acbs) {
```
andlar: Could this be something like `getFormattedAcbs`? If we needed to change this to have `/` delimited, or formatted as list items, or maybe with both the ACB name and ACB code, or something, the name wouldn't make sense. The method is there to get a list of the ACBs, formatted in a way that meets requirements; the fact that it's names separated by commas is an implementation detail

andlar (changes requested):

kekey1 (approved):

andlar (approved):

PR #767, opened 2020-06-24
UI automation test

palakpatel311:


Plow74 (dismissed):

README.md
```diff
@@ -44,6 +44,7 @@ yarn install
 * `yarn test:auto`: Run the unit tests continuously, re-running the tests on any file change
 * `yarn test:ahrq`: Run tests once in a fashion suitable for the deployment environment
 * `yarn test:ci`: Run the tests once in a fashion suitable for a Continuous Integration environment
+* `yarn e2e`: Run the end to end integration tests
```
andlar, line 47: I think you added the `yarn e2e:lint` command too, right? If so, it should be in the README as well

e2e/components/login/Login.po.js
```diff
@@ -0,0 +1,28 @@
+import BasePage from '../../utilities/BasePage.js';
+const config = require('../../config/mainConfig.js');
+
+const loginElements= {
+    LoginButton: '#login-toggle',
+    Username: '[name="username"]',
+}
+
+class LoginPage extends BasePage {
```
andlar: Would it make more sense for this to be a `LoginComponent` instead of a Page?

e2e/components/login/Login.po.js
```diff
@@ -0,0 +1,28 @@
+import BasePage from '../../utilities/BasePage.js';
+const config = require('../../config/mainConfig.js');
+
+const loginElements= {
+    LoginButton: '#login-toggle',
```
andlar: We use camelCase for almost all of the variables and member elements in javascript code. The only exception is in an actual Class name. This ought to be `loginButton`

e2e/components/login/Login.po.js
```diff
@@ -0,0 +1,28 @@
+import BasePage from '../../utilities/BasePage.js';
+const config = require('../../config/mainConfig.js');
+
+const loginElements= {
```
andlar: I just added a rule that requires `=`s to have spaces around them. All of the rest of the codebase has that state, except for about 10 issues that I fixed. It's in my OCD-3408 branch if you'd like to merge that branch into yours. That will show a few other style issues in other parts of this PR

e2e/components/login/Login.po.js
```diff
@@ -0,0 +1,28 @@
...
+    LoginButton: '#login-toggle',
+    Username: '[name="username"]',
+}
+
+class LoginPage extends BasePage {
+    open () {
+        super.open(config.baseUrl);
```
andlar: I'm not sure if component POs should have anything to do with the URL. I would think a test of these should just load the BasePage URL, or any other page that has the component that needs to be tested, and then do the asserts of the component pieces

e2e/components/login/Login.po.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    get login () {
```
andlar: This is a button, right? Would it make sense to name it `loginButton`? Or is it really the `toggleLoginComponent` button?

e2e/components/login/Login.po.js
```diff
@@ -0,0 +1,28 @@
...
+        return $(loginElements.LoginButton);
+    }
+
+    get username () {
```
andlar: This is an input field, right? To me, "get username" would be the text display of a user's username. Maybe `usernameInput`?

e2e/components/login/Login.po.js
```diff
@@ -0,0 +1,28 @@
...
+        return $(loginElements.Username);
+    }
+
+    gotoLoginPage () {
```
andlar: Components don't have pages, so what about `openLoginComponent () {`?

e2e/components/login/Login.spec.js
```diff
@@ -0,0 +1,15 @@
...
+
+beforeEach(async () => {
+    await SearchPage.open();
+});
+
+describe('Login page', () => {
+
+    it('should have username field', () => {
+        LoginPage.gotoLoginPage();
+        assert.equal(LoginPage.username.isDisplayed(),true);
```
andlar: Is there a reason to use assert? WebdriverIO comes with "expect", and this line could be done with `expect(LoginPage.username).toBeDisplayed();`

e2e/pages/resources/overview/OverviewPage.po.js
```diff
@@ -0,0 +1,47 @@
...
+        return $(overviewElements.ONCACBATLtable);
+    }
+    get acbatlTableRow (){
+        return $$(overviewElements.RowValue);
+    }
+    get acbatlTableCol (){
+        return $$(overviewElements.ColValue);
+    }
+
+    compareTableData (){
```
andlar: This function should be in the spec file. While it's specific to the Page, it's only of use for testing, and only for the single test that it's used for

e2e/pages/resources/overview/overviewPage.spec.js
```diff
@@ -0,0 +1,22 @@
+import OverviewPage from './OverviewPage.po.js';
+import SearchPage from '../../../pages/search/SearchPage.po.js';
```
andlar: Why does the OverviewPage need the Search page? Why can't this test just open the Overview page directly?

e2e/components/login/Login.spec.js
```diff
@@ -0,0 +1,15 @@
+import LoginPage from './Login.po.js'
```
andlar: The Overview spec file doesn't have a capitalized filename (overviewPage.spec.js). I don't think this one should either. We have to (should?) capitalize the files that export classes so imports look right, though now that I type that I'm not sure that's true either...

e2e/pages/search/SearchPage.po.js
```diff
@@ -0,0 +1,29 @@
...
+    chplResource: '#resource-toggle',
+    overview: '=Overview',
+}
+
+class SearchPage extends BasePage {
+    open () {
+        super.open(config.baseUrl);
+    }
+
+    get chplResource (){
```
andlar: These "get"s should probably both be something like `get chplResourceButton () {`

andlar (changes requested):

e2e/components/login/Login.spec.js
palakpatel311: Assert is easy to use and also could pass in with optional message at the end so helps in debugging.

e2e/pages/resources/overview/overviewPage.spec.js
palakpatel311: Search page is actually home page. so navigation to overview page should be needed from homepage. It is not good practice to just go directly overview page for testing.

e2e/components/login/Login.spec.js
palakpatel311: So by looking at sample projects on wdio site, we can follow having import with first letter capitalized in the test where we are using page but page name can still follow camelCase.

I am open to change either way

e2e/components/login/Login.spec.js
```diff
@@ -0,0 +1,15 @@
+import LoginComponent from './loginComponent.js'
```
andlar: This file should also be lowercase named

e2e/components/login/loginComponent.js
```diff
@@ -0,0 +1,24 @@
...
+
+    get toggleLoginComponent () {
+        return $(loginElements.loginButton);
+    }
+
+    get usernameInput () {
+        return $(loginElements.userName);
+    }
+
+    gotoLoginComponent () {
```
andlar: This doesn't "go" anywhere, right? I just opens the component, so it shouldn't be named "goto"

e2e/pages/resources/overview/OverviewPage.po.js
```diff
@@ -0,0 +1,47 @@
+import BasePage from '../../../utilities/basePage';
```
andlar: This file is named "OverviewPage.po.js", but is imported as `import OverviewPage from './overviewPage.po.js';`. That might work on a Windows machine, but Linux won't be able to import it, as Linux cares about file name cases. This one should be renamed as "overviewPage.po.js", or possibly just "overview.po.js", as the "po" part indicates that it's a page

e2e/pages/search/SearchPage.po.js
```diff
@@ -0,0 +1,29 @@
+import BasePage from '../../utilities/basePage';
```
andlar: Similar concerns about filename case as in the Overview page

e2e/utilities/BasePage.js
```diff
@@ -0,0 +1,6 @@
+export default class BasePage {
```
andlar: Is this a page object? Should be in a file named `base.po.js`, right?

andlar (changes requested):

e2e/components/login/Login.spec.js
andlar: If this has benefits great, but I don't see an optional message. I also don't know where any documentation is for the `assert` object. I was able to find webdriverio `expect` stuff by googling for it, but don't know what to look for to find the `assert` docs

palakpatel311: https://www.chaijs.com/api/assert/

e2e/pages/resources/overview/overviewPage.spec.js
andlar: Testing navigation ought to be done in a test under the components/navigation files. Tests of a given page shouldn't depend on any other page.

e2e/components/login/Login.spec.js
```diff
@@ -0,0 +1,15 @@
+import LoginComponent from './login.po'
```
andlar: This file needs a different filename: `login.spec.js` with the lowercase `L`;

e2e/components/login/login.po.js
```diff
@@ -0,0 +1,22 @@
...
+        return $(loginElements.userName);
+    }
+
+    openLoginComponent () {
+        this.toggleLoginComponent.click();
+        return this;
+    }
+}
+
+module.exports = new LoginComponent();
```
andlar: Have done some digging. This construction is used in commonJS and mostly for nodeJS and backend server JS development. The `export`/`import` construction is the standard for ES6, and is considered to be best practice going forward. I'd like to switch all of these over to that mode

In this case, I think the right statement should be `export default class LoginComponent;`, and where it's being imported would be `import LoginComponent from 'login.po';` which is already in place in the spec file

e2e/config/mainConfig.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
andlar, line 1: Here we'd need something like 

```
const config = {
baseUrl: 'https://etc',
etc.
}

export default config;
```

And in the Hooks file, it would be `import config from '../config/mainConfig;' and then it could be used the same way

Oh, and typing that last line made me recognize that `mainConfig` isn't in the right format. Maybe just `main`? As it's already in the config directory?

andlar (changes requested):

e2e/components/login/Login.spec.js
andlar: Have looked into this more. The `expect` matchers that come with wdio give us a lot of things we can check in that are specific to web pages, like styles, elements being enabled/disabled, things being visible, etc. It also extends jest matchers, which lets us use a whole bunch of other basic matchers, and means we don't need a separate package. As well, with the wdio expect we _can_ have optional display messages like:

```
expect(LoginComponent.usernameInput).toBeDisplayed({message: 'usernameInput is not displayed, but it should be'});
```

The other thing that concerns me is that I can't see how we're even getting `assert` into the project. I know it's in the `require('assert')` in the wdio.conf file, which looks through node_modules for the required thing, but there's nothing in package.json about chaijs, so I don't know how it's finding it

palakpatel311: This is where we are getting assert. 
before: function (capabilities, specs) {
        assert = require('assert')
    },

Can we use both assert and expect?
I am not a fan of chain commands like what expect uses.

andlar: I see that line as the place where our code requires the `assert` package, but I don't see anything that tells me where the package comes from. I'm not completely against keeping both assert and expect as options, but would be interested in seeing if the team can settle on one of them at some point, as having two different packages to use for the same purpose doesn't seem like a great idea

andlar (approved):

ashwinimore (approved):

PR #774, opened 2020-07-20
[#OCD-3426] fix!: Display modifications to new fields in Listing Reports

andlar:


kekey1 (approved):
Worth adding a unit test for this?

andlar:
Thought about unit tests, but that reports code is such a mess that I didn't think it was worth it. To even start with a test I think we'd have to build a fully valid "before" and "after" listing that only differed in these values, and while I've tried in the past to do that, every time we change the listing JSON structure it breaks all of those tests, or at least often does

tmy1313 (approved):

PR #777, opened 2020-07-24
[#OCD-3420] fix!: Only remove one of duplicate values in Criteria edit

andlar:


kekey1 (approved):
Approved but just curious of Test Standards had the same issue as the other fields?

andlar:
> Approved but just curious of Test Standards had the same issue as the other fields?

No, the problem with the other four is that they have additional data as well as the actual "item". We can't _edit_ a test standard the way we can the other four. I know the ticket is about removing them, but the way duplicate test standards are treated there's no way to even get them into the system. Uploading a listing with duplicates of the others doesn't remove them, because we don't know which of the duplicates we should keep, because they have extra info. With standards there's no extra info, so we just remove one (or more) as necessary

tmy1313 (approved):

PR #780, opened 2020-07-28
[#OCD-3364] feat-flag!: Display Direct Review data on Developers pages

andlar:


e2e/components/direct-reviews/direct-reviews.po.js
```diff
@@ -0,0 +1,115 @@
+const elements = {
+    root: 'div.direct-reviews',
+    directReviews: '.direct-review',
```
palakpatel311, line 3: Identifying HTML elements by class names are discouraged. Preferred ways are id, name, xpath.

e2e/components/direct-reviews/direct-reviews.spec.js
```diff
@@ -0,0 +1,108 @@
...
+});
+
+describe('the Direct Reviews component', () => {
+    describe('for Greenway Health', () => {
+        beforeEach(() => {
+            page.selectDeveloper('Greenway Health, LLC');
+        });
+
+        it('should have four direct reviews', () => {
+            expect(component.getDirectReviews().length).toBe(4);
```
palakpatel311: Will this developer have 4 DRs always?
If it can change in future, wondering if we should have this specific assertions

ashwinimore (dismissed):

e2e/components/direct-reviews/direct-reviews.spec.js
andlar: It will change in the future, and likely will before the month is out. That said, the current feature will only have these four, and will not have any different number, given the mock data we have. Unfortunately, until we connect with the real system we'll not be able to set it as any specific value, and even when we do we can't really even say that there will ever be a Developer that we can use for testing. This is one of those tests that will need to be updated over time

e2e/components/direct-reviews/direct-reviews.po.js
andlar: Why is it discouraged? I went down a rabbit hole of trying to figure out what the best way of identifying this was, and this article has something that makes sense to me: https://automationpanda.com/2019/01/15/web-element-locators-for-test-automation/ specifically in the "How do I write good locators?" section

palakpatel311: Typically, class names could be found multiple times in a page. When this happens, selenium will pick first element of all found ones or fail the TC. 
In this case, can we consider class name as unique even in future?

e2e/components/direct-reviews/direct-reviews.po.js
```diff
@@ -0,0 +1,115 @@
...
+    getDirectReviewTitle (directReview) {
+        return directReview.$(elements.title);
+    }
+
+    getDirectReviewSubtitle (directReview) {
+        return directReview.$(elements.subtitle);
+    }
+
+    getDirectReviewBegan (directReview) {
+        return directReview.$(elements.start).$(elements.drDataElement);
```
palakpatel311, line 51: I see chaining of locators are being used in get methods. Fox example, if we were to add a subheader or any new element, will it affect these?

kekey1 (dismissed):

palakpatel311 (dismissed):
We will see how to deal with locators in future.

Plow74 (dismissed):

e2e/components/direct-reviews/direct-reviews.po.js
andlar: The way these work the "chained" locator will look anywhere in the DOM below the "parent". It's not specific to direct parent/child relationships, but more "ancestor/descendant"

kekey1 (approved):

tmy1313 (approved):

PR #781, opened 2020-07-29
OCD-3441 - add XSD-MODIFCATION file to indicate that there is a change to the xsd

tmy1313:
OCD-3441


andlar (approved):
I think this will work. I usually put the ticket number in the file so we can follow why the file exists, but I don't think that's required

kekey1 (approved):

PR #783, opened 2020-07-30
OCD 3420: Automation test for uploading listing

palakpatel311:
image
image


README.md
```diff
@@ -44,7 +44,7 @@ yarn install
 * `yarn test:auto`: Run the unit tests continuously, re-running the tests on any file change
 * `yarn test:ahrq`: Run tests once in a fashion suitable for the deployment environment
 * `yarn test:ci`: Run the tests once in a fashion suitable for a Continuous Integration environment
-* `yarn e2e`: Run the end to end integration tests
+* `yarn e2e`: Run the end to end integration tests. This script needs to be run as "USERNAMEACB=UserName_for_ACB PASSWORDACB=PassWord_for_ACB USERNAMEADMIN=UserName_for_ADMIN PASSWORDADMIN=PassWord_for_ADMIN yarn e2e"
```
andlar: Requiring these to be entered from the command line is not how we usually handle credentials. Putting them in some sort of gitignored file and inserting it during a Bamboo build is the preferred way of doing things, especially as that allows different users to use different credentials and save them in their own local file systems

e2e/components/login/login.po.js
```diff
@@ -14,10 +16,34 @@ class LoginComponent {
         return $(loginElements.userName);
     }
 
+    get passwordInput () {
+        return $(loginElements.password);
+    }
+
+    get loginBtn () {
```
andlar: Prefer spelling out simple words, as opposed to abbreviating them

e2e/components/login/login.po.js
```diff
@@ -14,10 +16,34 @@ class LoginComponent {
...
+        return $(loginElements.login);
+    }
+
     openLoginComponent () {
         this.toggleLoginComponent.click();
         return this;
     }
+
+    loginAsACB () {
```
andlar, line 39: If you're going to add functionality to a page-object, it would be useful to add tests that cover that functionality

e2e/pages/administration/upload/upload.po.js
```diff
@@ -0,0 +1,39 @@
...
+    get uploadBtn () {
+        return $(uploadElements.uploadBtn);
+    }
+
+    get uploadSuccessfulText () {
+        return $(uploadElements.uploadSuccessfulText);
+    }
+
+    get expectedUploadSuccessfulText () {
+        return uploadAsserts.expectedUploadText;
```
andlar: I don't think expected values should be in the page-objects. The page-objects should be there to return values / elements / etc., but it's the spec files that should have the expected values

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,81 @@
+class ConfirmPage {
```
andlar, line 8: There are a lot of page elements here but no spec file that asserts anything about them. Any functionality in a page-object ought to be tested in a related spec file

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,81 @@
...
+
+    gotoConfirmListingPg (inspectListingId ) {
+        $('//button[@id="pending-listing-inspect-' + inspectListingId + '"]').waitAndClick();
+        this.inspectNextBtn.waitAndClick();
+        this.inspectNextBtn.waitAndClick();
+        this.inspectNextBtn.waitAndClick();
+        if (this.inspectConfirmBtn.isDisplayed()) {
+            this.inspectConfirmBtn.waitForDisplayed();
+        }
+        else {
```
andlar: Should be on the previous line

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,81 @@
...
+            this.inspectNextBtn.waitAndClick();
```
andlar: If we get into this block, does that mean the "next" button has been clicked four times? That should never happen, right? It's a maximum of three clicks to get to the confirm screen

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,81 @@
...
+
+    get yesConfirmation () {
+        return $('//button[text()="Yes"]');
+    }
+
+    get rejectBtn () {
+        return $('//table[@id="pending-listings-table"]/tfoot/tr/th/button');
+    }
+
+    gotoConfirmListingPg (inspectListingId ) {
```
andlar: Please spell out the `Page` part of this method name

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,81 @@
+class ConfirmPage {
+    constructor () { }
+
+    get inspectNextBtn () {
```
andlar: Same comment re: `Btn` should be `Button`

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,81 @@
...
+        return $('#inspect-next');
+    }
+
+    get inspectConfirmBtn () {
+        return $('#inspect-confirm');
+    }
+
+    get editcertifiedProduct () {
```
andlar: We usually use the term "listing" to refer to a "certified product", or at least that's the path we're heading towards. That said, with the "certified product", "certified" should be capitalized in this method to match camel-case

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,81 @@
...
+
+    get allTestProcedureVersion () {
+        return $$('*[id^="testProcedures-additional-input"]');
+    }
+
+    get testToolsName () {
+        return $('//*[@id="testTools "]');
+    }
+
+    get alltestToolsVersion () {
```
andlar: Similar camel-case issue

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
```diff
@@ -0,0 +1,106 @@
+class ListingEditPage {
```
andlar: What page is this? the file name doesn't describe where this page is; whether it's a Page or a Component

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
```diff
@@ -0,0 +1,106 @@
...
+    constructor () { }
+
+    get editcertifiedProduct () {
```
andlar: A lot of these elements look the same as in the confirm page-object. Are they the same elements? Can these be combined in some way? The UI code that shows the edit screen, whether in a listing or a pending listing, is the same code, so if that changes both of the page-objects would need to change, and that seems wrong

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
```diff
@@ -0,0 +1,106 @@
...
+
+    get closeListingEditBtn () {
+        return $('button.close.pull-right.ng-isolate-scope');
+    }
+
+    get yesConfirmation () {
+        return $('//button[text()="Yes"]');
+    }
+
+    gotoEditCriteria (editCriteriaId , cures) {
```
andlar: This doesn't "goto" a criteria. This opens a criteria for editing. The name of the method is unclear

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
```diff
@@ -0,0 +1,106 @@
...
+        this.editcertifiedProduct.click();
+        if (cures === true) {
```
andlar: `cures` is a boolean, right? Why not
```suggestion
        if (cures) {
```

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
```diff
@@ -0,0 +1,106 @@
...
+            //click on Edit for on the criteria
+            $('//*[@id="criteria_' + editCriteriaId + '_details_header_cures"]/span[3]/span/button/i').click();
+        }
+        else {
+            $('//*[@id="criteria_' + editCriteriaId + '_details_header"]/span[3]/span/button/i').click();
+        }
+    }
+
+    selectTestProcedures (name , version) {
```
andlar: Does this "add a test procedure" to the list of selected test procedures? If so, "selectTestProcedure" doesn't seem like it describes that

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,63 @@
+import UploadPage from '../../pages/administration/upload/upload.po';
```
andlar: The format we're moving towards for file names is "all lowercase, with `-` separating words"

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,63 @@
...
+import ConfirmPage from '../../pages/administration/confirm/confirm.po';
+import ListingEditPage from '../../pages/searchCHPL/listingEdit/listingEdit.po';
+import LoginComponent from '../../components/login/login.po';
+import Hooks from '../../utilities/hooks';
+import {assert} from 'chai';
```
andlar: Why does "assert" need to be imported? It used to just be a global...

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,63 @@
...
+
+let confirmPage , hooks, listingEditPage, loginComponent, uploadPage;
+
+describe('ACB can', () => {
+    // **Run once before the first test case**
+    before(() => {
```
andlar: Where did "before" come from? Why does it need to be a global while "beforeEach" doesn't?

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,63 @@
...
+        uploadPage = new UploadPage();
+        confirmPage = new ConfirmPage();
+        listingEditPage = new ListingEditPage();
+        loginComponent = new LoginComponent();
+        hooks = new Hooks();
+        hooks.open('/administration/upload');
+        loginComponent.loginAsACB();
+        uploadPage.uploadListing('../../../utilities/2015_v19_AQA1.csv');
+        assert.include(uploadPage.uploadSuccessfulText.getText(),uploadPage.expectedUploadSuccessfulText, 'File has uploaded successfully');
```
andlar: Why is there an assert in a "before" step? Before steps are for setting up the system, not asserting anything, right?

Also, why aren't all of these in the "beforeEach" block? By putting them in this block does that mean that these elements hold state between tests? Holding state between tests is usually regarded as bad form, as that means that tests are not independent of each other

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    it('add test procedure, test data, test tools, test functionality to v19 uploaded listing (170.315 (b)(3) cures criteria)', () => {
+        confirmPage.gotoConfirmListingPg('15.04.04.1722.AQA1.03.01.1.200620');
+        listingEditPage.gotoEditCriteria('170.315 (b)(3)' , true);
+        listingEditPage.selectTestFunctionality('(b)(3)(ii)(B)(4)');
+        listingEditPage.selectTestProcedures('ONC Test Method - Surescripts (Alternative)' , '1');
+        listingEditPage.selectTestData('ONC Test Method' , '2');
+        listingEditPage.selectTestTools('Inferno' , '3');
+        listingEditPage.saveCertifiedProduct.waitAndClick();
+        listingEditPage.viewDetailsCriteria('170.315 (b)(3)' ,true);
+        assert.include($('//*[@id="criteria_170.315 (b)(3)_details_row_Functionality_Tested_cures"]').getText(), 'Request to send an additional supply of medication (Resupply)');
```
andlar: There shouldn't be xpaths in spec files. Xpaths ought to only be in the page-object files

e2e/workflowTest/upload-confirm-test/aqa2_admin_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,63 @@
+import UploadPage from '../../pages/administration/upload/upload.po';
```
andlar: Is this file identical to the ONC-ACB one? If so, I think we need to figure out how to remove the duplicate stuff. If we don't any time we need to change one of these we'd need to change the other, and that doesn't make sense

package.json
```diff
@@ -44,6 +44,7 @@
     "angulartics": "^1.6.0",
     "angulartics-google-tag-manager": "^0.0.1",
     "bootstrap-sass": "3.4.1",
+    "chai": "^4.2.0",
```
andlar, line 47: Why is this being added?

andlar (changes requested):

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
palakpatel311: This is Mocha.hookfunction.
BeforeEach shows as jasmine. 
Not sure how to make before added as defined

e2e/workflowTest/upload-confirm-test/aqa2_admin_add_remove_testProcDataToolFunc.spec.js
palakpatel311: Yes it is. I have kept it seperate for now as we have on of the ticket which doesnt allow ACBs to remove test functionality and allows admin

package.json
palakpatel311: This is where assert should be from when you asked me in other PR. so chai needs to be added.

README.md
```diff
@@ -62,3 +62,7 @@ On all Yarn commands the CSS Linter, JS Linter and HTML Linters will run. Webpac
 [nodejs]: https://nodejs.org/en/download/
 [yarn]: https://yarnpkg.com/en/
 [webpack]: https://webpack.js.org/
+
+#### Automation credetials set up
+
+Rename e2e/config/credentialsEXAMPLE.js to e2e/config/credentials.js And Add Usernames and password in it.
```
andlar, line 76: The EXAMPLE file doesn't have any content in it. Without a sample of what's expected it's hard to know what to put in it

e2e/components/login/login.po.js
```diff
@@ -14,10 +19,47 @@ class LoginComponent {
...
+        if (this.usernameInput.isDisplayed()) {
+            this.usernameInput.addValue(credentials.usernameACB);
+        }
+        else {
+            this.openLoginComponent();
+            this.usernameInput.addValue(credentials.usernameACB);
```
andlar: This "addValue" happens in both branches of the `if`. Given that, it ought to be outside of the `if`

e2e/components/login/login.po.js
```diff
@@ -14,10 +19,47 @@ class LoginComponent {
...
+        }
+        this.passwordInput.addValue(credentials.passwordACB);
+        this.loginButton.click();
+        return this;
+    }
+
+    loginAsAdmin () {
+        this.usernameInput.addValue(credentials.usernameAdmin);
```
andlar, line 53: Why doesn't this `loginAsUser` function have the `if` construct from the `loginAsACB` function?

e2e/components/login/login.spec.js
```diff
@@ -6,14 +6,37 @@ let component, hooks;
 beforeEach(async () => {
     component = new LoginComponent();
     hooks = new Hooks();
-    await hooks.open('/resources/overview');
+    await hooks.open('/administration/upload');
 });
 
-describe('Login page', () => {
+describe('ACB can', () => {
```
andlar: I try to write my `describe` and `it` blocks in such a way that they make sentences. Describe blocks should look like a description of the state of the system / user doing something / something like that. So this one could be `describe('an ACB user'` 

It blocks should be written as `it('should do a thing'`, so the `login successfully` one would be `it('should be able to log in`

The result would be test specs that read "an ACB user ... should be able to log in"

e2e/components/login/login.spec.js
```diff
@@ -6,14 +6,37 @@ let component, hooks;
...
+    it('login successfully', () => {
+        component.loginAsACB();
+        component.logoutButton.waitForDisplayed();
+        assert.equal(component.logoutButton.isDisplayed(),true);
+    })
+
+    it('logsout successfully', () => {
+        component.logOut();
+        component.loginButton.waitForDisplayed();
+        assert.equal(component.loginButton.isDisplayed(),true);
```
andlar: Why isn't this just `assert.isTrue`? Using `assert.equal` when testing against a boolean is like writing `if (value === true) {`

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,51 @@
...
+    get inspectConfirmButton () {
+        return $('#inspect-confirm');
+    }
+
+    get yesConfirmation () {
+        return $('//button[text()="Yes"]');
+    }
+
+    get rejectButton () {
+        return $('//table[@id="pending-listings-table"]/tfoot/tr/th/button');
```
andlar: I don't know what our standard should be, but in the other files these various locator values are defined as constants / data / something at the top of the file. I'm not entirely sure I _like_ that way of doing things, but that is how we started doing it. Whether we go that way or move to inline locators is maybe worth a conversation, but until then keeping consistent with defining the locators at the top of the file matches our pattern

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,51 @@
...
+    gotoConfirmListingPg (inspectListingId ) {
+        $('//button[@id="pending-listing-inspect-' + inspectListingId + '"]').waitAndClick();
+        this.inspectNextButton.waitAndClick();
+        this.inspectNextButton.waitAndClick();
+        this.inspectNextButton.waitAndClick();
+        if (this.inspectConfirmButton.isDisplayed()) {
+            this.inspectConfirmButton.waitForDisplayed();
+        }
+        else {
+            this.inspectNextButton.waitAndClick();
```
andlar, line 35: Is this call going to click the "next" button a fourth time? Isn't that impossible?

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,51 @@
...
+    findListingtoReject (chplId) {
+        return $('//td[text()="' + chplId + '"]');
+    }
+
+    rejectListing (chplId) {
+        $('//td[text()="' + chplId + '"]/following-sibling::td[7]/input').waitAndClick();
+        if (this.rejectButton.isClickable()) {
+            this.rejectButton.waitAndClick();
+        }
+        else {
```
andlar: Should be `} else {`

e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -0,0 +1,27 @@
+import ConfirmPage from './confirm.po';
+import UploadPage from '../upload/upload.po';
+import LoginComponent from '../../../components/login/login.po';
+import Hooks from '../../../utilities/hooks';
+import {assert} from 'chai';
```
andlar: The `assert` is defined in the wdio.conf.js file in a before block. If we're defining that there it shouldn't need to be here too

e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    uploadPage.uploadListing('../../../utilities/2015_v19_AQA1.csv');
+});
+
+describe('Confirm page', () => {
+
+    it('allows user to reject a file', () => {
+        hooks.open('/administration/confirm/listings');
+        confirmPage.rejectListing('15.04.04.1722.AQA1.03.01.1.200620');
+        assert.equal(confirmPage.findListingtoReject('15.04.04.1722.AQA1.03.01.1.200620').isDisplayed(),false)
+    })
```
andlar: These blocks should have a `;` at the end of them

e2e/pages/administration/upload/upload.po.js
```diff
@@ -0,0 +1,32 @@
+const uploadElements = {
+    chooseUploadListing: '//*[@id="ngf-label-upload-button-listing"]/input[@id="ngf-upload-button-listing"]',
+    uploadButton: '//*[@id="main-content"]/div/ui-view/chpl-upload/div/div/chpl-upload-listings/div/div[2]/form/div/div[4]/button[1]',
+    uploadSuccessfulText: '//*[@id="main-content"]/div/ui-view/chpl-upload/div/div/chpl-upload-listings/div/div[2]/div',
+}
+
+class UploadPage {
+    constructor () { }
+
+    get ChooseUploadListingButton () {
```
andlar: Should be camel-case, not pascal-case

andlar (changes requested):

package.json
andlar: It didn't need to be added before, when it was required in the wdio.conf.js file. If it didn't need to be added to work then, why does it need to be added here?

palakpatel311: when i wanted to user other asserts like include, it did not detect chai and thats when i saw that chai is not even added in package.json. It worked earlier my guess is assert was coming from some other package or inbuilt.

README.md
palakpatel311: Changed back to credentials file

e2e/workflowTest/upload-confirm-test/aqa2_admin_add_remove_testProcDataToolFunc.spec.js
palakpatel311: Removed admin user test for now. I tried to do data providers which could be used to handle same TC with different credentials but it did not work the way i wanted it to be. I need some time to figure out this as this will be framework setting up for future.

e2e/components/login/login.po.js
palakpatel311: Fixed

e2e/components/login/login.po.js
palakpatel311: Fixed

e2e/components/login/login.spec.js
palakpatel311: Fixed.
Ashwini prefers not using should in the TC summary and thats what I was applying here.
But i like Should way.

e2e/components/login/login.spec.js
palakpatel311: Fixed

e2e/pages/administration/confirm/confirm.po.js
palakpatel311: Sometimes, selenium is clicking so fast that it was missing one of the next button thats why i had to add this if

e2e/pages/administration/confirm/confirm.spec.js
palakpatel311: Agree. Removed from all spec files

e2e/pages/administration/confirm/confirm.spec.js
palakpatel311: Fixed

e2e/pages/administration/upload/upload.po.js
palakpatel311: Fixed

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
```diff
@@ -0,0 +1,142 @@
...
+
+    get allTestProcedureVersion () {
+        return $$('*[id^="testProcedures-additional-input"]');
+    }
+
+    get testToolsName () {
+        return $('//*[@id="testTools "]');
+    }
+
+    get alltestToolsVersion () {
```
tmy1313: Should this be `allTestToolsVersion ()`?

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
```diff
@@ -0,0 +1,142 @@
...
+        return $$('*[id^="testTools-additional-input"]');
+    }
+
+    get testDataName () {
+        return $('//*[@id="testData "]');
+    }
+
+    get alltestDataVersion () {
```
tmy1313: Should this be `allTestDataVersion ()`?

e2e/pages/administration/upload/upload.po.js
```diff
@@ -0,0 +1,32 @@
+const uploadElements = {
+    chooseUploadListing: '//*[@id="ngf-label-upload-button-listing"]/input[@id="ngf-upload-button-listing"]',
+    uploadButton: '//*[@id="main-content"]/div/ui-view/chpl-upload/div/div/chpl-upload-listings/div/div[2]/form/div/div[4]/button[1]',
+    uploadSuccessfulText: '//*[@id="main-content"]/div/ui-view/chpl-upload/div/div/chpl-upload-listings/div/div[2]/div',
+}
+
```
tmy1313, line 6: Question - if these are defined here, aren't they global (when the file is imported)?  Is that what we want?  Should these strings be defined with in the class?

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -0,0 +1,56 @@
+const confirmElements = {
+    inspectNext: '#inspect-next',
+    inspectConfirm: '#inspect-confirm',
+    yesConfirmation: '//button[text()="Yes"]',
+    rejectButton: '//table[@id="pending-listings-table"]/tfoot/tr/th/button',
+}
```
tmy1313, line 6: Question - if these are defined here, aren't they global (when the file is imported)? Is that what we want? Should these strings be defined with in the class?

e2e/components/login/login.po.js
```diff
@@ -1,6 +1,11 @@
+import credentials from '../../config/credentials.js';
+
 const loginElements = {
-    loginButton: '#login-toggle',
+    loginButton: '//*[@id="login-toggle"]',
     userName: '[name="username"]',
+    password: '[name="password"]',
+    login: 'button=Log In',
+    logout: '//button[text()="Log Out"]',
 }
```
tmy1313, line 10: Question - if these are defined here, aren't they global (when the file is imported)? Is that what we want? Should these strings be defined with in the class?

e2e/.eslintrc
```diff
@@ -8,7 +8,11 @@
         "element": false,
         "by": false,
         "$": false,
-        "$$": false
+        "$$": false,
+        "process": false,
```
andlar: I don't see `process` used anywhere in the changed code. I also don't see either `before` or `after`

e2e/components/login/login.po.js
```diff
@@ -14,10 +19,47 @@ class LoginComponent {
...
+        return $(loginElements.logout);
+    }
+
     openLoginComponent () {
         this.toggleLoginComponent.click();
         return this;
     }
+
+    loginAsACB () {
+        if (this.usernameInput.isDisplayed() === false) {
```
andlar: Checking a variable against a boolean value inside an `if` test is redundant. Should just `if` the variable (with or without a `!` as necessary

```suggestion
        if (!this.usernameInput.isDisplayed()) {
```

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        listingEditPage.saveCertifiedProduct.waitAndClick();
+        listingEditPage.viewDetailsCriteria('170.315 (b)(3)' ,true);
+        assert.include(listingEditPage.getTestFunctionalityDetail('170.315 (b)(3)' ,true).getText(), 'Request to send an additional supply of medication (Resupply)');
+        assert.include(listingEditPage.getTestProcedureDetail('170.315 (b)(3)' ,true).getText(), 'ONC Test Method - Surescripts (Alternative)');
+        assert.include(listingEditPage.getTestDataDetail('170.315 (b)(3)' ,true).getText(), 'ONC Test Method');
+        assert.include(listingEditPage.getTestToolDetail('170.315 (b)(3)' ,true).getText(), 'Inferno');
+        listingEditPage.closeEditListing();
+    })
+
+    it('should be able to remove uploaded test procedure, test data, test tools (170.315 (b)(3) cures criteria)', () => {
```
andlar: What happens if this test runs before the one above it? If this test depends on the previous one happening then the tests are dependent, and that's something we really want to avoid

wdio.conf.js
```diff
@@ -123,13 +126,23 @@ exports.config = {
...
+        helpers: [require.resolve('@babel/register')],
+        expectationResultHandler: function(passed, assertion) {
+            /**
+             * only take screenshot if assertion failed
+             */
+            if(passed) {
+                return
+            }
+    
+            browser.saveScreenshot(`assertionError_${assertion.error.message}.png`)
```
andlar: This saves the output screenshots in the root directory, which then means they'll get caught up by git eventually, or need to be gitignored. Usual best practices would be to output all test results (including errors & messages) into whatever directory our other test output stuff goes in. For this project I think it's `test_reports`

andlar (changes requested):

e2e/pages/administration/upload/upload.po.js
andlar: I don't think these javascript files are imported directly as is. They're also not "smushed" together to make one big file either, so I think these consts don't land on the global scope. By exporting the "UploadPage" class, and then importing that class specifically, I believe that keeps the consts clear

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
palakpatel311: Fixed

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
palakpatel311: Fixed

e2e/pages/administration/upload/upload.po.js
palakpatel311: Agree. We are not importing constants here to spec files.

e2e/.eslintrc
palakpatel311: Fixed

e2e/components/login/login.po.js
palakpatel311: Fixed

e2e/workflowTest/upload-confirm-test/aqa1_acb_add_remove_testProcDataToolFunc.spec.js
palakpatel311: it blocks runs in sequence

wdio.conf.js
palakpatel311: Fixed

e2e/components/login/login.po.js
andlar: The comment says "fixed", but I don't see a change in the code. Am I missing something? There are a few other comments like that through the PR

e2e/pages/searchCHPL/listingEdit/listingEdit.po.js
palakpatel311: Moved it from confirm page

e2e/.eslintrc
andlar: I thought I saw this change, but don't see it now, and don't know how I could be missing it

e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    loginComponent.loginAsACB();
+    uploadPage.uploadListing('../../../resources/2015_v19_AQA1.csv');
+});
+
+describe('Confirm page', () => {
+
+    it('allows user to reject a file', () => {
+        hooks.open('#/administration/confirm/listings');
+        confirmPage.rejectListing('15.04.04.1722.AQA1.03.01.1.200620');
+        assert.equal(confirmPage.findListingtoReject('15.04.04.1722.AQA1.03.01.1.200620').isDisplayed(),false)
```
andlar: I think this ought to be either `assert.isFalse` if that exists, or something like `assert.isTrue(!confirmPage.find...` Or maybe there's a matcher / assertion that's the opposite of `.isDisplayed()`?

e2e/pages/administration/upload/upload.po.js
```diff
@@ -0,0 +1,32 @@
...
+        return $(uploadElements.uploadSuccessfulText);
+    }
+
+    uploadListing (uploadfilePath) {
+        const path = require('path');
+        const filePath = path.join(__dirname, uploadfilePath);
+        const remoteFilePath = browser.uploadFile(filePath);
+        this.chooseUploadListingButton.addValue(remoteFilePath);
+        this.uploadButton.waitAndClick();
+        browser.waitUntil( () => this.uploadSuccessfulText.isDisplayed() === true)
```
andlar: There are a few places where there are `...something === true` or `...something === false`. All of those should have the ` === boolean` part removed

e2e/pages/listing/listingEdit.po.js
```diff
@@ -0,0 +1,142 @@
...
+        return $('button.close.pull-right.ng-isolate-scope');
+    }
+
+    get yesConfirmation () {
+        return $('//button[text()="Yes"]');
+    }
+
+    gotoEditCriteria (editCriteriaId , cures) {
+        this.editcertifiedProduct.click();
+        if (cures === true) {
```
andlar: Here are a few more of the `...something === true` things

e2e/workflowTest/upload-confirm-test/aqa1_add_remove_testProcDataToolFunc.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        listingEditPage.selectTestProcedures('ONC Test Method - Surescripts (Alternative)' , '1');
+        listingEditPage.selectTestData('ONC Test Method' , '2');
+        listingEditPage.selectTestTools('Inferno' , '3');
+        listingEditPage.saveCertifiedProduct.waitAndClick();
+        listingEditPage.viewDetailsCriteria('170.315 (b)(3)' ,true);
+        assert.include(listingEditPage.getTestFunctionalityDetail('170.315 (b)(3)' ,true).getText(), 'Request to send an additional supply of medication (Resupply)');
+        assert.include(listingEditPage.getTestProcedureDetail('170.315 (b)(3)' ,true).getText(), 'ONC Test Method - Surescripts (Alternative)');
+        assert.include(listingEditPage.getTestDataDetail('170.315 (b)(3)' ,true).getText(), 'ONC Test Method');
+        assert.include(listingEditPage.getTestToolDetail('170.315 (b)(3)' ,true).getText(), 'Inferno');
+        listingEditPage.closeEditListing();
```
andlar: On these two tests, especially for the first of our workflow ones, I'd like to make sure we set them up correctly to start. The things I see that concern me:

- Why do anything _after_ the last assert in a give `it` block? If we're doing something, it either should cause a change we need to assert, or it's a sign that we're doing something outside of tests
- When the first line of an `it` block is `browser.refresh()` it makes me wonder why that's needed. That suggests that _something_ happened before the test itself that needs to be handled, but if that's the case then wherever that thing is that's causing that to be necessary needs to handle it
- In the `beforeAll` there's a single "upload a listing" thing. If we only upload it once, for the entire suite of tests in this spec, then it seems likely those tests will interact with each other in ways we can't predict. It makes sense to log in once, but it seems likely we'd need to upload the file for every test

andlar (changes requested):
I think we're really close here. I have a couple of comments about comparing to booleans which ought to be easy to resolve. The only thing that might be more complicated is in the workflow test file

e2e/workflowTest/upload-confirm-test/aqa1_add_remove_testProcDataToolFunc.spec.js
palakpatel311: 1. listingEditPage.closeEditListing(); is needed as between two it block we dont open new browser. browser is same and assertion is happening on the pop up so in order for browser to go back to main screen /administration/confirm...... I will have to close pop up. What i could do is move that to afterEach function so it serves 2 purposes.
3. Moved upload a listing to before Each to handle randomly running TC.

Rearraged a bit. Check and let me know

e2e/pages/administration/upload/upload.po.js
palakpatel311: Fixed

e2e/pages/administration/confirm/confirm.spec.js
palakpatel311: Fixed

e2e/pages/listing/listingEdit.po.js
palakpatel311: Fixed

andlar (dismissed):

tmy1313 (dismissed):

andlar (approved):

tmy1313 (approved):

PR #784, opened 2020-08-03
Ocd 3437 - Add "Ignore warnings" checkbox when editing or confirming certified product

tmy1313:
OCD-3437


Plow74 (dismissed):

src/app/components/listing/inspect/inspect.html
```diff
@@ -54,6 +54,10 @@ <h4 class="modal-title" id="inspect-label">Inspect Certified Product ({{ vm.cp.c
           </ul>
         </div>
       </div>
+      <div ng-if="vm.stage === 'cp' && vm.cp.warningMessages && vm.cp.warningMessages.length !== 0" class="col-sm-12 text-left">
+        <input type="checkbox" ng-model="vm.acknowledgeWarnings" id="acknowledge-warnings" class="edit-listing__acknowledge-warning-checkbox">
+        <label for="acknowledge-warnings" class="edit-listing__acknowledge-warning-label">I have reviewed the warning(s) above and wish to proceed with this update</label>
```
andlar: For any of the places where we have things like "warning(s)" in the rest of the system, if we can tell if the number of things is !== 1, I've used an `ng-if` to append or not append the "s". Request doing that here

src/app/components/listing/edit.html
```diff
@@ -429,12 +429,16 @@
...
   </div>
-  <div class="text-danger text-left" ng-if="$ctrl.messages.errors">
+  <div class="text-danger text-left" ng-if="$ctrl.messages.errors && $ctrl.messages.errors.length > 0">
     The following error(s) must be fixed:
     <ul>
       <li ng-repeat="msg in $ctrl.messages.errors | orderBy:'toString()' track by $index">{{ msg }}</li>
     </ul>
   </div>
+  <div ng-if="$ctrl.messages.warnings && $ctrl.messages.warnings.length > 0">
+    <input type="checkbox" ng-model="$ctrl.acknowledgeWarnings" id="acknowledge-warnings" class="edit-listing__acknowledge-warning-checkbox">
```
andlar: Should have a name value as well, which is camelCase, to be able to use it in the `ngForm[name]` thing as needed

src/app/components/listing/edit.component.js
```diff
@@ -21,6 +21,7 @@ export const ListingEditComponent = {
             this.hasAnyRole = authService.hasAnyRole;
             this.networkService = networkService;
             this.utilService = utilService;
+            this.acknowledgeWarnings = false;
```
andlar: I don't usually set things to false in constructors, when `undefined` evaluates as falsy anywhere it needs to be checked against. Not necessarily a required change, but I tend to try to not have lines of code that aren't needed

src/app/admin/components/certifiedProduct/listing/edit.scss
```diff
@@ -0,0 +1,11 @@
+label.edit-listing__acknowledge-warning-label {
```
andlar: Can we remove this entire file? CSS is global, and these styles are defined in one of the components directories identically. Part of the path forward is to remove the entirety of the `/admin/*` stuff eventually, so not having duplicate styles would help with that

src/app/admin/components/certifiedProduct/listing/edit.controller.js
```diff
@@ -30,6 +30,7 @@
         vm.requiredIcsCode = requiredIcsCode;
         vm.save = save;
         vm.updateListing = updateListing;
+        vm.acknowledgeWarnings = false;
```
andlar: Same comment re: constructor defaults of false

andlar (changes requested):

andlar (dismissed):
Approving the PR, though there's a conflict in the yarn.lock file. I don't think that should hold up moving forward with testing

Plow74 (dismissed):

yarn.lock
```diff
@@ -1268,10 +1278,17 @@
   dependencies:
     "@wdio/logger" "6.0.16"
 
+<<<<<<< HEAD
```
andlar: This doesn't look right...

andlar (changes requested):

andlar (dismissed):

src/app/components/listing/inspect/inspect.controller.js
```diff
@@ -88,20 +88,23 @@
...
-                            objectId: error.data.objectId,
-                            status: 'resolved',
-                        });
-                    } else {
-                        vm.errorMessages = error.data.errorMessages;
-                    }
-                });
+            networkService.confirmPendingCp({
+                pendingListing: vm.cp,
+                warningAcknowledgement: vm.acknowledgeWarnings,
```
kekey1: Either go with "acknowledgeWarnings" as in listing edit or change listing edit to match this one. They seem like they should be the same.

kekey1 (changes requested):

andlar (approved):

kekey1 (approved):

PR #787, opened 2020-08-17
OCD-3203: Allow Product editing on Developer page

andlar:


ashwinimore (dismissed):
AQA review

src/app/pages/organizations/developers/developer/view.component.js
```diff
@@ -144,8 +151,44 @@ export const DevelopersViewComponent = {
...
+            this.networkService.updateProduct(request)
+                .then(response => {
+                    if (!response.status || response.status === 200 || angular.isObject(response.status)) {
+                        this.$state.go('organizations.developers.developer', {
+                            developerId: this.developer.developerId,
+                            action: undefined,
+                            productId: undefined,
+                        }, {reload: true});
+                    } else {
+                        if (response.data.errorMessages) {
```
Plow74, line 171: Just noting the duplicate code.

Plow74, line 171: I know this isn't an exact duplicate but could be close enough to consider de-duplicating if possible.

Plow74 (dismissed):

e2e/components/login/login.po.js
```diff
@@ -43,7 +42,6 @@ class LoginComponent {
         this.usernameInput.addValue(credentials.usernameACB);
         this.passwordInput.addValue(credentials.passwordACB);
         this.loginButton.click();
-        return this;
     }
 
     loginAsAdmin () {
```
palakpatel311, line 47: When should we take these TCs out and refactor code wherever it is used since we have new ones now?

e2e/components/login/login.readonly.spec.js
```diff
@@ -40,3 +40,39 @@ describe('an Admin user', () => {
...
 })
+
+describe('when logging in', () => {
+    afterEach(() => {
+        component.logOut();
+    });
+
+    it('should be able to log in as ACB', () => {
+        component.logIn('acb');
+        component.logoutButton.waitForDisplayed();
```
palakpatel311, line 47: Do we really need here for logout button to be displayed?
I think getting text from togglelogincomponent will work without this wait

e2e/components/login/login.po.js
andlar: I didn't want to remove the old ones because there are a bunch of things currently using them now, and some that are in progress. Once the updated methods get into staging then we can refactor anything using the old ones to use the new ones and go from there. Basically, I wanted to have both possibilities available for now, and migrate when we can

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -1,10 +1,13 @@
+/* eslint-disable no-console,angular/log */
 import DevelopersPage from './developers.po';
 import Hooks from '../../../utilities/hooks'
 
 let hooks, page;
 
 describe('the Developers page', () => {
     beforeEach(async () => {
+        browser.setWindowSize(1600, 1024); // demo of a bigger screen (esp. useful for screenshots)
+        browser.setWindowRect(0, 0, 1600, 1024); // not sure if both are required
```
palakpatel311, line 10: Setwindowsize is been added in hooks which will apply to all test. You can remove setWindowRect and setWindowSize from here

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -17,11 +20,33 @@ describe('the Developers page', () => {
...
+
+        describe('when looking at a specific Product', () => {
+            let name = 'MediaDent';
+            let product;
+            beforeEach(() => {
+                product = page.getProduct(name);
+                product.scrollIntoView({block: 'center', inline: 'center'});
+                browser.waitUntil(() => page.getVersionCount(product).getText() === '12 Versions');
+                page.selectProduct(product);
+                page.getProductInfo(product).waitForDisplayed({timeout: 55000});
```
palakpatel311, line 45: Can you add timeout in config file. I have added one as longtimeout for one of my test

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -0,0 +1,122 @@
...
+                        product.scrollIntoView({block: 'center', inline: 'center'});
+                        browser.waitUntil(() => page.getVersionCount(product).getText() === '1 Version');
+                        page.selectProduct(product);
+                        page.getProductInfo(product).waitForDisplayed({timeout: 55000});
+                        expect(contact.get(product)).toEqual(poc);
+                    });
+                });
+            });
+
+            describe('when planning to change a Product name', () => {
```
palakpatel311, line 87: I wonder if we need this describe? 
This describe block has only one describe block and both has similar name

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -0,0 +1,122 @@
...
+                        actionBar.save();
+                        page.productsHeader.waitForDisplayed();
+                        toast.clearAllToast();
+                        product = page.getProduct(newName);
+                        product.scrollIntoView({block: 'center', inline: 'center'});
+                        browser.waitUntil(() => page.getVersionCount(product).getText() === '1 Version');
+                        page.selectProduct(product);
+                        page.getProductInfo(product).waitForDisplayed({timeout: 55000});
+                        page.editProduct(product);
+                        page.editProductsHeader.waitForDisplayed();
```
palakpatel311, line 114: Can we make a method of editing product as whole. I see some of the same steps in this and the above one
                        page.productsHeader.waitForDisplayed();
                        toast.clearAllToast();
                        product.scrollIntoView({block: 'center', inline: 'center'});
                        browser.waitUntil(() => page.getVersionCount(product).getText() === '1 Version');
                        page.selectProduct(product);
                        page.getProductInfo(product).waitForDisplayed({timeout: 55000});

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -0,0 +1,122 @@
...
+                        expect(page.editProductName).toBeDisplayed();
```
palakpatel311, line 115: Is it updating this fast?
I was expecting to see some spinner wait in developer dashboard test but I havent seen one.

e2e/components/login/login.readonly.spec.js
andlar: I was following the pattern of the login specs from the original set. I don't know if it's necessary to wait, but this makes it clear that we're doing so

e2e/pages/organizations/developers/developers.readonly.spec.js
andlar: That might be true in whatever PR has that change, but this code doesn't have those hooks. When this code and that code merge then these could be removed, but they can't yet

e2e/pages/organizations/developers/developers.readonly.spec.js
andlar: I can't use the timeout that's in the config because that's in a different branch that this code doesn't have access to. I _could_ change this to be a constant or another config value, but I'd rather wait until this code gets merged with that one to enable replacing all of these with something standard

e2e/pages/organizations/developers/developers.spec.js
andlar: By putting this describe here it the two "edit product" blocks have the same structure, which shows up well in the spec output. It can't be combined with the other block because changing a Product name means we can't reselect it, so it's the only test that can be done wrt that edit during an entire AQA run

e2e/pages/organizations/developers/developers.spec.js
andlar: I don't want to combine these. One of the "Clean Code" principles is that we want to keep our methods small, short, and have them do one thing only. Combining all of these into a single "loadProductAndWaitForItAndEditProductAndWaitForItToBeReady" method would not make the code any more clear, and also make that one method only usable for that one activity, whereas keeping them as separate methods mean they can be mixed and matched for other uses

e2e/pages/organizations/developers/developers.spec.js
andlar: While the spinner might show up, the update happens fast enough that I don't need to wait for the spinner to disappear. The tests _do_ pass. Also, the "wait for the spinner" code is in that other branch, so I don't have access to it here anyway

e2e/pages/organizations/developers/developers.spec.js
palakpatel311: Can we rephrase the wording then? both describe says same

e2e/pages/organizations/developers/developers.readonly.spec.js
palakpatel311: Okay doing it later when that code gets in makes sense

e2e/components/login/login.readonly.spec.js
palakpatel311: If code works without this waitfordisplayed, I would remove it. I dont think its necessary to add.

e2e/pages/organizations/developers/developers.spec.js
palakpatel311: yes why cant we have load product and edit product as one?
Edit product will cover selecting one product and editing?

e2e/components/contact/contact.po.js
```diff
@@ -0,0 +1,62 @@
+const elements = {
+    container: 'chpl-contact',
+    editFull: '#contactFull',
```
kekey1, line 3: All the places where it says "Full" - I feel like the naming is a little weird. Full what? Can we just make it say FullName?

kekey1 (dismissed):
One super small comment, not worth holding up approval for

e2e/components/contact/contact.po.js
andlar: The HTML has it with the ID `contactFull`. I could update the HTML, but am trying to do as little to the HTML when working on AQA stuff as possible. I'm sort of trying to treat the AQA work as separate from the Feature work as a matter of principle. I will add missing IDs / classes / etc. if required, but am generally trying to avoid reworking stuff that already exists and works

e2e/pages/organizations/developers/developers.spec.js
tmy1313: I think the idea here is that we make small, atomic functions, and then compose larger methods that perform the specific set of steps that are necessary to meet the needs of the test.  The idea isn't always to make re-usable code, but to make more readable and maintainable code.  I think by combining these two ideas into a single concept we would also be breaking the functions "do one thing" principal of Clean Code.

palakpatel311 (dismissed):

e2e/components/login/login.readonly.spec.js
andlar: It doesn't work without this line
image

kekey1 (approved):

tmy1313 (approved):

PR #788, opened 2020-08-19
OCD 3437: Test for ignore warnings checkbox on upload

palakpatel311:
image


e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -12,15 +14,16 @@ beforeEach(() => {
...
+    uploadPage.waitForSuccessfulUpload();
 });
 
 describe('Confirm page', () => {
 
     it('allows user to reject a file', () => {
         hooks.open('#/administration/confirm/listings');
-        confirmPage.rejectListing('15.04.04.1722.AQA1.03.01.1.200620');
-        assert.isFalse(confirmPage.findListingtoReject('15.04.04.1722.AQA1.03.01.1.200620').isDisplayed())
+        confirmPage.rejectListing('15.04.04.1722.AQA3.03.01.1.200620');
```
Plow74: Consider putting these values in a const. ex: `const LISTING_ID_TO_REJECT = 15.04.04.1722.AQA3.03.01.1.200620`. It could improve readability and maintenance since the id seems to be hard coded in 2 places.

e2e/workflowTest/upload-confirm-test/bypass_warnings.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+    beforeAll(function () {
+        hooks.open('#/administration/upload');
+        uploadPage.uploadListing('../../../resources/2015_v19_AQA4.csv');
+        uploadPage.waitForSuccessfulUpload();
+        hooks.open('#/administration/confirm/listings');
+    })
+
+    it('should show warning bypass checkbox when upload listing file has warnings', () => {
+        confirmPage.gotoConfirmListingPage('15.04.04.1722.AQA4.03.01.1.200620');
```
Plow74: Same comment as above. Putting these sort of reused or ambiguous (to the first-time reader) strings may be good candidates for descriptive variable names and prevent an error because the string is defined at a single point.

e2e/workflowTest/upload-confirm-test/bypass_warnings.spec.js
```diff
@@ -0,0 +1,56 @@
...
+        hooks.open('#/administration/confirm/listings')
+    })
+
+    it('should not show warning bypass checkbox when upload listing file doesnt have warnings', () => {
+        confirmPage.gotoConfirmListingPage('15.04.04.1722.AQA1.03.01.1.200620');
+        assert.isFalse(confirmPage.warningCheckbox.isDisplayed());
+    })
+})
+
+describe('Confirming a listing', () => {
```
andlar: Why do both this and the previous describe blocks have the same description?

e2e/workflowTest/upload-confirm-test/bypass_warnings.spec.js
```diff
@@ -0,0 +1,56 @@
...
+    // **Run once before each test case**
+    beforeEach(function () {
+        uploadPage.uploadListing('../../../resources/2015_v19_AQA1.csv');
+        uploadPage.waitForSuccessfulUpload();
+        hooks.open('#/administration/confirm/listings')
+    })
+
+    it('should not show warning bypass checkbox when upload listing file doesnt have warnings', () => {
+        confirmPage.gotoConfirmListingPage('15.04.04.1722.AQA1.03.01.1.200620');
```
andlar: Does this method "confirm" the listing? My reading of the code suggests it doesn't. If it does actually confirm the listing then the name needs to indicate that. If it doesn't confirm the listing then there should never be a checkbox. The checkbox should only appear if there are warnings on the confirmation action. If this does show the checkbox without attempting to confirm, then we'll need to change the feature code

wdio.conf.js
```diff
@@ -236,8 +244,9 @@ exports.config = {
     /**
      * Function to be executed after a test (in Mocha/Jasmine).
      */
-    // afterTest: function(test, context, { error, result, duration, passed, retries }) {
-    // },
+    afterTest: function(test, context, { error, result, duration, passed, retries }) {
+        browser.refresh();
```
andlar: Wait, why are we refreshing the page after a test? This is a global action and I don't want to make a global change without talking about it

e2e/components/login/login.readonly.spec.js
```diff
@@ -6,7 +6,7 @@ let component, hooks;
 beforeEach(async () => {
     component = new LoginComponent();
     hooks = new Hooks();
-    await hooks.open('#/resources/overview');
+    await hooks.open('#/administration/upload');
```
andlar: Why go to this page? Not all users will even be able to get to this page, but all of them can get to the resources/overview one

andlar (changes requested):

wdio.conf.js
palakpatel311: 1. When we are editing an uploaded listing, it opens pop up, once one it block test is done, that pop up is still open and i had to add after method to close that pop up in order to start next it block.
2. It makes sense to start clean on each it block so refresh makes screen clear after each it block execution.

FYI: it block runs sequentially not parallel. I verified.

e2e/pages/administration/confirm/confirm.spec.js
palakpatel311: Fixed

e2e/workflowTest/upload-confirm-test/bypass_warnings.spec.js
palakpatel311: Fixed

e2e/workflowTest/upload-confirm-test/bypass_warnings.spec.js
palakpatel311: Updated all test cases after our call

e2e/workflowTest/upload-confirm-test/bypass_warnings.spec.js
palakpatel311: Updated all test cases after our call

e2e/components/login/login.readonly.spec.js
palakpatel311: Updated back to overview page

e2e/components/login/login.readonly.spec.js
```diff
@@ -6,7 +6,7 @@ let component, hooks;
 beforeEach(async () => {
     component = new LoginComponent();
     hooks = new Hooks();
-    await hooks.open('#/resources/overview');
+    await hooks.open('#/administration/overview');
```
andlar: I don't think this page exists

andlar (changes requested):

wdio.conf.js
andlar: Cleaning up after a test makes sense to be in the test cases that make a mess, not in a global hook that runs after every test. This is a step that will add time to every test, but will only be needed on a few tests

If "starting clean" is required, then whatever describe block needs to start clean should have the refresh call in it, not globally

I agree that "it" blocks run sequentially, and not in parallel. That said, each individual "it" block should be able to run and pass whether or not any other "it" block has run. Each "it" block is an individual test, and if a test depends on a different test having been executed first, that adds interrelationships between tests that shouldn't be there

palakpatel311: I ll update browser.refresh in test spec when we deal with edit screen. and again this will be temporary until we get new edit screen.

Agree on It block. I have it set up as independent for this test.

e2e/components/login/login.readonly.spec.js
palakpatel311: Updated

Plow74 (dismissed):

andlar (dismissed):

ashwinimore (dismissed):

athomas-git (dismissed):

andlar (approved):

ashwinimore (approved):

PR #789, opened 2020-08-19
OCD-3049: back to back api documentation upload should be successful

palakpatel311:
[#OCD-3049]


e2e/pages/administration/upload/upload.po.js
```diff
@@ -1,8 +1,14 @@
 const uploadElements = {
     chooseUploadListing: '//*[@id="ngf-label-upload-button-listing"]/input[@id="ngf-upload-button-listing"]',
-    uploadButton: '//*[@id="main-content"]/div/ui-view/chpl-upload/div/div/chpl-upload-listings/div/div[2]/form/div/div[4]/button[1]',
+    uploadButton: '.btn.btn-ai-success',
```
tmy1313, line 5: Should this button be updated to have an ID?  I could see a scenario where we add another button to the puge using the class `btn-ai-success`.

tmy1313 (changes requested):

e2e/pages/administration/upload/upload.po.js
andlar: I think the better option would be for this page to not have the upload components at all. Create a new `e2e/components/**/upload-api-documentation` component and put this selector in that component. Then the administration-page can import that upload-api-documentation-component object and use the given `.btn.btn-ai-success` selector, because then there will only be one, if the component is scoped to the `<chpl-upload-api-documentation>` element

e2e/pages/administration/upload/upload.spec.js
```diff
@@ -4,24 +4,49 @@ import Hooks from '../../../utilities/hooks';
...
+    afterEach(function () {
+        if (!loginComponent.logoutButton.isDisplayed()) {
+            loginComponent.openLoginComponent();
+        }
+        loginComponent.logOut();
     })
 
 })
+
+describe('API documentation file', () => {
```
andlar: I think the things that happen in a beforeEach block inside a describe block ought to be noted in the description of the describe block. So this could be something like `when uploading API documentation files as ADMIN`, whereas the previous would could be `when uploading Listings as ONC-ACB`

e2e/pages/administration/upload/upload.spec.js
```diff
@@ -4,24 +4,49 @@ import Hooks from '../../../utilities/hooks';
...
+        assert.include(uploadPage.listingUploadText.getText(),'was uploaded successfully. 1 pending products are ready for confirmation.', 'File has uploaded successfully');
     })
 
     it('allows uploading v18 template', () => {
         uploadPage.uploadListing('../../../resources/2015_v18_AQA2.csv');
-        assert.include(uploadPage.uploadSuccessfulText.getText(),'was uploaded successfully. 1 pending products are ready for confirmation.', 'File has uploaded successfully');
+        assert.include(uploadPage.listingUploadText.getText(),'was uploaded successfully. 1 pending products are ready for confirmation.', 'File has uploaded successfully');
+    })
+
+    afterEach(function () {
```
andlar: Can we move the `afterEach` block to be immediately after the `beforeEach` block? I know it happens after all of the tests, but having the "set up" and "tear down" steps in a sort of header area makes it clear that things _will_ happen after the tests run. It'll be especially useful as we add more tests; we won't need to scroll all the way down to the bottom of a random block to figure out if something's happening. Also, I think it'll help make it clear that, at least for this block, we're logging in and logging out for each test. By putting the "afterEach" block at the end that's harder to tell

e2e/pages/administration/upload/upload.spec.js
```diff
@@ -4,24 +4,49 @@ import Hooks from '../../../utilities/hooks';
...
+        loginComponent.logOut();
     })
 
 })
+
+describe('API documentation file', () => {
+
+    beforeEach(function () {
+        loginComponent.loginAsAdmin();
+    })
```
andlar: If we're logging out in after each test above, why aren't we after each of the ones in this block?

andlar (changes requested):

e2e/pages/administration/upload/upload.spec.js
palakpatel311: Again one of those things when i like to see the test in the order it happens. so I like to write before , Tc execution and after block sequence writing. I think we should leave this to personal preference.

e2e/pages/administration/upload/upload.po.js
palakpatel311: Why component? If i remember correctly we said that components will be those which would be accessed from any page. I dont see upload page to have any components. 
Also, to Todd's point, having id for upload button makes sense. But regardless we only allow user to upload one upload at a time (either listing or surveillance or api doc), In this case upload button with btn-ai-success will be only once on page.

e2e/pages/administration/upload/upload.spec.js
palakpatel311: Uploading listing had 2 Tcs but upload API doc has only one TC so logging out after that would just add more execution time with no value.
Also, updated beforeEach to beforeAll in upload listing block.

e2e/pages/administration/upload/upload.spec.js
palakpatel311: fixed

e2e/pages/administration/upload/upload.po.js
palakpatel311: Updated as per our conversation

andlar: I like the new components, but I think this page object now has stuff in it that's part of the component stuff, and should be removed from the page

palakpatel311: I did not remove anything from upload.po as i know some tests are using method from that which are in different branch so need to clean up after wards.

andlar (approved):

tmy1313 (approved):

PR #793, opened 2020-08-25
Ocd 3393 - Add ability to validate and save Real World Testing data

tmy1313:


src/app/pages/listing/listing.scss
```diff
@@ -0,0 +1,3 @@
+.listing__rwt-container {
+    display: grid;
```
andlar: Doesn't this need more styling? Just "display grid" won't do anything without some other stuff, right?

src/app/pages/listing/history/history.component.js
```diff
@@ -253,6 +253,26 @@ export const ListingHistoryComponent = {
             if (prev.curesUpdate !== curr.curesUpdate) {
                 activity.change.push('2015 Edition Cures Update status changed from ' + (prev.curesUpdate ? 'True' : 'False') + ' to ' + (curr.curesUpdate ? 'True' : 'False'));
             }
+            if (prev.rwtPlanUrl !== curr.rwtPlanUrl) {
```
andlar: Was this a requirement? I don't remember if it is

src/app/pages/listing/listing.html
```diff
@@ -105,6 +105,29 @@ <h2 class="product-heading">ONC-Authorized Testing Laborator{{ $ctrl.listing.tes
...
+          <div ng-if="$ctrl.displayRealWorldTesting($ctrl.listing) && ($ctrl.listing.rwtPlanUrl || $ctrl.listing.rwtPlanSubmissionDate || $ctrl.listing.rwtResultsUrl || $ctrl.listing.rwtResultsSubmissionDate)">
+            <h2>Real World Testing</h2>
+            <div class="listing__rwt-container">
+              <div class="data-label" ng-if="$ctrl.listing.rwtPlanUrl">
+                <label for="rwtPlanUrl">Plan</label>
+              </div>
+              <div ng-if="$ctrl.listing.rwtPlanUrl">
+                <p><a ai-a href="{{ $ctrl.listing.rwtPlanUrl }}" name="rwtPlanUrl" id="rwt-plan-url">{{ $ctrl.listing.rwtPlanUrl }}</a></p>
+              </div>
+              <div ng-if="$ctrl.listing.rwtPlanSubmissionDate && ($ctrl.hasAnyRole(['ROLE_ADMIN', 'ROLE_ONC']) || $ctrl.canManageAcb($ctrl.listing.certifyingBody) || $ctrl.canManageDeveloper($ctrl.listing.developer))">
```
andlar: `canManageAcb` also checks against ROLE_ADMIN/ROLE_ONC. It shouldn't need to be here too

src/app/services/login.service.js
```diff
@@ -33,6 +34,19 @@
                  || (userRole === 'ROLE_ONC' && targetRole !== 'ROLE_ADMIN' && targetRole !== 'ROLE_ONC'));
         }
 
+        function canManageAcb (acb) {
```
andlar, line 37: There should be a test for this

src/app/services/util.service.js
```diff
@@ -149,6 +150,13 @@
             return '';
         }
 
+        function displayRealWorldTesting (listing) {
```
andlar: This would be good to have in the unit tests too

kekey1 (dismissed):

src/app/pages/listing/history/history.component.js
tmy1313: image

src/app/pages/listing/listing.scss
tmy1313: By default a 1 column grid is created if the column layout is not specified.
image

andlar: Is there a Listing ID of some listing I can find on DEV where I can examine this? If the default styling of the grid basically mimics what you'd see if the style wasn't there then I'm not sure what the point would be, but don't know where to test this

src/app/pages/listing/history/history.component.js
andlar: Suggest double-checking against the "confirm date" thing in the requirements; if they're not displayed to the end users then activity shouldn't be in the eye, but otherwise fine with this change

src/app/services/login.service.spec.js
```diff
@@ -189,6 +189,70 @@
...
 
+        describe('when current user is managing acbs', () => {
+            beforeEach(() => {
+                let user = angular.copy(mock.user);
+                user.Authority = undefined;
+                $localStorage.currentUser = user;
+            });
+            it('should know ADMIN can manage any acb', () => {
+                $localStorage.currentUser.Authority = 'ROLE_ADMIN';
+                expect(auth.canManageAcb({id: 5}));
```
andlar: Doesn't this need at `.toBe(true)` or something like it? Same question for all of the following ones

src/app/services/login.service.spec.js
```diff
@@ -189,6 +189,70 @@
...
+            describe('when the user is role ACB', () => {
+                it('should allow user to manage acb', () => {
+                    $localStorage.currentUser.Authority = 'ROLE_ACB';
+                    $localStorage.currentUser.orgs = [{id: 2}, {id: 3}];
+                    expect(auth.canManageAcb({id: 3}));
+                });
+                it('should not allow user to manage acb', () => {
+                    $localStorage.currentUser.Authority = 'ROLE_ACB';
+                    $localStorage.currentUser.orgs = [{id: 2}, {id: 3}];
+                    expect(auth.canManageAcb({id: 4}));
```
andlar: image
What's the difference here?

src/app/services/login.service.spec.js
```diff
@@ -189,6 +189,70 @@
...
+            });
+            it('should know ADMIN can manage any developer', () => {
+                $localStorage.currentUser.Authority = 'ROLE_ADMIN';
+                expect(auth.canManageDeveloper({id: 5}));
+            });
+            it('should know ONC can manage any developer', () => {
+                $localStorage.currentUser.Authority = 'ROLE_ONC';
+                expect(auth.canManageDeveloper({id: 5}));
+            });
+            it('should know ACB cannot manage any developer', () => {
```
andlar: Pretty sure all ACBs should be able to manage all Developers

andlar (changes requested):

src/app/pages/listing/history/history.component.js
tmy1313: We discussed this last week, and Jim did not feel this was a problem.

src/app/pages/listing/listing.scss
tmy1313: https://chpl.ahrqdev.org/#/listing/10414

andlar: Taking off the "display: grid" value makes almost no difference; it brings the next section a few pixels up on the screen. We already have a ton of unused/wasted CSS. While I have no problem leaving the class in the HTML for XPATH/CSS/future-proofing reasons, I'd rather remove the class entirely from the css, as the styling doesn't make any significant difference

src/app/services/login.service.spec.js
andlar: Test description doesn't match execution

Also, can you add a linebreak between tests? Stacking them all together is not consistent with our other test files

src/app/services/util.service.js
andlar: Not sure if this util function is needed. It's only in one place, and could be replaced with `ng-if="$ctrl.listing.rwtEligibilityYear...` which would be clear, not add to our `util` service (which is already a service I wish we didn't have) and wouldn't require adding tests to the `util.service.spec.js` file

kekey1 (approved):

kekey1 (changes requested):
Needs conflict resolved in yarn file; otherwise I don't have any comments to add.

src/app/components/listing/edit.component.js
```diff
@@ -214,6 +220,16 @@ export const ListingEditComponent = {
                     this.idFields.suffix;
             }
             this.listing.certificationDate = this.listing.certDate.getTime();
+            if (this.listing.rwtPlanSubmissionDateObject) {
+                this.listing.rwtPlanSubmissionDate = this.listing.rwtPlanSubmissionDateObject.getTime();
```
andlar: These dates are LocalDate objects in Java, now, right? So they won't be of type `long` in the API, which means this init & save stuff is using the wrong types of the dates

andlar (changes requested):

src/app/admin/components/certifiedProduct/management.directive.js
```diff
@@ -36,6 +36,8 @@
         vm.isDeveloperBanned = isDeveloperBanned;
         vm.loadCp = loadCp;
         vm.loadSurveillance = loadSurveillance;
+        //vm.localDateToString = DateUtil.localDateToString;
```
kekey1: Do we need the commented out line?

kekey1 (changes requested):

src/app/admin/components/certifiedProduct/listing/edit.controller.js
```diff
@@ -81,6 +82,12 @@
             } else {
                 vm.cp.meaningfulUseUserHistory = [];
             }
+            if (vm.cp.rwtPlansCheckDate) {
+                vm.cp.rwtPlansCheckDateObject = vm.DateUtil.localDateToTimestamp(vm.cp.rwtPlansCheckDate);
+            }
+            if (vm.cp.rwtResultsCheckDate) {
+                vm.cp.rwtResultsCheckDateObject = new Date(vm.cp.rwtResultsCheckDate);
```
andlar: Why is the `Plans` one different from the `Results` one? They're the same object type coming from the API, right?

src/app/admin/components/certifiedProduct/listing/edit.html
```diff
@@ -151,6 +151,41 @@ <h4 class="modal-title" id="edit-listing-label">Edit Certified Product</h4>
       <label for="cp-other-acb">Other ACB</label>
       <input class="input-sm form-control" type="text" ng-model="vm.cp.otherAcb" id="cp-other-acb">
     </div>
+    <div ng-if="vm.cp.rwtEligibilityYear">
+      Real World Testing
+      <div class="form-horizontal">
+        <div class="form-group form-group-sm">
+            <label class="col-sm-6 control-label" for="rwt-plans-url">Plans URL</label>
+            <div class="col-sm-6">
+              <input class="input-sm form-control" type="url" ng-model="vm.cp.rwtPlansUrl" id="rwt-plans-url" name="rwtPlansUrl" ng-maxlength="1024">
```
andlar, line 160: This input needs a URL validation to reduce the chance of an improperly formatted URL getting into the system

src/app/admin/components/certifiedProduct/listing/edit.html
```diff
@@ -151,6 +151,41 @@ <h4 class="modal-title" id="edit-listing-label">Edit Certified Product</h4>
...
+            </div>
+            <label class="col-sm-6 control-label" for="rwt-plans-check-date">Plans Last ONC-ACB Completeness Check</label>
```
andlar, line 164:
```suggestion
            <label class="col-sm-6 control-label" for="rwt-plans-check-date">Plans' Last ONC-ACB Completeness Check</label>
```

Is this possessive? Should there be an apostrophe here?

src/app/pages/listing/history/history.component.js
```diff
@@ -253,6 +253,26 @@ export const ListingHistoryComponent = {
             if (prev.curesUpdate !== curr.curesUpdate) {
                 activity.change.push('2015 Edition Cures Update status changed from ' + (prev.curesUpdate ? 'True' : 'False') + ' to ' + (curr.curesUpdate ? 'True' : 'False'));
             }
+            if (prev.rwtPlansUrl !== curr.rwtPlansUrl) {
+                if (!prev.rwtPlansUrl) {
+                    activity.change.push('Real World Testing Plans URL added: ' + curr.rwtPlansUrl);
+                } else if (!curr.rwtPlansUrl) {
+                    activity.change.push('Real World Testing Plans URL removed: ' + prev.rwtPlansUrl);
+
```
andlar: Extra blank line here isn't needed

andlar (changes requested):

andlar (approved):
Will leave the comment about text changes around, but let that go to the PO. I'm good with the functionality here

kekey1 (dismissed):

src/app/pages/listing/history/history.component.js
```diff
@@ -253,6 +253,24 @@ export const ListingHistoryComponent = {
             if (prev.curesUpdate !== curr.curesUpdate) {
                 activity.change.push('2015 Edition Cures Update status changed from ' + (prev.curesUpdate ? 'True' : 'False') + ' to ' + (curr.curesUpdate ? 'True' : 'False'));
             }
+            if (prev.rwtPlansUrl !== curr.rwtPlansUrl) {
+                if (!prev.rwtPlansUrl) {
+                    activity.change.push('Real World Testing Plans URL added: ' + curr.rwtPlansUrl);
+                } else if (!curr.rwtPlansUrl) {
+                    activity.change.push('Real World Testing Plans URL removed: ' + prev.rwtPlansUrl);
+                } else {
+                    activity.change.push('Real World Testing Plans URL changed from ' + prev.rwtPlansUrl + ' to ' + curr.rwtPlansUrl)
```
andlar: Missing a semi-colon, as the linter calls out

src/app/pages/listing/history/history.component.js
```diff
@@ -253,6 +253,24 @@ export const ListingHistoryComponent = {
...
+                }
+            }
+            if (prev.rwtResultsUrl !== curr.rwtResultsUrl) {
+                if (!prev.rwtResultsUrl) {
+                    activity.change.push('Real World Testing Results URL added: ' + curr.rwtResultsUrl);
+                } else if (!curr.rwtResultsUrl) {
+                    activity.change.push('Real World Testing Results URL removed: ' + prev.rwtResultsUrl);
+                } else {
+                    activity.change.push('Real World Testing Results URL changed from ' + prev.rwtResultsUrl + ' to ' + curr.rwtResultsUrl)
```
andlar: Also missing a semi-colon

andlar (changes requested):

src/app/pages/listing/listing.component.js
```diff
@@ -2,20 +2,23 @@ export const ListingComponent = {
...
             this.$uibModal = $uibModal;
+            this.DateUtil = DateUtil;
             this.authService = authService;
             this.isOn = featureFlags.isOn;
             this.networkService = networkService;
             this.utilService = utilService;
             this.certificationStatus = utilService.certificationStatus;
             this.hasAnyRole = authService.hasAnyRole;
+            this.canManageAcb = authService.canManageAcb;
+            this.canManageDeveloper = authService.canManageDeveloper;
```
andlar: Don't think these are used any more

src/app/pages/listing/listing.html
```diff
@@ -105,6 +105,28 @@ <h2 class="product-heading">ONC-Authorized Testing Laborator{{ $ctrl.listing.tes
             <h2 class="product-heading">Relied Upon Software</h2>
             <p>{{ $ctrl.listing.productAdditionalSoftware }}</p>
           </div>
+          <div ng-if="$ctrl.listing.rwtEligibilityYear && ($ctrl.listing.rwtPlansUrl || $ctrl.listing.rwtPlansCheckDate || $ctrl.listing.rwtResultsUrl || $ctrl.listing.rwtResultsCheckDate)">
+            <h2>Real World Testing</h2>
```
andlar: Why doesn't this have the same class as the other `<h2>` items in this file?

andlar (approved):

kekey1 (approved):

PR #803, opened 2020-09-25
OCD 3487: AQA test for CMS ID

palakpatel311:


e2e/components/cms-widget/cms-widget.po.js
```diff
@@ -0,0 +1,95 @@
...
+    cmsIdReverseLookup: '=CMS ID Reverse Lookup',
+    progressBar: '.progress-bar',
+    certId: '#get-ehr-cert-id',
+    removeProducts: '#cms-remove-all',
+    baseCriteria: '=Base Criteria',
+    compareProducts: '#cms-compare',
+    cmsCertificationId: 'strong.ng-binding',
+    compareWidgetDropdown: '#compare-widget-dropdown',
+    missingBaseCriteriaListOr: '.cms-widget__missing-or',
+    missingBaseCriteriaListAnd: '.cms-widget__missing-and',
```
kekey1, line 13: A locator was also added for "cms-widget__missing-additional" - do we need to test that one too?

e2e/components/cms-widget/cms-widget.readonly.spec.js
```diff
@@ -0,0 +1,181 @@
...
+            assert.isFalse(cms.progressBar.isDisplayed());
+        });
+
+        it('should not have base criteria link', () => {
+            assert.isFalse(cms.baseCriteriaLink.isDisplayed());
+        });
+
+        it('should not have missing base criteria list', () => {
+            assert.isFalse(cms.missingBaseCriteriaListOr.isDisplayed());
+            assert.isFalse(cms.missingBaseCriteriaListAnd.isDisplayed());
```
kekey1, line 54: Also this one? "cms-widget__missing-additional"

wdio.conf.js
```diff
@@ -208,6 +225,33 @@ exports.config = {
         this.scrollIntoView();
         browser.execute(runInBrowser,this);
         }, true)
+
+        browser.addCommand("waitForFileExists", function (filePath, timeout) {
+            return new Promise(function (resolve, reject) {
+            var timer = setTimeout(function () {
+                watcher.close();
+                reject(new Error('File did not exists and was not created during the timeout.'));
```
kekey1:
```suggestion
                reject(new Error('File did not exist and was not created before the timeout.'));
```

kekey1 (changes requested):

e2e/components/cms-widget/cms-widget.po.js
palakpatel311: How do we get this? I tried couple of listing even with 0% but I did not see this.

kekey1: Turns out those "combo" messages only appear if you try to create a CMS ID with a combination of 2014 and 2015 listings. Since there is no longer a "+ CMSID" button next to any of the 2014 listings I guess there is nothing to test!

kekey1 (dismissed):

e2e/components/cms-widget/cms-widget.po.js
```diff
@@ -0,0 +1,95 @@
...
+    get removeProductsButton () {
+        return $(elements.removeProducts);
+    }
+
+    get baseCriteriaLink () {
+        return $(elements.baseCriteria);
+    }
+
+    get noProductsSelectedText () {
+        return $(asserts.noProductsSelected);
```
andlar: What is the difference between `asserts` and `elements`?

e2e/components/cms-widget/cms-widget.readonly.spec.js
```diff
@@ -0,0 +1,181 @@
+import CmsWidgetComponent from './cms-widget.po';
+import SearchPage from '../../pages/search/search.po';
+import Hooks from '../../utilities/hooks';
+
+const config = require('../../config/mainConfig');
+const path = require('path');
+const fs = require('fs');
+
+let cms, hooks, search;
+let ListingId1 = 9851;
```
andlar: We use `camelCase` on variables, not `PascalCase`. The only place with `PascalCase` is for Classes

e2e/components/cms-widget/cms-widget.readonly.spec.js
```diff
@@ -0,0 +1,181 @@
...
+let ListingId2 = 9879;
+let search1 = '2621';//using developer code to search listing
```
andlar, line 12: Is this developer code or developer database ID?

e2e/components/cms-widget/cms-widget.readonly.spec.js
```diff
@@ -0,0 +1,181 @@
...
+let search2 = '2155';//using developer code to search listing
+let cmsId;
+let chplPublicGuide = 'https://www.healthit.gov/sites/default/files/policy/chpl_public_user_guide.pdf';
```
andlar: This is an expected value, right? Then it should be `const` not `let`

e2e/components/cms-widget/cms-widget.readonly.spec.js
```diff
@@ -0,0 +1,181 @@
...
+            assert.isTrue(cms.cmsCertificationIdText.isDisplayed());
+            cmsId = cms.cmsCertificationIdText.getText();
+        });
+
+        it('PDF is downloaded after generating CMS ID', () => {
+            cms.downloadPDFButton.click();
+            const fileName = cmsId + '.pdf';
+            const filePath = path.join(global.downloadDir, fileName);
+            browser.waitForFileExists(filePath,config.timeout);
+            //Assertion to check if file downloaded ?
```
andlar: What's this comment for?

wdio.conf.js
```diff
@@ -284,8 +328,23 @@ exports.config = {
      * @param {Array.<Object>} capabilities list of capabilities details
      * @param {<Object>} results object containing test results
      */
-    // onComplete: function(exitCode, config, capabilities, results) {
-    // },
+     onComplete: function() {
```
andlar: Why is this function here? It looks like you're trying to remove things after the tests run. I'd lean towards having those things removed as part of the `yarn e2e:clean` command, because removing them by default makes it hard to debug stuff. If we _do_ keep it, the indentation is all off

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    cms = new CmsWidgetComponent();
+    search = new SearchPage();
+    cmsLookup = new CmsLookupPage();
+    hooks = new Hooks();
+});
+
+describe('On cms reverse look up page', () => {
+    describe('When searching for a CMS ID which was generated before', () => {
+        beforeAll( () => {
+            //creating cms id first as they are different cms ids between different environments
```
andlar: Shouldn't create CMS IDs separately. Once a CMS ID has been generated in PROD, it will get to all of our environments, so we should use those as things we can execute tests against.

Even trying to generate a new CMS ID from scratch, if it uses a set of Listings that have already generated a CMS ID, should generate the same one as the first time

e2e/components/cms-widget/cms-widget.po.js
```diff
@@ -0,0 +1,95 @@
...
+
+    get missingBaseCriteriaListOr () {
+        return $(elements.missingBaseCriteriaListOr);
+    }
+
+    get missingBaseCriteriaListAnd () {
+        return $(elements.missingBaseCriteriaListAnd);
+    }
+
+    get downloadPDFButton () {
```
andlar: PDF isn't an acronym, or if it is, it's not used that way
```suggestion
    get downloadPdfButton () {
```

andlar (changes requested):

e2e/components/cms-widget/cms-widget.po.js
palakpatel311: I am using asserts to check assertions. such as this text is only used in assertions and which means it is not element on which you could perform any action.
Actually moved to elements after rethinking

e2e/components/cms-widget/cms-widget.readonly.spec.js
palakpatel311: Developer code

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
palakpatel311: Agree. But CMSID for listing1 and listing2 is different in dev , qa and stage. so i cant hard code cmsId in cmsId reverse look up.

How can I find already generated CmsId from Prod?

wdio.conf.js
palakpatel311: Moved this to as part of yarn e2e:clean

e2e/components/cms-widget/cms-widget.po.js
palakpatel311: Updated

e2e/components/cms-widget/cms-widget.readonly.spec.js
palakpatel311: Removed

e2e/components/cms-widget/cms-widget.readonly.spec.js
palakpatel311: Changed

e2e/components/cms-widget/cms-widget.readonly.spec.js
palakpatel311: Changed

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
palakpatel311: Used cmsid created from prod

andlar (dismissed):

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -0,0 +1,50 @@
+import CmsLookupPage from './cms-lookup.po';
+import Hooks from '../../../utilities/hooks';
+
+const config = require('../../../config/mainConfig');
+const path = require('path');
+const fs = require('fs');
+
+let cmsLookup, hooks;
+var listingIds = ['15.07.07.1447.BE02.01.00.1.160815','15.07.07.1447.BE01.02.01.1.161014','15.07.07.1447.EP03.02.03.1.161209','15.07.07.1447.EP03.03.04.1.170403'];
```
kekey1: Is it possible to use the database IDs here? There are a number of actions that users can do that cause the CHPL Product Numbers to change which would break this test.

kekey1 (changes requested):

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
palakpatel311: No. Assertion here is to check that these IDs shows up on the screen when finding cms id and as well as in pdf. so if i switch to database id, I cant test anything.
But i did add these Ids in aqa wiki so no one should be using it

kekey1: Understood. I think it's possible a user in prod would make the change and that would migrate down into our test data, but I can live with it!

kekey1 (dismissed):

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
palakpatel311: Agree. But we have to see if these listings will change frequently. if it does, we can change listing ids later

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -16,35 +16,65 @@ beforeAll(async () => {
...
+                cmsLookup.searchField.clearValue();
+                cmsLookup.searchField.addValue(inputs[i].cmsId);
+                cmsLookup.searchIdButton.click();
+                hooks.waitForSpinnerToDisappear();
+            });
+
+            it('should show correct listings for the CMS ID', () => {
+                browser.waitUntil( () => $(cmsLookup.lookupResultsTable).isDisplayed());
+                var ls = [];
+                var length = $$('//*[@id="lookupCertIdResults"]/tbody/tr').length;
```
andlar: These element locators ought to be in the page object, right?

e2e/pages/resources/cms-lookup/cms-lookup-dp.js
```diff
@@ -0,0 +1,17 @@
+module.exports = [
+    {
+        testName: 'only 2015 listings',
+        cmsId: '0015EYE3ZT3QFP4',
+        listingIds: ['15.07.07.1447.BE02.01.00.1.160815','15.07.07.1447.BE01.02.01.1.161014','15.07.07.1447.EP03.02.03.1.161209','15.07.07.1447.EP03.03.04.1.170403'],
```
andlar: These aren't listing IDs, they're chpl product numbers. IDs are database values

```suggestion
        chplProductNumbers: ['15.07.07.1447.BE02.01.00.1.160815','15.07.07.1447.BE01.02.01.1.161014','15.07.07.1447.EP03.02.03.1.161209','15.07.07.1447.EP03.03.04.1.170403'],
```

andlar (changes requested):

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
palakpatel311: I think this locator was very specific to this test thats why I dint bother to move that to PO.

andlar: The locator that is the "length of the rows" is specific to the test, but the locator that is the "rows of the table" would not be, and could be used in other tests to verify that certain listings appeared or didn't appear in the lookup table

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -29,11 +29,11 @@ describe('On cms reverse look up page', () => {
             it('should show correct listings for the CMS ID', () => {
                 browser.waitUntil( () => $(cmsLookup.lookupResultsTable).isDisplayed());
                 var ls = [];
-                var length = $$('//*[@id="lookupCertIdResults"]/tbody/tr').length;
+                var length = cmsLookup.rowsLookupResultsTable.length;
                 for ( var j = 1; j <= length; j++ ) {
                     ls.push($('//*[@id="lookupCertIdResults"]/tbody/tr[' + j + ']/td[6]').getText());
```
andlar: This should be a locator too. I've done things in the Product edit where I pass, to the Page Object file, an element, and have that file return some sub-element from that element. Something like 
```suggestion
                    ls.push(page.getChplProductNumber(cmsLookup.rowsLookupResultsTable(j)).getText());
```

palakpatel311: Not sure if it will work in this case as its table, but i did what i think will be fine here.

andlar (dismissed):

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -0,0 +1,80 @@
...
+            });
+
+            it('should have download results button and download file should contain correct listings Ids', () => {
+                cmsLookup.downloadResultsButton.scrollAndClick();
+                const fileName = 'CMS_ID.' + inputs[i].cmsId + '.csv';
+                const filePath = path.join(global.downloadDir, fileName);
+                browser.waitForFileExists(filePath,config.timeout);
+                assert.isTrue(fs.existsSync(filePath));
+                const fileContents = fs.readFileSync(filePath, 'utf-8');
+                var isInclude = false;
```
kekey1: I think if _any_ of the chplProductNumbers are found this will be "true" but do we want to check that _all_ of them are found? Needs the logic reorganized a bit.

kekey1 (changes requested):

e2e/pages/resources/cms-lookup/cms-lookup.po.js
```diff
@@ -37,7 +37,7 @@ class CmsLookupPage {
     chplProductNumberFromTable (rowNumber) {
         return $('//*[@id="lookupCertIdResults"]/tbody/tr[' + rowNumber + ']/td[6]');
     }
-    
+
```
andlar: Why is this added?

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -16,7 +16,7 @@ beforeAll(async () => {
 });
 
 describe('On cms reverse look up page', () => {
-    for (var i in inputs) {
+    for (const i in inputs) {
```
andlar: This looks weird. `i` is usually the indexing variable, which _must_ change. But `const` variables don't change. What's going on here?

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -43,13 +43,13 @@ describe('On cms reverse look up page', () => {
...
-                var isInclude = false;
+                var isInclude = 0;
                 for ( var k = 0; k < inputs[i].chplProductNumbers.length; k ++) {
                     if (fileContents.includes(inputs[i].chplProductNumbers[k])) {
-                        isInclude = true;
+                        isInclude ++;
                     }
                 }
-                assert.isTrue(isInclude);
+                assert.equal(isInclude,inputs[i].chplProductNumbers.length, 'All chpl product numbers arent present in the download file');
```
andlar: `arent` isn't a word. It should be `aren't`, as these error messages ought to be grammatically correct

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
```diff
@@ -43,13 +43,13 @@ describe('On cms reverse look up page', () => {
                 browser.waitForFileExists(filePath,config.timeout);
                 assert.isTrue(fs.existsSync(filePath));
                 const fileContents = fs.readFileSync(filePath, 'utf-8');
-                var isInclude = false;
+                var isInclude = 0;
```
andlar: `isInclude` looks like a boolean value. It shouldn't be used to hold a count of anything

andlar (changes requested):

e2e/pages/resources/cms-lookup/cms-lookup.po.js
palakpatel311: This is the way i am getting chplproduct number from table in the spec file

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
palakpatel311: so beforeall in the describe block seems to be changing i values and running it for only third values from array(I think its running in async mode but not positive) so i have to keep i as const through out one describe block.
I value will still get increased after iteration.

e2e/pages/resources/cms-lookup/cms-lookup.po.js
andlar: The line I was asking about was the added blank line. Not the locator

e2e/pages/resources/cms-lookup/cms-lookup.readonly.spec.js
andlar: What about?
```suggestion
    inputs.forEach(input => {
```

Then, anywhere below you're using `inputs[i]` you can replace it with just `input`?

palakpatel311: Tried. It sys 0 test and doesnt run anything. Okay worked

andlar (approved):

kekey1 (approved):

PR #807, opened 2020-10-01
OCD-3508- Edit SVAP information for specific criteria in listing details

tmy1313:
OCD-3508


src/app/components/listing/details/criteria/edit.component.js
```diff
@@ -74,6 +77,19 @@ export const CertificationCriteriaEditComponent = {
             this.close({$value: this.cert});
         }
 
+        svapOnChange (action) {
+            switch (action.action) {
+            case 'Remove':
+                this.cert.svaps = this.cert.svaps
+                    .filter(svap=> !(svap.svapId === action.item.item.svapId));
```
andlar:
```suggestion
                    .filter(svap => svap.svapId !== action.item.item.svapId);
```
The "test data" filter has the "not (this and this and this)" construct because there are multiple values to check. Test standard, which is similar in that it only has the standard, just does the id !== id check. Request the simpler check

src/app/components/listing/details/criteria/edit.component.spec.js
```diff
@@ -1,7 +1,7 @@
 (() => {
     'use strict';
 
-    describe('the Certification Criteria Modal controller', () => {
+    fdescribe('the Certification Criteria Modal controller', () => {
```
andlar:
```suggestion
    describe('the Certification Criteria Modal controller', () => {
```

src/app/components/listing/details/criteria/edit.component.spec.js
```diff
@@ -46,10 +46,24 @@
                         {'id': 2, 'name': '(a)(2)(ii)'},
                         {'id': 3, 'name': '(a)(1)(i)'},
                     ],
+                    //allowedSvaps: [
+                    //    {svapId: 1, regulatoryTextCitation: 'reg1', approvedStandardVersion: 'ver1'},
```
andlar: Please delete

src/app/components/listing/details/criteria/edit.component.spec.js
```diff
@@ -46,10 +46,24 @@
...
+                    //    {svapId: 2, regulatoryTextCitation: 'reg2', approvedStandardVersion: 'ver2'},
+                    //],
                     resources: mock.resources,
                     hasIcs: false,
                     isConfirming: false,
                 };
+                scope.resolve.cert = ({
```
andlar: Why not just put these values in the `mock.cert` object defined at the top of the file?

src/app/components/listing/details/criteria/edit.html
```diff
@@ -29,6 +29,35 @@ <h4 class="modal-title" id="edit-criteria-label">Certification Criteria: <span n
...
+                    items="$ctrl.cert.allowedSvaps"
+                    selected-item-keys="$ctrl.selectedSvapKeys"
+                    placeholder="Select an SVAP"
+                    on-change="$ctrl.svapOnChange(action)">
+                  </chpl-expandable-list>
+                </div>
+              </div>
+            </div>
+          </td>
+          <td ng-if="$ctrl.cert.testStandards && $ctrl.cert.testStandards.length > 0 && !$ctrl.canEdit()">
```
andlar: Why are you adding test standard stuff to this row?

src/app/components/listing/details/criteria/view.html
```diff
@@ -28,6 +28,19 @@
             </td>
             <td>{{ $ctrl.cert.gap ? 'True' : 'False' }}</td>
           </tr>
+          <tr ng-if="$ctrl.cert.success && $ctrl.cert.svaps !== null" id="criteria_{{ $ctrl.cert.criterion.number }}_details_row_Svap{{ $ctrl.cert.criterion.title.includes('Cures') ? '_cures' : ''}}">
+            <td>
+              Standards Version Advancement Process <button class="btn btn-link btn-xs pull-right" uib-tooltip="TBD" tooltip-trigger="'mouseenter focus click'"><span class="sr-only">TBD</span><i class="fa fa-info-circle fa-lg"></i></button>
```
andlar: Reminder for the TBD

src/app/components/listing/details/criteria/view.html
```diff
@@ -28,6 +28,19 @@
...
+            </td>
+            <td ng-if="$ctrl.cert.svaps && $ctrl.cert.svaps.length > 0">
+              <ul>
+                <li ng-repeat="m in $ctrl.cert.svaps">
+                  <span ng-if="m.approvedStandardVersion"><ai-ellipsis text="{{ m.regulatoryTextCitation + ' ' + m.approvedStandardVersion }}" max-length="100" word-boundaries="true"></ai-ellipsis></span>
```
andlar: Aren't you building the display value in the component? Why concatenate those values again here?

src/app/components/util/expandable-list.component.js
```diff
@@ -126,6 +126,7 @@ export const ExpandableListComponent = {
 
         _createSelectedItem (item) {
             var selectedItem = {'item': item};
+            selectedItem.item.name = selectedItem.item[this.itemText];
```
andlar, line 129: Curious why this needed to be added when none of the other expandable-list stuff changed, and it didn't look like the SVAP stuff needed anything special from this component

andlar (changes requested):

kekey1 (dismissed):
Nothing to add to Andrew's comments

src/app/components/listing/details/criteria/edit.component.spec.js
tmy1313: Honestly, I couldn't figure out where it goes.  That string is not formatted, and is difficult to follow.  I think there is a lot of data in there that is not used by the tests.

src/app/components/listing/details/criteria/view.html
tmy1313: I'm building the display value in the edit component, not the view component.  I need it in the edit component for the expandable list control.  Should i change this?

src/app/components/util/expandable-list.component.js
tmy1313: This was actually a bug in the expandable list component.  All other instances of the component use `item-text = "name"` .  The svap instance of the expandable list component, uses `item-text = "displayText"`.  So `selectedItem.item.name` always worked.  But when I used a different attribute name for the item name (displayText), there was nothing to display.  Can discuss further with you if you would like.

src/app/components/listing/details/criteria/edit.component.spec.js
andlar: The `scope.resolve` object is created in line 42. In that object, the cert is a copy of the `mock.cert` object, which is created on line 8. If you add the `allowedSvaps` and `svaps` from lines 55-60 to just after line 20 it'll put these things there. What string are you talking about that's "not formatted"? The `mock.resources` has a lot of stuff, but you shouldn't need to touch that

src/app/components/listing/details/criteria/view.html
andlar: No, this is fine. Missed that this is in the view component

andlar (dismissed):

kekey1 (dismissed):

src/app/pages/reports/listings/listings.component.js
```diff
@@ -206,6 +206,15 @@ export const ReportsListingsComponent = {
                 if (testFunctionality.length > 0) {
                     obj.changes.push('<li>Test Functionality changes<ul>' + testFunctionality.join('') + '</li>');
                 }
+                var svapKeys = [{key: 'regulatoryTextCitation', display: 'SVAP'}];
+                var svap = this.ReportService.compareArray(prev[i].svaps, curr[i].svaps, svapKeys, 'regulatoryTextCitation');
+                if (svap.length > 0) {
+                    obj.changes.push('<li>SVAP changes<ul>');
```
andlar, line 212: This doesn't have the same structure as the testToolsUsed report stuff. Does this report changes in multiple SVAP at the same time? SVAP and Test Tools have the same basic structure, right? If so, why aren't the tests against changes the same?

andlar (dismissed):

kekey1 (dismissed):

src/app/components/listing/details/criteria/view.scss
```diff
@@ -0,0 +1,11 @@
+.criteria__tooltip-button {
```
andlar: This looks like a candidate for an `atoms.scss` style

andlar (dismissed):

kekey1 (dismissed):

src/app/components/listing/details/criteria/edit.component.spec.js
```diff
@@ -159,6 +166,50 @@
...
+                    ctrl.cert.svaps = [];
+                    let action = {
+                        action: 'Add',
+                        item: {
+                            item: {
+                                svapId: 1,
+                                regulatoryTextCitation: 'reg1',
+                                approvedStandardVersion: 'ver1',
+                            }
+                        }
```
andlar: This line reports a linting error about commas. There are a few others that show up when running the tests. During that execution the linting errors are displayed for the *.spec.js files. Doesn't need to be fixed now, but does before it goes to staging

andlar (changes requested):

src/app/components/listing/details/criteria/edit.component.spec.js
tmy1313: I did not get an error about this particular file, but fixed another linting error.  Let m know if this is still an issue.

kekey1 (dismissed):

src/app/components/listing/details/criteria/view.html
```diff
@@ -32,6 +32,22 @@
             </td>
             <td>{{ $ctrl.cert.gap ? 'True' : 'False' }}</td>
           </tr>
+          <tr ng-if="$ctrl.cert.success && $ctrl.cert.svaps !== null" id="criteria_{{ $ctrl.cert.criterion.number }}_details_row_Svap{{ $ctrl.cert.criterion.title.includes('Cures') ? '_cures' : ''}}">
+            <td ng-if="$ctrl.cert.svaps && $ctrl.cert.svaps.length > 0">
```
andlar: Why is the `ng-if` here? With this `if` and the `ng-if` in the `<tr>` above, if the listing has an empty array of svaps we'd have a table row with no data inside it, which breaks the rules of HTML (I know, I know, "HTML rules", but still...)

If the SVAP data shouldn't be there at all when there isn't data, then the `ng-if` should be in the `<tr>` check. If it should show something like "SVAP: none", when there aren't any, then this first `<td>` of the row should keep the SVAP td always, and there should be two other `<td>` options; one that shows when there's data, one that shows "none" when there isn't, in a similar fashion to how the "Standard" below works

src/app/components/listing/details/criteria/edit.html
```diff
@@ -29,6 +29,33 @@ <h4 class="modal-title" id="edit-criteria-label">Certification Criteria: <span n
...
+                    items="$ctrl.cert.allowedSvaps"
+                    selected-item-keys="$ctrl.selectedSvapKeys"
+                    placeholder="Select an SVAP"
+                    on-change="$ctrl.svapOnChange(action)">
+                  </chpl-expandable-list>
+                </div>
+              </div>
+            </div>
+          </td>
+          <td ng-if="$ctrl.cert.svaps && $ctrl.cert.svaps.length > 0 && !$ctrl.canEdit()">
```
andlar, line 51: When does this `<td>` get displayed? The way I read it, if it's displayed there's no heading for the section

andlar (changes requested):

src/app/components/listing/details/criteria/edit.html
tmy1313: If for some reason the criterion were removed and the user is ROLE-ACB, they would only "see" the existing SVAPs for the criterion.  Let me know if this isn't correct.

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #808, opened 2020-10-08
OCD-3443: Add AQA for listings and ACB name on products bar for a developer

palakpatel311:
[#OCD-3443]

### DIFF
--- a/e2e/pages/organizations/developers/developers.readonly.spec.js
+++ b/e2e/pages/organizations/developers/developers.readonly.spec.js
@@ -40,7 +40,6 @@ describe('the Developers page', () => {
             beforeEach(() => {
                 product = page.getProduct(name);
                 product.scrollIntoView({block: 'center', inline: 'center'});
-                browser.waitUntil(() => page.getVersionCount(product).isDisplayed());
                 page.selectProduct(product);
                 page.getProductInfo(product).waitForDisplayed({timeout: 55000});
             });
@@ -70,7 +69,6 @@ describe('the Developers page', () => {
             page.getDeveloperPageTitle(developer).waitForDisplayed();
             product = page.getProduct(productName);
             product.scrollIntoView({block: 'center', inline: 'center'});
-            browser.waitUntil(() => page.getVersionCount(product).getText() === '3 Versions');
             page.selectProduct(product);
             page.getProductInfo(product).waitForDisplayed({timeout: 55000});
         });
--- a/e2e/pages/organizations/developers/developers.readonly.spec.js
+++ b/e2e/pages/organizations/developers/developers.readonly.spec.js
@@ -40,13 +40,21 @@ describe('the Developers page', () => {
             beforeEach(() => {
                 product = page.getProduct(name);
                 product.scrollIntoView({block: 'center', inline: 'center'});
-                browser.waitUntil(() => page.getVersionCount(product).getText() === '5 Versions');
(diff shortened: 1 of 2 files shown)


e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -76,3 +84,20 @@ describe('the Developers page', () => {
         });
     });
 });
+
+describe('When user is on the Developer page for a Developer that doesnt exist', () => {
+    beforeEach(async () => {
+        browser.setWindowSize(1600, 1024);
```
andlar: These first four lines don't do anything that are needed for the tests, right? So why have them?

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -76,3 +84,20 @@ describe('the Developers page', () => {
         });
     });
 });
+
+describe('When user is on the Developer page for a Developer that doesnt exist', () => {
```
andlar:
```suggestion
describe('when a user is on the Developer page for a Developer that doesn\'t exist', () => {
```

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -76,3 +84,20 @@ describe('the Developers page', () => {
...
+        browser.setWindowSize(1600, 1024);
+        browser.setWindowRect(0, 0, 1600, 1024);
+        contact = new ContactComponent();
+        page = new DevelopersPage();
+        hooks = new Hooks();
+        const dummyDeveloperId = '0000';
+        await hooks.open('#/organizations/developers/' + dummyDeveloperId);
+    });
+
+    it('it should go to Home page', () => {
```
andlar:
```suggestion
    it('should go to Home page', () => {
```

Don't need the "double it"

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -40,14 +40,22 @@ describe('the Developers page', () => {
...
 
             it('should have product Contact information', () => {
                 expect(contact.get(product)).toHaveTextContaining('Tamara Grassle');
             });
+
+            it('should have ACB name', () => {
+                expect(page.getAcbName(product).getText()).toBe('UL LLC');
+            });
+
+            it('should have ACB name', () => {
```
andlar: This is the same `it` block as the previous test

andlar (changes requested):

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -40,14 +40,22 @@ describe('the Developers page', () => {
             beforeEach(() => {
                 product = page.getProduct(name);
                 product.scrollIntoView({block: 'center', inline: 'center'});
-                browser.waitUntil(() => page.getVersionCount(product).getText() === '5 Versions');
+                browser.waitUntil(() => page.getVersionCount(product).isDisplayed());
```
andlar: Is this still needed? I think it was here originally because we had to wait for the "get products" and "get versions" calls to finish, but with those no longer happening I don't think we need to wait for anything. Once the product can be scrolled into view it should be fully complete

palakpatel311: Removed and its working

andlar (approved):

tmy1313 (approved):

PR #811, opened 2020-10-12
OCD-3375: Edit basic listing data in place

andlar:


src/app/services/util.service.js
```diff
@@ -141,10 +141,15 @@
 
         function certificationStatus (listing, options) {
             if (listing.certificationEvents && listing.certificationEvents.length > 0) {
-                if (options && options.editing) {
-                    return $filter('orderBy')(listing.certificationEvents.map(event => { event.eventDate = event.statusDateObject.getTime(); return event; }),'-eventDate')[0].status.name;
-                }
-                return $filter('orderBy')(listing.certificationEvents,'-eventDate')[0].status.name;
+                let events = listing.certificationEvents
```
kekey1, line 144: There is a "currentStatus" field in the details object that has this same logic running in the API. Worth it to have the logic in two places?

kekey1 (dismissed):

kekey1 (dismissed):

src/app/services/util.service.js
andlar: Assuming you're okay with approving this as written, I'll take a look at this and see if there's something I can do with it. I think it might be an artifact of "how to do editing of certification status" things, but will dig in. Suggest we don't close this conversation until we can dig a little deeper, but also okay with the code moving forward for testing

kekey1 (dismissed):

src/app/admin/components/certifiedProduct/management.html
```diff
@@ -296,12 +296,10 @@ <h3 ng-if="vm.inspectingCp">System developer information</h3>
...
                 <br />
-                <chpl-listing-details ng-if="vm.activeCP !== ''"
-                                      default-all="true"
-                                      edit-mode="false"
-                                      hide-direct-review="true"
-                                      listing="vm.activeCP">
-                </chpl-listing-details>
+                <chpl-listing-details-view listing="vm.activeCP"
+                                           hide-direct-review="true"
+                                           >
```
tmy1313, line 301: Should the `>` be on the previous line?

src/app/pages/listing/edit.html
```diff
@@ -0,0 +1,42 @@
...
+      <div class="panel-heading">
+        <h2 class="panel-title">Edit Listing Information</h2>
+      </div>
+      <div class="panel-body">
+        <chpl-listing-edit
+          listing="$ctrl.listingBasic"
+          on-change="$ctrl.updateBasic(listing, messages, reason)"
+          resources="$ctrl.resources"
+          show-form-errors="$ctrl.showFormErrors"
+          >
```
tmy1313, line 13: Should this be on the above line?

tmy1313 (dismissed):

kekey1 (dismissed):
There are a couple of comments from last time this was PR-ed. One from Todd and one from me. I don't feel that either are enough to hold up my approval but would like to know if they are/should be resolved at some point.

tmy1313 (dismissed):

src/app/services/util.service.js
andlar: Figured it out. During editing, the "currentStatus" field isn't updated; that's only updated after a save. However, if the "current Status" is going to become one of a few specific ones we need to display a message with things like "setting this listing to this status may cause <insert thing here>". I think I can remove the use of this function when not editing; it's only used in a half dozen places at that time

kekey1 (approved):

tmy1313 (approved):

PR #816, opened 2020-10-15
OCD-3434: Create Upload for Real World Testing

tmy1313:


src/app/pages/listing/listing.html
```diff
@@ -124,6 +124,21 @@ <h2 class="product-heading">Real World Testing</h2>
...
               <p>Last ONC-ACB Completeness Check: {{ $ctrl.DateUtil.localDateToString($ctrl.listing.rwtResultsCheckDate) }}</p>
             </div>
+            <div ng-if="$ctrl.listing.rwtPlansUrl">
+              <p><a ai-a href="{{ $ctrl.listing.rwtPlansUrl }}" name="rwtPlansUrl" id="rwt-plans-url">{{ $ctrl.listing.rwtPlansUrl }}</a></p>
+            </div>
+            <div ng-if="$ctrl.listing.rwtPlansCheckDate && $ctrl.canViewRwtDates()">
+              <p>Last ONC-ACB Completeness Check: {{ $ctrl.DateUtil.localDateToString($ctrl.listing.rwtPlansCheckDate) }}</p>
+            </div>
+            <div class="data-label" ng-if="$ctrl.listing.rwtResultsUrl">
+              <label for="rwtRsultsURL">Results</label>
```
andlar: Two things:
1) I think the change to this file shouldn't be in this PR
2) If it is, the "for" value should be an id, which should be in the dash-case format, not the camelCase one, and "results" is misspelled

andlar (changes requested):
I was trying really hard to find something to ask to be changed, and was sad. But then a file appeared that probably shouldn't be here at all! Oh joyous day!

Anyway, the only requested change is that I think there's some extra stuff here that shouldn't be. The functionality as per the Feature looks great!

andlar (approved):

kekey1 (approved):

PR #821, opened 2020-10-28
OCD 3484: AQA test for search listing

palakpatel311:


e2e/pages/search/search-home.readonly.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+});
+
+describe('On search home page- "API Info for 2015 Ed. Products" button', () => {
+
+    it('should be displayed', () => {
+        assert.isTrue(page.homeSearchPageButtons('API Info for 2015 Ed. Products').isDisplayed());
+    });
+    it('should be clickable', () => {
+        assert.isTrue(page.homeSearchPageButtons('API Info for 2015 Ed. Products').isDisplayed());
```
andlar: This doesn't assert anything about clickability

e2e/pages/search/search-home.readonly.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    });
+    it('should have correct link to API documentation page', () => {
+        assert.equal(page.homeSearchPageButtons('API Info for 2015 Ed. Products').getAttribute('href'),browser.options.baseUrl + apiInfoLink);
+    });
+});
+
+describe('On search home page- "SED Info for 2015 Ed. Products" button', () => {
+
+    it('should be displayed', () => {
+        assert.equal(page.homeSearchPageButtons('SED Info for 2015 Ed. Products').isDisplayed(),true);
```
andlar: Why are these `assert.equal` but the ones for API documentation are `assert.isTrue`? I'd prefer the second type, because checking against `= true` is literally what the assert says, but in either case they ought to be the same

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,333 @@
+import SearchPage from './search.po.js';
+import Hooks from '../../utilities/hooks';
+
+let hooks, page;
+var countBefore;
```
andlar: Generally speaking we shouldn't have global variables. These ought to be defined and used in the scope where they're needed

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,333 @@
...
+
+    describe('When browsing all listings', () => {
+        beforeEach(() => {
+            page.browseAllButton.click();
+            page.waitForUpdatedListingResultsCount();
+        });
+
+        it('should show all listing (more than 500)', () => {
+            var count = page.listingTotalCount();
+            assert.isAbove(count,500);
```
andlar: Why is the number 500 being used? It ought to be saved as a const as well, with a meaningful name, to avoid the use of magic numbers

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,333 @@
...
+            page.waitForUpdatedListingResultsCount();
+        });
+        afterEach(() => {
+            page.clearFilters.click();
+        });
+
+        it('should show filtered listing results', () => {
+            var count = page.listingTableFirstPageRowCount();
+            for (var i = 1; i <= count; i++) {
+                assert.include(page.getColumnText(i,3),productName);
```
andlar: The `3` here ought to be a const as well. Something like `const PRODUCT_COL_IDX = 3;`

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,333 @@
...
+            page.statusRetiredFilterOption.click();
+            page.waitForUpdatedListingResultsCount();
+        });
+        afterEach(() => {
+            page.clearFilters.click();
+        });
+
+        it('should filter listing results', () => {
+            countAfter = page.listingTotalCount();
+            assert.isBelow(countAfter,countBefore);
```
andlar: Setting the "retired" status to true should mean there are _more_ listings, not fewer, right? So why is the assert asserting that something `isBelow`?

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,333 @@
...
+        });
+
+        it('should atleast show 1 retired listing', () => {
```
andlar:
```suggestion
        it('should at least show 1 retired listing', () => {
```

andlar (changes requested):

e2e/pages/search/search-home.readonly.spec.js
palakpatel311: Oops. I missed these.

e2e/pages/search/search.readonly.spec.js
palakpatel311: I used 500 hoping that listings will be more than that number always. But welcome for suggestion.
I will move that number to be constant.

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,360 @@
...
+const totalListing = 500;
+const PRODUCT_COL_IDX = 3;
+const DEVELOPER_COL_IDX = 2;
+const CHPLID_COL_IDX = 6;
+const STATUS_COL_IDX = 7;
+const EDITION_COL_IDX = 1;
+
+describe('On search page - ', () => {
+    var countBefore;
+    var countAfter;
```
andlar: It's better that these aren't global in scope, but they're still not deep enough. Most of the describe blocks don't use these variables; these variables should only exist in scopes where they're needed. If there are groups of the describe blocks below that all need these, then that group of describe blocks ought to be grouped in a describe block together

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,360 @@
...
+const config = require('../../config/mainConfig');
+const developerName = 'Allscripts';
+const productName = 'TouchWork';
+const acbId = '170008R01';
+const chplId = '15.99.04.3078.Ninj.01.00.0.200629';
+const version = '19';
+const totalListing = 500;
+const PRODUCT_COL_IDX = 3;
+const DEVELOPER_COL_IDX = 2;
+const CHPLID_COL_IDX = 6;
```
andlar: Suggest calling this "CHPL_PRODUCT_NUMBER_COL_IDX". The "CHPL ID" is the database ID, and though we sometimes refer to that product number as the id, it's not really correct

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,360 @@
...
+
+        it('should at least show 1 retired listing', () => {
+            let isInclude = false;
+            for (var i = 1; i <= page.listingTableFirstPageRowCount(); i++) {
+                if (page.getColumnText(i,STATUS_COL_IDX).includes('Retired')) {
+                    isInclude = true;
+                    break;
+                }
+            }
+            assert.equal(true, isInclude);
```
andlar: Any reason this isn't assert.isTrue?

andlar (dismissed):

andlar (approved):

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+    it('should show appropriate listing results', () => {
+        var count= page.listingTableRowCount();
+        assert.equal(count,1);
+    });
+
+});
+
+describe('On search page- When searching listings by ONC-ACB Id', () => {
```
ashwinimore: 'When searching listings by _CHPL ID_ or _CHPL product number_' instead of  ONC-ACB ID.
That's my only comment, not holding off on approval just for this, make edit when you can

ashwinimore (approved):

e2e/pages/search/search.readonly.spec.js
palakpatel311: This test is for searching listing by ONC-ACB Id, there is a separate test for searching listing by CHPL id.

e2e/pages/search/search.readonly.spec.js
```diff
@@ -0,0 +1,82 @@
...
+    it('should show appropriate listing results', () => {
+        var count= page.listingTableRowCount();
+        for (var i=1; i<=count;i++){
+            assert.include(page.getColumnText(i,3),productName);
+        }
+    });
+
+});
+
+describe('On search page- When searching listings by ONC-ACB Id', () => {
```
ashwinimore: ref: prev. comment- looking at acbID parameter, looks like this one is searching listings by ONC-ACB Id

palakpatel311: Looks like you are looking at outdated file. I see onc-acb id only once in latest file

ashwinimore: I reviewed commit by commit so that's possible(i did see updates for other things I came across but didn't one for this), it appears that it's already taken care of then

PR #826, opened 2020-11-06
OCD 3402: Automation test for updated text on download

palakpatel311:


andlar (dismissed):

athomas-git (dismissed):

e2e/pages/charts/product/product.readonly.spec.js
```diff
@@ -13,6 +13,8 @@ beforeAll(async () => {
 describe('on charts page - Unique product chart', () => {
 
     it('should only show 2015 edition products', () => {
+        //stale element error
+        browser.pause(5000);
```
andlar: Is there some way to do a "wait until" call? I'm not sure what we're waiting for, but are there any elements out there that could be looked for?

palakpatel311: Updated with waituntil

e2e/pages/charts/product/product.readonly.spec.js
```diff
@@ -13,6 +13,7 @@ beforeAll(async () => {
 describe('on charts page - Unique product chart', () => {
 
     it('should only show 2015 edition products', () => {
+        browser.waitUntil( () => page.chartTitle.isDisplayed() , 5000);
```
andlar: What's the `5000` for? Is that a timeout? This test, on QA, ran in <1 second last night, so if it's a timeout I don't think it's needed. If it is a timeout it shouldn't be a magic number. If we're going to have these kinds of things there should be a `const TIMEOUT = 5000;` defined somewhere, probably somewhere "global-ish" so all of our tests can use the same stuff, and standardize on things

Speaking of standardization, the spacing on this line doesn't match the line below, or the way most of our other anonymous functions are shown. I'd expect this line to look like
```suggestion
        browser.waitUntil(() => page.chartTitle.isDisplayed(), 5000);
```
if no other changes are made

palakpatel311: 5000 was used to just wait until 5000 secs for chartTitle to be appear. I dont want it to wait for default time out which is more secs. 
I moved 5000 in config file so any test can use it.

andlar (approved):

athomas-git (approved):

PR #830, opened 2020-11-19
OCD 3554: AQA test for new role ONC_STAFF

palakpatel311:


e2e/pages/users/user.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        actionBarComponent.save();
+    });
+
+    it('change title successfully', () => {
+        page.editUser('AQA ONC STAFF');
+        page.userTitle.clearValue();
+        page.userTitle.addValue('Mr');
+        actionBarComponent.save();
+        assert.include(page.userInformation('AQA ONC STAFF').getText(),'Mr');
+    });
```
ashwinimore, line 34: I am thinking ....when this test runs for the first time, it works for the purpose where the original value is cleared and then 'Mr' is added & saved. On next run, even if the save doesn't go through, the test will pass because it had worked on previous run and 'Mr' input was saved. Does that sound correct? If yes, then using something dynamic like date function can help with new input value everyday (won't work for multiple runs on same day but good for overnight runs). I have same thoughts on phoneNumber test below.

palakpatel311: Maint plan in bamboo does take copy of DB before running any test and sets it back after test to the same DB copy as taken before test.

e2e/components/login/login.readonly.spec.js
```diff
@@ -49,6 +49,18 @@ describe('when logging in', () => {
...
         expect(component.toggleLoginComponent.getText()).toBe('AQA ACB');
     });
+
+    it('should be able to log in as ONC_STAFF with an email address', () => {
+        component.logInWithEmail('oncstaff');
+        component.logoutButton.waitForDisplayed();
+        expect(component.toggleLoginComponent.getText()).toBe('AQA ONC STAFF');
+    });
+
+    it('should be able to log in as developer with an email address', () => {
```
andlar: This test will fail (or at lest should) on STG, as this depends on the role-developer flag being turned on

e2e/pages/users/user.spec.js
```diff
@@ -0,0 +1,61 @@
...
+
+    it('change phone number successfully', () => {
+        page.editUser('AQA ONC STAFF');
+        page.userPhoneNumber.clearValue();
+        page.userPhoneNumber.addValue('000-000-0000');
```
andlar: What if you generated a phone number by doing something like:
```
let number = (new Date()).getTime() % 1000000;
```
That way you'd have a number that would be mostly unique each time it ran, and (on the off change) someone editing this user to have a phone number of 000-000-0000 we'd still have a valid test

e2e/pages/users/user.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        loginComponent.logInWithEmail('oncstaff');
+        page.usersButton.click();
+        page.userManagementButton.click();
+    });
+
+    afterAll(function () {
+        page.lockedCheckbox.scrollAndClick();
+        page.enabledCheckbox.scrollAndClick();
+        page.pwChangeCheckbox.scrollAndClick();
+        actionBarComponent.save();
```
andlar: If this works I think that's a problem. I don't think a user should be able to unlock themselves, even if they're logged in. Or at least, there's an argument to be made that that could be the case. It's possible that maybe _no_ user should be able to modify their own "checkbox" states. One way to fix that in this test would be for the acting user to be not ROLE_ONC_STAFF, but instead ROLE_ONC or ROLE_ADMIN. I recognize that doesn't help with the requirement for ROLE_ONC_STAFF to be able to manage other ROLE_ONC_STAFF users, but I think the only way to do that would be to create two ROLE_ONC_STAFF users and use one to manage the other

andlar (changes requested):

e2e/pages/users/user.spec.js
andlar: Palak's right about the plan resetting the database after the run, which is why this test isn't in a `user.readonly.spec.js` file. That said, it wouldn't be a bad idea to make the title that the user is set to have be something that's somewhat random, even if the randomness is just based on a `new Date()` value in some way. That way, if for some reason someone edits this user, outside of the tests, to have a title of "Mr", the test will still be valid

e2e/components/login/login.readonly.spec.js
palakpatel311: Oh yes, I can comment it out for now then

e2e/pages/users/user.spec.js
palakpatel311: Done

e2e/pages/users/user.spec.js
palakpatel311: Done

e2e/pages/users/user.spec.js
palakpatel311: ROLE_ONC_STAFF can check checkboxes for  unlock, enabled them selves and password change in next login right now. So this is a problem/issue. This is open question to PO.
I have changed this TC as admin logs in and edit on staff roles details.

e2e/components/login/login.readonly.spec.js
```diff
@@ -56,11 +56,11 @@ describe('when logging in', () => {
         expect(component.toggleLoginComponent.getText()).toBe('AQA ONC STAFF');
     });
 
-    it('should be able to log in as developer with an email address', () => {
-        component.logInWithEmail('developer');
-        component.logoutButton.waitForDisplayed();
-        expect(component.toggleLoginComponent.getText()).toBe('AQA Developers');
-    });
+    // it('should be able to log in as developer with an email address', () => {
```
andlar: Not sure if our system supports it for the AQA tests, but the unit tests in this feature can ignore a test by marking it as `xit` instead of `it`, and it will e`x`clude that test
```suggestion
    xit('should be able to log in as developer with an email address', () => {
```

ashwinimore (dismissed):

e2e/components/login/login.readonly.spec.js
```diff
@@ -49,6 +49,18 @@ describe('when logging in', () => {
...
     });
+
+    it('should be able to log in as ONC_STAFF with an email address', () => {
+        component.logInWithEmail('oncstaff');
+        component.logoutButton.waitForDisplayed();
+        expect(component.toggleLoginComponent.getText()).toBe('AQA ONC STAFF');
+    });
+
+    xit('should be able to log in as developer with an email address', () => {
+    //     component.logInWithEmail('developer');
```
andlar: When the test is ignored, the code inside the test shouldn't be commented out. We shouldn't have commented out code in our system at all

andlar (dismissed):

ashwinimore (dismissed):

kekey1 (approved):

andlar (approved):

PR #832, opened 2020-12-02
OCD 3208: AQA test for merging versions

palakpatel311:


e2e/pages/organizations/developers/developers.po.js
```diff
@@ -113,6 +116,26 @@ class DevelopersPage {
         this.getEditButton(product).click();
         product.$('.product__product-info-item-action').$('.dropdown-menu').$$('li')[1].click();
     }
+
+    getVersionMerge () {
```
andlar: The method name suggests it returns an object, but the functionality does a `click`. That disconnect makes this harder to tell what's happening

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -99,6 +99,7 @@ describe('the Developers page', () => {
                         expect(contact.get(product)).toHaveTextContaining(poc.phone);
                     });
                 });
+
```
andlar: Is there a reason to add this space? I'd rather have spaces added in between the `it` blocks, and not during all of the closing places. That's the way the ~1700-ish unit tests have the spacing laid out (mostly)

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -214,4 +215,73 @@ describe('the Developers page', () => {
...
+                    expect(page.getVersionCount(product).getText()).toBe('7 Versions');
+                });
+                it('version name is required field', () => {
+                    page.mergeButton.click();
+                    page.getVersionMerge();
+                    page.moveVersionToBeMerged(0);
+                    page.versionName.clearValue();
+                    expect(page.errorMessage.getText()).toBe('Field is required');
+                });
+                it('should create new version successfully', () => {
```
andlar, line 254: What happens if this test runs before the two above? If tests can be rearranged and they fail, then the tests aren't independent. I'd suggest that for any test that changes data, it should use data that no other test depends on. I think both of the above tests should be in a `.readonly` file, as they could be run at any time

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -214,4 +215,73 @@ describe('the Developers page', () => {
...
+        describe('when on the "Greenway Health, LLC" Developer page', () => {
+            beforeEach(() => {
+                let developer = 'Greenway Health, LLC';
+                page = new DevelopersPage();
+                page.selectDeveloper(developer);
+                page.getDeveloperPageTitle(developer).waitForDisplayed();
+            });
+
+            describe('when merging versions of "Greenway Intergy Meaningful Use Edition" product', () => {
+                let name = 'Greenway Intergy Meaningful Use Edition';
```
andlar: For these things, if they're `let`, they should be defined here as `let name;`, and instanciated in the `beforeEach` block. If they're constants they can defined here as `const name = 'Greenway...';`

andlar (changes requested):

e2e/pages/organizations/developers/developers.spec.js
palakpatel311: Above 2 tests uses admin login so I don't think those should be in .readonly spec file.

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -214,4 +214,86 @@ describe('the Developers page', () => {
...
+                    page.mergeButton.click();
+                    page.versionMergeButton.click();
+                    page.versionName.addValue(newVersion);
+                    actionBar.cancel();
+                    actionConfirmation.yes.click();
+                    page.productsHeader.waitForDisplayed();
+                    product = page.getProduct(name);
+                    product.scrollIntoView({block: 'center', inline: 'center'});
+                    browser.waitUntil(() => page.getVersionCount(product).isDisplayed());
+                    expect(page.getVersionCount(product).getText()).toBe('7 Versions');
```
kekey1: Instead of looking for a count (which could *not* be 7 if the test below that actually completes merge runs BEFORE this test - and the order is not guaranteed) could you just assert that "v11" is still in the list of available versions to merge? Meaning that nothing happened and the cancel worked.

e2e/pages/organizations/developers/developers.spec.js
kekey1: I don't know the standard for when something is readonly but I tend to agree with Andrew that the tests above do not modify the developer data since they do not complete a merge. I don't really see login as changing data (although I could see the argument of it changes the user's last login time but that seems trivial). I'd vote for moving the tests that cancel the merge and try to complete the merge w/o required data to readonly

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -214,4 +214,86 @@ describe('the Developers page', () => {
...
+                    product.scrollIntoView({block: 'center', inline: 'center'});
+                    page.selectProduct(product);
+                    page.getProductInfo(product).waitForDisplayed({timeout: 55000});
+                    page.selectVersion(product, productId, version);
+                });
+
+                it('should create new version successfully', () => {
+                    page.mergeButton.click();
+                    page.versionMergeButton.click();
+                    page.moveVersionToBeMerged(0);
```
kekey1: This is hard to follow for me as far as what is expected from the test. You are merging "v10.10" with "page.moveVersionToBeMerged(0)" - so the first version in the list of available versions. I would find it easier to understand if you specified which version you are merging ("v12" or whatever) and then in the things you assert at the end of the test you can check that neither "v10.10" nor "v12" show up as versions of this product.

kekey1 (changes requested):

kekey1 (dismissed):

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -1,18 +1,25 @@
...
 import Hooks from '../../../utilities/hooks';
-
-let contact, hooks, page;
+import LoginComponent from '../../../components/login/login.po';
+import ActionBarComponent from '../../../components/action-bar/action-bar.po';
+import ActionConfirmationComponent from '../../../components/action-confirmation/action-confirmation.po';
+
+let actionBar, actionConfirmation, contact, hooks,login, page;
+
+beforeEach(async () => {
```
andlar: Can we put this beforeEach block back inside the `the Developers page` describe block, and then move the `when Logged in as an admin` block inside that one too? The actions that are happening in this beforeEach block aren't described anywhere, either with a `describe` or with an `it` block, and I think that makes it less clear when/why things are happening

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -95,3 +102,63 @@ describe('When a user is on the Developer page for a Developer that doesn\'t exi
...
+
+    afterEach(() => {
+        login.logOut();
+    });
+
+    describe('when on the "Greenway Health, LLC" Developer page', () => {
+        const name = 'Greenway Intergy Meaningful Use Edition';
+        const productId = 837;
+        let product, version;
+        let timestamp = (new Date()).getTime();
```
andlar:
```suggestion
        let timestamp;
```

Setting the variable to a value should happen in the 'beforeEach' block. That way, if there were multiple `it` blocks covered by the `beforeEach`, they would each have a different value for the timestamp, which helps ensure we're actually using "random-ish" values for these things

e2e/pages/organizations/developers/developers.readonly.spec.js
```diff
@@ -95,3 +102,63 @@ describe('When a user is on the Developer page for a Developer that doesn\'t exi
...
+        beforeEach(() => {
+            let developer = 'Greenway Health, LLC';
+            page = new DevelopersPage();
+            page.selectDeveloper(developer);
+            page.getDeveloperPageTitle(developer).waitForDisplayed();
+        });
+
+        describe('when merging versions of "Greenway Intergy Meaningful Use Edition" product', () => {
+            version = 'v11';
```
andlar: This one and the `newVersion` values should be set in the beforeEach block too

e2e/pages/organizations/developers/developers.readonly.spec.js
andlar: The "not readonly" script has the above described structure, and I think it makes it more clear

palakpatel311: If I move beforeEach block inside describe block, not all other previously added tests needs to have login. Only test which needs login is the one I added. 
I also thought of keeping my added test out of these describe block, bit in that case, it will be duplicate lines of code for the same beforeEach block along with login code.

e2e/pages/organizations/developers/developers.readonly.spec.js
palakpatel311: Did.

e2e/pages/organizations/developers/developers.readonly.spec.js
palakpatel311: Done

kekey1 (dismissed):

e2e/pages/organizations/developers/developers.readonly.spec.js
andlar: While it's true that only a few of the other tests need to have a logged in user, you could nest those tests inside the "the Developers page" block too. The other problem with moving this beforeEach block out here is that it now is affecting the block with the describe: "When a user is on the Developer page for a Developer that doesn\'t exist", which it shouldn't do

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -214,4 +214,59 @@ describe('the Developers page', () => {
...
+
+        afterEach(() => {
+            login.logOut();
+        });
+
+        describe('when on the "Greenway Health, LLC" Developer page', () => {
+            const name = 'Greenway Intergy Meaningful Use Edition';
+            const productId = 837;
+            let product, version;
+            let timestamp = (new Date()).getTime();
```
andlar: This assignment should happen inside the `beforeEach` block

e2e/pages/organizations/developers/developers.spec.js
```diff
@@ -214,4 +214,59 @@ describe('the Developers page', () => {
...
+                page = new DevelopersPage();
+                page.selectDeveloper(developer);
+                page.getDeveloperPageTitle(developer).waitForDisplayed();
+                hooks.waitForSpinnerToDisappear();
+            });
+
+            describe('when merging versions of "Greenway Intergy Meaningful Use Edition" product', () => {
+                version = 'v10.10';
+                const versionToBeMerged = 'v9.30';
+                let newVersion = 'New version - ' + timestamp;
```
andlar: Same comment re: assignment

andlar (dismissed):

kekey1 (dismissed):

andlar (commented):
Why does this PR have all of the commits from the feature branch for OCD-3208?

andlar (approved):

kekey1 (approved):

PR #833, opened 2020-12-07
OCD-3549 -- Allow upload of Listings with variable column sets 

andlar:


src/app/components/listing/inspect/inspect.controller.js
```diff
@@ -5,7 +5,7 @@
         .controller('InspectController', InspectController);
 
     /** @ngInject */
-    function InspectController ($log, $uibModal, $uibModalInstance, developers, inspectingCp, networkService, resources, utilService) {
+    function InspectController ($log, $uibModal, $uibModalInstance, beta, developers, inspectingCp, networkService, resources, utilService) {
```
tmy1313, line 8: Is this a flag to turn a feature on/off?  If so, should we be using a flag?

tmy1313 (dismissed):

kekey1 (dismissed):

kekey1 (dismissed):

src/app/components/listing/inspect/inspect.controller.js
andlar: There is a flag, but both versions of the feature need to exist on the page(s) simultaneously, so this value is here to tell the component which version to render

tmy1313 (dismissed):

kekey1 (approved):

tmy1313 (approved):

PR #845, opened 2021-01-12
OCD-3545: Add XSD modification

kekey1:
[#OCD-3545]


andlar (approved):
Definitely appreciate the contents of the file. I like seeing why this is here, as it helps us make sure we still have it when it should be there

tmy1313 (approved):

PR #847, opened 2021-01-15
OCD 3584: ONC Staff access to scheduled and background job page

palakpatel311:


e2e/pages/administration/jobs/background/background.spec.js
```diff
@@ -0,0 +1,30 @@
...
+let background, hooks, loginComponent;
+
+beforeEach(async () => {
+    loginComponent = new LoginComponent();
+    background = new BackgroundPage();
+    hooks = new Hooks();
+    await hooks.open('#/search');
+});
+
+describe('ONC-Staff user can see', () => {
```
andlar: When in `describe` blocks, the phrasing I usually use is something where the text in the describe phrase should explain the setup, so here it'd be something like 
```suggestion
describe('when an ONC-Staff user is logged in', () => {
```

e2e/pages/administration/jobs/background/background.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    beforeEach(function () {
+        loginComponent.logInWithEmail('oncstaff');
+        hooks.open('#/administration/jobs/background');
+    });
+
+    it('correct background jobs', () => {
```
andlar: `it` blocks be should in the form "should <fill in the blank>", so here I'd go with something like 
```suggestion
    it('should see the right set of background jobs', () => {
```

e2e/pages/administration/jobs/background/background.spec.js
```diff
@@ -0,0 +1,30 @@
...
+        var actualResult = [];
+        var expectedResult = ['MUU Upload','Surveillance Upload','Export Quarterly Report','Export Annual Report'];
+        var length = background.backgroundJobRowsCount;
+        for ( var j = 0; j < length; j++ ) {
+            actualResult.push(background.backgroundJobNames(j).getText());
+        }
+        assert.equal(actualResult.toString(),expectedResult.toString());
```
andlar: This test doesn't change data, right? So the file could be `background.readonly.spec.js` right?

e2e/pages/administration/jobs/scheduled/scheduled.spec.js
```diff
@@ -0,0 +1,30 @@
...
+let hooks, loginComponent, scheduled;
+
+beforeEach(async () => {
+    loginComponent = new LoginComponent();
+    scheduled = new ScheduledPage();
+    hooks = new Hooks();
+    await hooks.open('#/search');
+});
+
+describe('ONC-Staff user can see', () => {
```
andlar: Same comments re: format of `describe` and `it` blocks, as well as the filename itself

andlar (changes requested):

e2e/pages/administration/jobs/background/background.po.js
```diff
@@ -0,0 +1,18 @@
+const backgroundElements = {
+    backgroundJobsRoot: 'chpl-jobs-background-types',
+};
+
+class BackgroundPage {
+    constructor () { }
+
+    get backgroundJobRowsCount () {
```
andlar: I think this element should just be the rows, not the count of them. A PageObject element that is the count of something isn't really useful for any other purpose, while an object that is a set of Rows could be used to get a count, to iterate through to find an object, to see if it's null, etc. All of those actions could be used in the logic portion of a spec file, but only if the PageObject element is the actual element that's on the page

e2e/pages/administration/jobs/background/background.po.js
```diff
@@ -0,0 +1,18 @@
...
+        return $(backgroundElements.backgroundJobsRoot).$('tbody').$$('tr').length;
+    }
+
+    backgroundJobNames (rowNumber) {
```
andlar: The object title suggests this is multiple objects that's returned, but it looks like it's only returning the name on the given row, which is singular. Which is it?

e2e/pages/administration/jobs/scheduled/scheduled.po.js
```diff
@@ -0,0 +1,18 @@
+const scheduledElements = {
+    scheduledJobsRoot: 'chpl-jobs-scheduled-jobs',
+};
+
+class ScheduledPage {
+    constructor () { }
+
+    get scheduledJobRowsCount () {
```
andlar: Same comments re: style & pluralization as in the background.po.js file

e2e/pages/administration/jobs/scheduled/scheduled.readonly.spec.js
```diff
@@ -0,0 +1,30 @@
...
+let hooks, loginComponent, scheduled;
+
+beforeEach(async () => {
+    loginComponent = new LoginComponent();
+    scheduled = new ScheduledPage();
+    hooks = new Hooks();
+    await hooks.open('#/search');
+});
+
+describe('When an ONC-Staff user is logged in', () => {
```
andlar:
```suggestion
describe('when an ONC-Staff user is logged in', () => {
```
I'd prefer to keep all of our `describe` and `it` block text lower-case. The basic reason is that these sentence fragments can be combined to make full sentences, but the nesting can change in a variety of ways, and they can change over time, and if we start with a capital letter at one point, but then nest that block later, the capital letter will show up in the middle of a sentence. It's simpler to just keep them all as lowercase, always

andlar (dismissed):
I'm leaving some comments with a few things I'd prefer to see changed, or addressed, or discussed, but am okay with this going forward as is as far as the logic goes

andlar (dismissed):

e2e/pages/administration/jobs/scheduled/scheduled.readonly.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe('when an ONC-Staff user is logged in', () => {
+    beforeEach(function () {
+        loginComponent.logInWithEmail('oncstaff');
+        hooks.open('#/administration/jobs/scheduled');
+    });
+
+    it('should see the right set of scheduled jobs', () => {
+        var actualResult = [];
+        var expectedResult = ['All Broken Surveillance Rules Report','Developer Access Report','Inherited Certification Status Errors Report','Overnight Broken Surveillance Rules Report','Pending "Change Request" Report','Questionable Activity Report','Questionable URL Report','Summary Statistics Email','Trigger Developer Ban Notification'];
```
athomas-git: Is 'Real World Testing Email Report' job missing for ONC_STAFF ?

palakpatel311: Yes you are right. Added.

athomas-git (dismissed):

andlar (dismissed):

ashwinimore (dismissed):

athomas-git (dismissed):

e2e/components/cms-widget/cms-widget.po.js
```diff
@@ -85,8 +85,10 @@ class CmsWidgetComponent {
 
   addListingToCms (listingId) {
     this.certIdButton(listingId).scrollIntoView({block: 'center', inline: 'center'});
+    browser.pause(2000);
```
andlar: Can this be made explicit? After adding an item to the widget, the "processing" thing appears, and the length of the list changes from "whatever it was before" to "before + 1". Either of those things could be used as a check on when it was done processing

e2e/pages/administration/jobs/background/background.readonly.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  background = new BackgroundPage();
+  hooks = new Hooks();
+  await hooks.open('#/search');
+});
+
+describe('when an ONC-Staff user is logged in', () => {
+  beforeEach(function () {
+    loginComponent.logInWithEmail('oncstaff');
+    hooks.open('#/administration/jobs/background');
+    browser.pause(2000); //Have to add this wait as jobs gets in order in micro seconds.
```
andlar: If we don't care about the order of the Jobs, and instead only care about the existence of them, we wouldn't need to wait. The test below could loop through the jobs that were found and compare them to a Set of expected jobs, as well as comparing the length of the list to the size of the Set. Any jobs found that don't exist in the Set, or any difference in length, could be asserted against and used as a test

andlar: The same could work with the other type of jobs page

andlar (changes requested):

e2e/pages/administration/jobs/scheduled/scheduled.readonly.spec.js
```diff
@@ -0,0 +1,40 @@
...
+  scheduled = new ScheduledPage();
+  hooks = new Hooks();
+  await hooks.open('#/search');
+});
+
+describe('when an ONC-Staff user is logged in', () => {
+  beforeEach(function () {
+    loginComponent.logInWithEmail('oncstaff');
+    hooks.open('#/administration/jobs/scheduled');
+    browser.pause(2000); //Have to add this wait as jobs gets in order in micro seconds
```
andlar:
```suggestion
```

e2e/pages/administration/jobs/scheduled/scheduled.readonly.spec.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+    var isExist;
+    assert.equal(actualResult.size,expectedResult.size);
+    for (var value of actualResult) {
+      if (!expectedResult.has(value)) {
+        return isExist = false;
+      }
+      return isExist = true;
+    }
+    assert.isTrue(isExist);
```
andlar: We shouldn't need to loop through these things twice

In pseudocde it should be done with something like:
```
const expected = new Set(['things here'])
// expect the table to have the same number of rows as the expected data
expect(scheduled.scheduledJobsRows.length).toBe(expected.size)

// expect each row in the table to be in the set of expected data
scheduled.scheduledJobRows.forEach(row => {
  expect(expected.contains(row.getText()).toBe(true)
})
```

andlar (changes requested):

e2e/pages/administration/jobs/background/background.readonly.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  beforeEach(function () {
+    loginComponent.logInWithEmail('oncstaff');
+    hooks.open('#/administration/jobs/background');
+  });
+
+  it('should see the right set of background jobs', () => {
+    const expected = new Set(['Export Annual Report','Export Quarterly Report','MUU Upload','Surveillance Upload']);
+    assert.equal(background.backgroundJobRows.length,expected.size);
+    background.backgroundJobRows.forEach(row => {
+      assert.exists(expected , row.getText());
```
andlar: This test won't work as anticipated. Please check out the PR I created that touches on this

https://github.com/palakpatel311/chpl-website/pull/9/files

andlar (changes requested):

e2e/pages/administration/jobs/background/background.readonly.spec.js
palakpatel311: I kept the comments for the future references for myself.

andlar (approved):

athomas-git (approved):

ashwinimore (approved):

PR #854, opened 2021-02-09
OCD-3599: Add React as usable component

andlar:
I have an error in the console about needing a key in the list that makes up the FuzzyTypes, but don't know how to get them there. Would appreciate a suggestion


src/app/components/action-bar/action-bar.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+//const dependencies = {
+//  getChplLogService: () => require('../../services/services-bridge.jsx').default,
+//};
+
+function ChplActionBar ({
+  errorMessages,
+  isDisabled,
+  takeAction,
+  warningMessages,
```
Plow74: Instead of passing in this object, use props. Props are immutable so if the values being passed in need to be changed, then a new variable will need to be used, ex: let _errorMessages = props.errorMessages. It could be that React will see this object as props anyway, but the convention is to use the props keyword.

src/app/components/action-bar/action-bar.jsx
```diff
@@ -0,0 +1,122 @@
...
+}) {
+  const [acknowledgeWarnings, setAcknowledgeWarnings] = useState(false);
+  const [errors] = useState(errorMessages ? errorMessages.sort((a, b) => a < b ? -1 : a > b ? 1 : 0) : []);
+  const [showMessages, setShowMessages] = useState(false);
+  const [warnings] = useState(warningMessages ? warningMessages.sort((a, b) => a < b ? -1 : a > b ? 1 : 0) : []);
```
Plow74: I think this is fine but there is probably a bigger conversation here on error handling in general. Should an action bar be responsible for showing errors? Should there be an component that is responsible for showing feedback from the system? Should we use Error Boundries? https://reactjs.org/docs/error-boundaries.html

src/app/components/fuzzy-type/fuzzy-type.jsx
```diff
@@ -0,0 +1,167 @@
...
+import {arrayOf, func, shape, string} from 'prop-types';
+import {getAngularService} from './';
+
+import {ChplActionBar} from '../action-bar/';
+
+const dependencies = {
+  getChplLogService: () => require('../../services/services-bridge.jsx').default,
+};
+
+function ChplFuzzyType ({fuzzyType, takeAction}) {
```
Plow74: Consider changing this to use the props convention.

src/app/components/util/ellipsis.jsx
```diff
@@ -0,0 +1,45 @@
+import React, {useState} from 'react';
+import ReactTooltip from 'react-tooltip';
+import {bool, number, string} from 'prop-types';
+
+function ChplEllipsis ({text, maxLength, wordBoundaries}) {
```
Plow74: Consider using the props convention.

src/app/components/fuzzy-type/fuzzy-type.jsx
```diff
@@ -0,0 +1,167 @@
...
+  const cancelAdd = () => {
+    setAdding(false);
+    setNewFuzzyType(undefined);
+    setShowFormErrors(false);
+  };
+
+  const edit = () => {
+    chplLog.info('edit');
+    setEditing(true);
+    takeAction(fuzzyType, 'edit');
```
Plow74: Should takeAction return a promise so that updating state only happens once the promise resolves?

Plow74 (dismissed):

tmy1313 (dismissed):

Plow74 (approved):
Everything looks good to me!

kekey1 (approved):

tmy1313 (approved):

PR #876, opened 2021-03-26
OCD-3596 - API Key registration should require additional security measures

tmy1313:


package.json
```diff
@@ -20,16 +23,18 @@
     "e2e:lint": "eslint e2e",
     "e2e:clean": "rm test_reports/e2e/* || true",
     "report:generate": "npx allure generate --clean ./test_reports/e2e/allure-results/ -o ./test_reports/e2e/allure-report",
-    "report:open": "npx allure open ./test_reports/e2e/allure-report"
+    "report:open": "npx allure open ./test_reports/e2e/allure-report",
+    "test:react": "jest"
```
andlar: Can we move this to be in the section of scripts that has the other `test:*` scripts?

package.json
```diff
@@ -20,16 +23,18 @@
...
   "dependencies": {
     "@babel/polyfill": "^7.4.3",
-    "@babel/preset-react": "^7.12.10",
+    "@babel/preset-react": "^7.12.13",
     "@babel/register": "^7.4.0",
     "@js-joda/core": "^3.1.0",
     "@js-joda/locale_en-us": "^3.1.1",
     "@js-joda/timezone": "^2.3.0",
     "@material-ui/core": "^4.11.3",
+    "@material-ui/lab": "^4.0.0-alpha.57",
```
andlar: I'm leery of `alpha` stuff in a prod system. Is this required?

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -0,0 +1,84 @@
...
+import {string} from 'prop-types';
+import { ThemeProvider } from '@material-ui/core/styles';
+import theme from '../../themes/theme';
+import {makeStyles} from '@material-ui/core/styles';
+import Card from '@material-ui/core/Card';
+import CardContent from '@material-ui/core/CardContent';
+import CardHeader from '@material-ui/core/CardHeader';
+import Typography from '@material-ui/core/Typography';
+import Alert from '@material-ui/lab/Alert';
+import {getAngularService} from './';
```
andlar: I've been thinking about how we organize our imports. I have been leaning towards something like:

```
import React
import (things that come from React)
import (things that come from Material-UI)
blank line
import (things that come from our code)
```

My theory is that the first lines are the most "required" things, with "theming" stuff as secondary, and separating them all from our custom-written code with a blank line. I'm up for conversations about this, but wanted to put the idea out there

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -0,0 +1,84 @@
...
+  },
+  centeredGridItem: {
+    gridColumn: '2 / 4',
+  },
+}));
+
+function ChplApiKeyConfirm (props) {
+  const networkService = getAngularService('networkService');
+  const [apiKey, setApiKey] = useState({});
+  const [confirmError, setConfirmError] = useState(false);
```
andlar, line 26: I'm wondering about conventions here too. I was starting with this philosophy:

```
set state(s)
getAngularService(s)
useStyles()
was there anything else?
```

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -0,0 +1,84 @@
...
+          <Card>
+            <CardHeader title="API Email Confirmation"/>
+            <CardContent>
+              { confirmError ?
+                <>
+                  <Alert severity="error" data-testid="api-key-alert">There was an error confirming your email!</Alert>
+                  <Typography variant="body1" paragraph={ true }>
+                    Possible reasons:
+                  </Typography>
+                  <ul className="MuiTypography-body1">
```
andlar: I think this should just be `body1`, though I'm not sure. I _think_ we shouldn't be putting classes with `Mui` in their names in our code directly

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -0,0 +1,84 @@
...
+                    <li>This email confirmation request has already been processed.  Be sure to check you email for an email with your API key.</li>
+                    <li>The confirmation token in the URL is not correct.</li>
```
andlar: Both of these `li` items can't exist at the same time, right?

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -0,0 +1,84 @@
...
+                :
+                <>
+                  <Typography variant="body1" paragraph={ true }>
+                    Congratulations on sucessfully registering your email for use with CHPL API.  CHPL has sent you an email containing your new API key.
+                  </Typography>
+                  <Typography variant="body1" paragraph={ true } data-testid="api-key-display">
+                    Your API key is: &nbsp; <strong>{ apiKey.apiKey }</strong>
+                  </Typography>
+                  <Typography variant="body1" paragraph={ true }>
+                    Your API key must be included in all API calls via either a header with the name &lsquo;API-Key&lsquo; or as a URL parameter named &lsquo;api_key&lsquo;.
```
andlar: Can we remove the special left/right quote marks? We don't have them anywhere else in the CHPL, and I'd like to keep that consistent

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -0,0 +1,84 @@
...
+                    <li>The confirmation token in the URL is not correct.</li>
+                  </ul>
+
+                </>
+                :
+                <>
+                  <Typography variant="body1" paragraph={ true }>
+                    Congratulations on sucessfully registering your email for use with CHPL API.  CHPL has sent you an email containing your new API key.
+                  </Typography>
+                  <Typography variant="body1" paragraph={ true } data-testid="api-key-display">
```
andlar: A question of style. There are three `Typography` blocks here all with the same attributes. Can they / should they be combined? Can we put these three lines in one block, maybe with `<br />` tags? I honestly don't know, and am looking for the most idiomatic way of doing this kind of thing going forward

src/app/components/api-key/api-key-registration.jsx
```diff
@@ -0,0 +1,115 @@
...
+const validationSchema = yup.object({
+  email: yup.string()
+    .required('Email is required')
+    .email('Enter a valid email'),
+  nameOrganization: yup.string()
+    .required('Name or Organization is required'),
+});
+
+function ChplApiKeyRegistration () {
+  //const $log = getAngularService('$log');
```
andlar: Probably don't need this

src/app/components/api-key/api-key-registration.jsx
```diff
@@ -0,0 +1,115 @@
...
+    initialValues: {email: '', nameOrganization: ''},
+    validationSchema: validationSchema,
+    onSubmit: (values) => {
+      createRequest(values);
+    },
+    validateOnChange: false,
+    validateOnMount: true,
+  });
+
+  const createRequest = (values) => {
```
andlar, line 47: I usually don't include the parentheses when the function has only one parameter. I'm not against having it, but if we choose to go that way we should make that part of our style guide, whether or not our linting tools can catch it

```suggestion
  const createRequest = values => {
```

src/app/components/api-key/api-key-registration.jsx
```diff
@@ -0,0 +1,115 @@
...
+    networkService.requestApiKey({ email: values.email, name: values.nameOrganization })
+      .then((response) => {
+        if (response.success) {
+          toaster.pop({
+            type: 'success',
+            body: 'To confirm your email address, an email was sent to: ' + values.email + '  Please follow the instructions in the email to obtain your API key.',
```
andlar: Does the response have the email address? I usually try to use the response to fill out the responding messages, but if that data isn't there this is fine

src/app/components/api-key/api-key-registration.jsx
```diff
@@ -0,0 +1,115 @@
...
+        }
+      }, (error) => {
+        toaster.pop({
+          type: 'error',
+          body: error.data.errorMessages[0],
+        });
+      });
+  };
+
+  const classes = useStyles();
```
andlar: I feel like this should be up at the start of the function with the other "set up the function" things. I could see the argument for it being here, just before it's used, so am up for discussion

src/app/components/api-key/api-key-registration.jsx
```diff
@@ -0,0 +1,115 @@
...
+            <TextField fullWidth
+                       variant="outlined"
+                       id='name-organization'
+                       name='nameOrganization'
+                       label='Name or Organization'
+                       value={ formik.values.nameOrganization }
+                       onChange={ formik.handleChange }
+                       onBlur={ formik.handleBlur }
+                       error={ formik.touched.nameOrganization && Boolean(formik.errors.nameOrganization) }
+                       helperText={ formik.touched.nameOrganization && formik.errors.nameOrganization } />
```
andlar: I want to call this out as as example of what I think we should do for all of our input things, specifically with respect to the order of things, and one that I'd like to repeat:

1. Styles
2. id
3. name
4. label
5. value
6. on*
7. error*

src/app/components/api-key/api-key-registration.test.jsx
```diff
@@ -0,0 +1,225 @@
+import React from 'react';
+import * as angularReactHelper from '../../services/angular-react-helper.jsx';
+import { render, cleanup, screen, waitFor, fireEvent } from '@testing-library/react';
+import { toBeInvalid, toBeValid, toBeEnabled, toBeDisabled, toHaveValue } from '@testing-library/jest-dom';
```
andlar: I've been trying to alphabetize the imports where I can. I don't know if we can lint that, but I think it'd be helpful for consistency

```suggestion
import { toBeDisabled, toBeEnabled, toBeInvalid, toBeValid, toHaveValue } from '@testing-library/jest-dom';
```

src/app/components/api-key/api-key-registration.test.jsx
```diff
@@ -0,0 +1,225 @@
...
+when(angularReactHelper.getAngularService).calledWith('networkService').mockReturnValue(networkServiceMock);
+when(angularReactHelper.getAngularService).calledWith('toaster').mockReturnValue(toasterMock);
+
+describe('the ChplApiKeyRegistration component', () => {
+
+  afterEach(() => {
+    cleanup();
+  })
+
+  //it('should renders without crashing', () => {
```
andlar: Probably can be removed, or maybe `xit` (or whatever jest does)

src/app/components/api-key/api-key-registration.test.jsx
```diff
@@ -0,0 +1,225 @@
...
+
+      userEvent.type(email, 'abc');
+      userEvent.tab();
+
+      await waitFor(() => expect(email).toBeInvalid());
+    });
+  });
+
+  describe('when a response is received from API', () => {
+
```
andlar:
```suggestion
```

src/app/components/api-key/api-key-registration.test.jsx
```diff
@@ -0,0 +1,225 @@
...
+    it('should display a toaster indicating success when the response is success', async () => {
```
andlar, line 166: I just noticed the `async` in the `it` definition. Am curious about that; what's it for? I'm assuming it's required, but I'd like to know more about what's going on for my own edification

src/app/components/api-key/api-key-registration.test.jsx
```diff
@@ -0,0 +1,225 @@
...
+      userEvent.tab();
+      userEvent.click(registerButton);
+      await waitFor(() => {
+        expect(toasterMock.pop).toHaveBeenCalledWith({
+          type: 'error',
+          body: 'ErrorMessage to display',
+        });
+      });
+    });
+
```
andlar:
```suggestion
```

These tests are great. I'm super excited to fold this pattern into the rest of the stuff we work on. Good work!

src/app/components/components.module.js
```diff
@@ -21,4 +23,6 @@ angular
   ])
   .component('chplEllipsisBridge', reactToAngularComponent(ChplEllipsis))
   .component('chplFuzzyTypeBridge', reactToAngularComponent(ChplFuzzyType))
+  .component('chplApiKeyConfirmBridge', reactToAngularComponent(ChplApiKeyConfirm))
```
andlar, line 25: I've been trying to alphabetize both the `.component` lines, and the `import` ones above. I don't know that it matters, but I feel it mimics what we have in our java `import` blocks in some way, and I hope it might help us recognize issues in the future

src/app/services/network.service.js
```diff
@@ -27,6 +27,10 @@ export class NetworkService {
     return this.apiPOST(url, userObj);
   }
 
+  confirmApiKey (hash) {
+    return this.apiPOST('/key/confirm', hash);
```
andlar, line 32: The `network.service` file is the _one file_ I think we have a lot of code in, that has nearly 100% coverage. Can you add tests to the relevant spec file for this service?

yarn-error.log
```diff
@@ -0,0 +1,17858 @@
+Arguments:
```
andlar: This file shouldn't be committed to git. I'm assuming it was an accidental response to a `git add .`, but it shouldn't be here

src/app/pages/resources/chpl-api/chpl-api.html
```diff
@@ -28,12 +28,7 @@ <h2>Access API Documentation</h2>
         </p>
       </div>
       <div class="col-sm-5">
-        <div class="panel panel-ai">
-          <div class="panel-heading">Register</div>
-          <div class="panel-body">
-            <ai-api-key></ai-api-key>
```
andlar, line 34: Does the `aiApiKey` component still exist? Is it used somewhere else? If so, no issues, but if it's removed from use we should delete it from the project

src/app/pages/registration/registration.state.js
```diff
@@ -19,7 +19,7 @@ let states = [{
   data: { title: 'CHPL Registration' },
 },{
   name: 'registration.confirm-user',
-  url: '/confirm-user/{hash}',
+  url: '/~/{hash}',
```
andlar: What's the `~` for? I've not seen that in any of our URLs before

src/app/pages/registration/confirm-api-key.component.js
```diff
@@ -0,0 +1,19 @@
+export const ConfirmApiKeyPage = {
```
andlar, line 1: Can we have a spec file for this component? I know it won't have much in it, but it fits our structure, and can `expect(the HTML to exist)`

src/app/pages/registration/components/api-key.html
```diff
@@ -1,29 +1,5 @@
 <div class="row">
   <div class="col-sm-12">
-    <div ng-if="!vm.hasKey && !vm.admin">
-      <p>You must register to use this API.</p>
-      <div class="form form-horizontal" ng-form="vm.registerForm">
-        <div class="form-group">
-          <label for="userName" class="col-sm-4 control-label">Name or Organization <span class="text-danger">*</span></label>
-          <div class="col-sm-8">
-            <input type="text" id="userName" name="userName" class="form-control" ng-model="vm.user.name" placeholder="name or organization" required></input>
-            <div class="text-danger" ng-if="vm.registerForm.userName.$touched && vm.registerForm.userName.$error.required">Name or Organization is required</div>
```
andlar, line 10: If these variables are removed from the `vm` object, and they're not used in the rest of the HTML, then they should probably be removed from whatever component/directive creates them, as well as whatever functions this deleted block of code references that are otherwise no longer used

andlar (changes requested):
I have a lot of comments. A big chunk of them are related to the migration to React, and the philosophy around that, but I want to call out that I think the functionality implemented here is awesome, and I'm really looking forward to moving along this path in the future

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -0,0 +1,84 @@
...
+  }, []);
+
+  const classes = useStyles();
+
+  return (
+    <ThemeProvider theme={ theme }>
+      <div className={ classes.grid }>
+        <div className={ classes.centeredGridItem }>
+          <Card>
+            <CardHeader title="API Email Confirmation"/>
```
kekey1: Could this read "API Key Email Confirmation" or maybe just "API Key Confirmation"?

src/app/components/api-key/api-key-confirm.jsx
kekey1: This is good information for the user and I feel like it should also be on the API Key Registration page (https://chpl.ahrqdev.org/#/resources/chpl-api). We only mention the api_key URL parameter there in passing but it's stated much more clearly in this text.

kekey1 (commented):
Just leaving a couple of comments based on using the feature. Nice job! I'll certainly approve after conflicts are resolved and Andrew's notes are addressed.

package.json
tmy1313: I understand, but I'm not sure how we want to move forward.  I'm using it to display the Alert when there is a problem confirming the API Key request:
image
I think this is a component we are going to want use in the future.  I see 3 paths forward:
1. Leave this alone for now - typically items from the "lab" are incorporated into the "core" after they have been determined to stable enough and that there is enough of need for the component.  By using this component, I feel that we are contributing to the "need" portion of the component.  https://material-ui.com/components/about-the-lab/
2. Redesign the component.  This would be fairly easy.  But as stated above - I think we are going to want the Alert component eventually.
3. Find a third-party Alert component that is supported/maintained.

Let me know how you want to proceed.

src/app/components/api-key/api-key-confirm.jsx
tmy1313: I think so, too.  But when I include the `<ul>` in the `<Typography>` tag, I get the following warning when running the tests:
image

I'm not sure what the correct structure should be here...  Any ideas?

src/app/components/api-key/api-key-confirm.jsx
tmy1313: We don't know why the confirmation failed.  Not sure how much information we should return on why the confirmation failed - not sure if we could be creating a security issue by telling the user exactly why the confirmation failed.  Kind of like telling the user that their password is incorrect - instead you just get a generic message about invalid credentials.

If you want, we can change the response from the API to provide a reason and just display that failure reason as an error message.

Let me know how you want to proceed with this.

src/app/components/api-key/api-key-registration.jsx
tmy1313: This response is only failure/success.

src/app/components/api-key/api-key-registration.test.jsx
tmy1313: The `async` keyword makes the function run asynchronously and acts as an indicator that the function should expect the possibility of the `await` keyword.  
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await

src/app/pages/resources/chpl-api/chpl-api.html
tmy1313: There is a second piece of functionality in the Angular `aiApiKey` component.  It displays a list of api keys and allows an admin/onc to revoke the api key.  So, we only replaced a portion of the component's functionality.

yarn-error.log
tmy1313: Deleted this file and added it to `.gitignore` in case it reappears.

src/app/services/network.service.js
tmy1313: Tests added.

src/app/pages/registration/registration.state.js
tmy1313: I think this was an accidental deletion.  I reverted it back to the original value.

src/app/components/api-key/api-key-confirm.jsx
andlar: I'm fine leaving it generic, and think that's generally a good idea. My original comment was that it was impossible for both of these `li` items to be true at the same time, so we shouldn't have them static here

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -65,7 +64,7 @@ function ChplApiKeyConfirm (props) {
                     Your API key is: &nbsp; <strong>{ apiKey.apiKey }</strong>
                   </Typography>
                   <Typography variant="body1" paragraph={ true }>
-                    Your API key must be included in all API calls via either a header with the name &lsquo;API-Key&lsquo; or as a URL parameter named &lsquo;api_key&lsquo;.
+                    Your API key must be included in all API calls via either a header with the name &apos;API-Key&apos; or as a URL parameter named &apos;api_key&apos;.
```
andlar: Is there a reason this isn't just using the standard `'` character? I don't think we've used this HTML entity anywhere else in our codebase

package.json
andlar: Could you get the AngularJS toaster service and use that? At least that way it's consistent with our other alert messages, and when we figure out how we're going to replace that toaster we can do them en masse.

I'm leery of item 1 as it's making a UX choice without going through our UX review process. Item 2 is basically using the toaster service temporarily, and we'll get to item 3 (maybe with the alpha code you're using now) but we'll make that decision when the team figures out how we want to handle these kinds of things

src/app/components/api-key/api-key-confirm.jsx
andlar: Will resolve this one in a bit, pending a style guide conversation

src/app/components/api-key/api-key-confirm.jsx
andlar: Are these kinds of things in any styleguide out there?

src/app/components/api-key/api-key-confirm.jsx
andlar: It looks like you can set a component on the Typography tag of "div" to allow it to nest. The default for Typography is "p", which doesn't allow block elements inside of it, according to the HTML spec

Reference: https://stackoverflow.com/a/53494821/16593

package.json
tmy1313: I still the think the card needs to have some sort of error message displayed in it.  Not sure it would make sense to display possible reasons for failure without indicating in the same "area" that an error occurred.  Is there any relationship between this an displaying errors in the action bar??  Not sure...

Maybe we should throw this over the fence for Matt to evaluate?

src/app/components/api-key/api-key-confirm.jsx
tmy1313: Let me know if this is not what you meant...

src/app/components/api-key/api-key-confirm.jsx
tmy1313: Not that I have seen.  I think that by adding these types of rules that cannot necessarily be enforced via eslint, we will have rules that we can only enforce during code reviews creating a "secret convention."  I am not a huge fan of expanding the "standard" conventions unless we really have to - I am of the opinion that AirBnB has spent more time and money developing this standard than us, and that it is pretty complete.  Also, I'm not sure that I am comfortable saying anything should be in any sort of order based on the limited React coding I have done.

src/app/pages/registration/components/api-key.directive.js
```diff
@@ -46,20 +45,6 @@
         });
     }
 
-    function register () {
-      if (vm.user.name && vm.user.email) {
-        let label = '...@' + vm.user.email.split('@')[1];
-        $analytics.eventTrack('Register For API Key', { category: 'CHPL API', label: label });
```
andlar, line 52: I don't see this analytics event getting tracked in the new component. I don't think we have a react version of this service yet, but we should be able to get the angular version and use that after a successful registration

tmy1313: Important to only use the domain - ...@domain.com

andlar (dismissed):

kekey1 (dismissed):

.eslintrc
```diff
@@ -23,46 +24,11 @@
...
-        "angular/file-name": ["off"],
-        "angular/function-type": ["off"],
-        "angular/service-name": ["off"],
-    },
     "overrides": [
     {
     "files": ["*.jsx", "*.spec.js"],
     "rules": {
-    "no-console": ["off"],
+    "max-len": ["off"],
```
andlar: This rule should be in the set of rules that were deleted. Here it only applies to the `*.jsx` and `*.spec.js` files, which is not what we want

lint-summary.js
```diff
@@ -0,0 +1,20 @@
+module.exports = results => {
+  const byRuleId = results.reduce(
+    (map, current) => {
+      current.messages.forEach(({ ruleId, line, column }) => {
+        if (!map[ruleId]) {
+          map[ruleId] = [];
+        }
+
+        const occurrence = `${current.filePath}:${line}:${column}`;
```
andlar: I see we're building up a set of messages, but later the only thing we do with it is display the count. Feel like we should either just count them, or display the messages, but this half-and-half seems odd

lint-summary.js
```diff
@@ -0,0 +1,20 @@
...
+      });
+      return map;
+    }, {}
+  );
+
+  return Object.entries(byRuleId)
+    //.map(([ruleId, occurrences]) => `${ruleId} (total: ${occurrences.length})\n${occurrences.join('\n')}`)
+    .map(([ruleId, occurrences]) => `${ruleId} (total: ${occurrences.length})`);
+    //.join('\n########################\n');
+};
```
andlar: Missing a new-line

package.json
```diff
@@ -68,6 +69,7 @@
     "ng-file-upload": "^12.2.13",
     "ng-idle": "^1.3.2",
     "ngstorage": "^0.3.11",
+    "prop-types": "^15.7.2",
```
andlar: I think this should be in `devDependencies`. We're not publishing this package as something others can consume as a library, and putting it in the `dependencies` just makes our bundled PROD slightly bigger, for no benefit

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -37,38 +37,42 @@ function ChplApiKeyConfirm (props) {
...
+                  </>
+                )
+                : (
+                  <>
+                    <Typography variant="body1" gutterBottom>
+                      Congratulations on sucessfully registering your email for use with CHPL API.  CHPL has sent you an email containing your new API key.
+                    </Typography>
+                    <Typography variant="body1" gutterBottom data-testid="api-key-display">
+                      Your API key is: &nbsp;
+                      {' '}
```
andlar: Why did this line-break happen? I'm curious about what rule was broken here, when the output is a single line of text

src/app/components/api-key/api-key-registration.jsx
```diff
@@ -33,35 +32,25 @@ const validationSchema = yup.object({
...
-    initialValues: {email: '', nameOrganization: ''},
-    validationSchema: validationSchema,
-    onSubmit: (values) => {
-      writeAnalytics(values);
-      createRequest(values);
-    },
-    validateOnChange: false,
-    validateOnMount: true,
-  });
+  let formik = {};
```
andlar, line 40: Wondering why this couldn't be a `const`. Again, curious about the rule, and why it's better to do it this way

.babelrc
```diff
@@ -5,6 +5,7 @@
             "include":["es6.promise","es7.promise.finally"],
             "corejs":"2.0",
             "debug":true
-        }]
+        }],
+        "@babel/preset-react",
```
andlar: Is this necessary? In the webpack.config.js file we're already using the `@babel/preset-react` on all of the *.js and *.jsx files

src/app/components/api-key/api-key-confirm.test.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import {
+  render, cleanup, screen, waitFor,
```
andlar, line 3: Do these need to be on a new line? Most of the other multi-imports are on a single line

src/app/components/api-key/api-key-registration.test.jsx
```diff
@@ -0,0 +1,229 @@
...
+      });
+    });
+  });
+
+  describe('when entering valid information', () => {
+    beforeEach(async () => {
+      render(<ChplApiKeyRegistration />);
+    });
+
+    it('should not have any invlid fields', async () => {
```
andlar:
```suggestion
    it('should not have any invalid fields', async () => {
```

src/app/services/network.service.js
```diff
@@ -27,888 +28,895 @@ export class NetworkService {
...
+    return this.apiPUT(`/users/${user.userId}`, user);
   }
 
-  updateVersion (versionObject) {
+  updateVersion(versionObject) {
     return this.apiPUT('/versions', versionObject);
   }
 
-  ////////////////////////////////////////////////////////////////////
+  /// /////////////////////////////////////////////////////////////////
```
andlar:
```suggestion
  /*
   * Helper functions
   */
```

src/app/components/api-key/api-key-confirm.jsx
tmy1313: I'm not sure what this was....  I reverted the code, linted and got 0 errors wrt to this.  Then ran  --fix and it made not difference.  I have reverted the change and removed the `{' '}`.

src/app/components/api-key/api-key-registration.jsx
tmy1313: I think...
* we need to define the `formik` variable near the top of the function, so that it can be referenced in other functions
* toward the bottom of the function we 'redefine' the `formik` variable using `formik = useFormik...` 
Isn't the `formik = useFormik...` redefining the variable, therefore it cannot be a `const`?

src/app/components/api-key/api-key-confirm.test.jsx
tmy1313: Not sure about this error.   This is the result when I put it all on a single line...
image

.babelrc
```diff
@@ -5,6 +5,6 @@
             "include":["es6.promise","es7.promise.finally"],
             "corejs":"2.0",
             "debug":true
-        }]
+        }],
```
andlar: I think this extra comma isn't needed, and as this isn't a *.js file, even though it looks like it, I don't think it should be here

.eslintrc
```diff
@@ -30,9 +33,8 @@
     },
     "overrides": [
     {
-    "files": ["*.jsx", "*.spec.js"],
+    "files": ["*.jsx", "*.js"],
     "rules": {
-    "no-console": ["off"],
```
andlar, line 35: Why was this rule removed? Why was the selector changed? The way this is now this entire section of overrides doesn't do anything

src/app/components/api-key/api-key-registration.jsx
```diff
@@ -0,0 +1,133 @@
...
+              fullWidth
+              variant="outlined"
+              id="name-organization"
+              name="nameOrganization"
+              label="Name or Organization"
+              required={ true }
+              value={formik.values.nameOrganization}
+              onChange={formik.handleChange}
+              onBlur={formik.handleBlur}
+              error={formik.touched.nameOrganization && Boolean(formik.errors.nameOrganization)}
```
andlar: Is the `Boolean` required? If the error= is just looking for falsy/truthy then we shouldn't need to cast it

andlar (commented):
There's a theme.js file that's committed here, as an empty file. Should there be something in it?

.eslintrc
```diff
@@ -30,9 +33,9 @@
     },
     "overrides": [
     {
-    "files": ["*.jsx", "*.spec.js"],
```
andlar, line 33: I think we're still trying to encourage the use of the angular $log service in our AngularJS files, which is why I don't think the removal of the `spec.` selector is correct. We're eventually going to ignore this rule entirely, unless React has a similar log service, on our trip towards using console.log, but I'd rather not allow it in our legacy AngularJS stuff yet

andlar (commented):
Why is there a `theme/theme.js` file getting committed in this PR with no content?

andlar (approved):

kekey1 (approved):

PR #885, opened 2021-04-13
OCD- 3649 : Update AQA test to have process listing instead of inspect list…

palakpatel311:
…ing button

[#OCD-3649]

### DIFF
--- a/e2e/pages/administration/confirm/confirm.po.js
+++ b/e2e/pages/administration/confirm/confirm.po.js
@@ -90,11 +90,11 @@ class ConfirmPage {
   }
 
   rejectListing (chplId) {
-    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
+    $('#reject-pending-listing-' + chplId).scrollAndClick();
     if (this.rejectButton.isClickable()) {
       this.rejectButton.waitAndClick();
     } else {
-      $('//input[@id="reject-pending-listing-' + chplId + '"]').waitAndClick();
+      $('#reject-pending-listing-' + chplId).waitAndClick();
       this.rejectButton.waitAndClick();
     }
   }
--- a/e2e/pages/administration/confirm/confirm.po.js
+++ b/e2e/pages/administration/confirm/confirm.po.js
@@ -81,11 +81,11 @@ class ConfirmPage {
     $('//button[@id="process-pending-listing-' + pendingListingId + '"]').scrollAndClick();
   }
 
-  rejectListingcheckbox (chplId) {
+  rejectListingCheckbox (chplId) {
     $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
   }
 
-  findListingtoReject (chplId) {
+  findListingToReject (chplId) {
     return $('//td[text()="' + chplId + '"]');
   }
 
--- a/e2e/pages/administration/confirm/confirm.po.js
+++ b/e2e/pages/administration/confirm/confirm.po.js
@@ -82,7 +82,7 @@ class ConfirmPage {
   }
 
   rejectListingcheckbox (chplId) {
-    $('#reject-pending-listing-' + chplId).scrollAndClick();
+    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
   }
 
   findListingtoReject (chplId) {
--- a/e2e/pages/administration/confirm/confirm.po.js
+++ b/e2e/pages/administration/confirm/confirm.po.js
@@ -90,11 +90,11 @@ class ConfirmPage {
   }
 
   rejectListing (chplId) {
-    $('#reject-pending-listing-' + chplId).scrollAndClick();
+    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
     if (this.rejectButton.isClickable()) {
       this.rejectButton.waitAndClick();
     } else {
-      $('#reject-pending-listing-' + chplId).waitAndClick();
+      $('//input[@id="reject-pending-listing-' + chplId + '"]').waitAndClick();
       this.rejectButton.waitAndClick();
     }
   }
--- a/e2e/pages/administration/confirm/confirm.po.js
+++ b/e2e/pages/administration/confirm/confirm.po.js
@@ -82,7 +82,7 @@ class ConfirmPage {
   }
 
   rejectListingcheckbox (chplId) {
-    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
+    $('#reject-pending-listing-' + chplId).scrollAndClick();
   }
 
   findListingtoReject (chplId) {
--- a/e2e/pages/administration/confirm/confirm.po.js
+++ b/e2e/pages/administration/confirm/confirm.po.js
@@ -3,7 +3,7 @@ const confirmElements = {
   inspectNext: '#inspect-next',
   inspectConfirm: '#inspect-confirm',
   yesConfirmation: '//button[text()="Yes"]',
-  rejectButton: '//table[@id="pending-listings-table"]/tfoot/tr/th/button',
+  rejectButton: '#reject-selected-pending-listings',
   warningCheckbox: '#acknowledge-warnings',
   confirmButton: '#inspect-confirm',
   toastContainertitle: '.ng-binding.toast-title',
@@ -17,6 +17,10 @@ class ConfirmPage {
 
   get inspectNextButton () {
     return $(confirmElements.inspectNext);
+  }
+
+  get inspectLabel () {
+    return $('#inspect-label');
   }
 
   get inspectConfirmButton () {
@@ -60,7 +64,7 @@ class ConfirmPage {
   }
 
   gotoConfirmListingPage (inspectListingId ) {
-    $('//button[@id="pending-listing-inspect-' + inspectListingId + '"]').scrollAndClick();
+    $('//button[@id="process-pending-listing-' + inspectListingId + '"]').scrollAndClick();
     this.inspectNextButton.waitAndClick();
     this.inspectNextButton.waitAndClick();
     this.inspectNextButton.waitAndClick();
@@ -77,19 +81,22 @@ class ConfirmPage {
     $('//button[@id="process-pending-listing-' + pendingListingId + '"]').scrollAndClick();
   }
 
+  rejectListingcheckbox (chplId) {
+    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
+  }
+
   findListingtoReject (chplId) {
     return $('//td[text()="' + chplId + '"]');
   }
 
   rejectListing (chplId) {
-    $('//td[text()="' + chplId + '"]/following-sibling::td[7]/input').scrollAndClick();
+    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
     if (this.rejectButton.isClickable()) {
       this.rejectButton.waitAndClick();
     } else {
-      $('//td[text()="' + chplId + '"]/following-sibling::td[7]/input').waitAndClick();
+      $('//input[@id="reject-pending-listing-' + chplId + '"]').waitAndClick();
       this.rejectButton.waitAndClick();
     }
-    this.yesConfirmation.waitAndClick();
   }
 
   confirmListing () {
--- a/e2e/pages/administration/confirm/confirm.spec.js
+++ b/e2e/pages/administration/confirm/confirm.spec.js
@@ -36,8 +36,8 @@ describe('when user is on confirm listing page', () => {
     it('should allow user to reject a file', () => {
       hooks.open('#/administration/confirm/listings');
       confirmPage.rejectListing(rejectListingId1);
-      assert.equal(toast.toastTitle.getText(), 'Success');
-      assert.isFalse(confirmPage.findListingtoReject(rejectListingId1).isDisplayed());
+      expect(toast.toastTitle.getText()).toHaveTextContaining('Success');
+      expect(confirmPage.findListingtoReject(rejectListingId1).isDisplayed()).toBe(false);
     });
   });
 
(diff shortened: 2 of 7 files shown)


e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -33,6 +36,7 @@ describe('when user is on confirm listing page', () => {
     it('should allow user to reject a file', () => {
       hooks.open('#/administration/confirm/listings');
       confirmPage.rejectListing(rejectListingId1);
+      assert.equal(toast.toastTitle.getText(), 'Success');
```
andlar: We're switching to `expect`, right?
```suggestion
      expect(toast.toastTitle.getText()).toBe('Success');
```

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -77,19 +81,22 @@ class ConfirmPage {
...
+    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
+  }
+
   findListingtoReject (chplId) {
     return $('//td[text()="' + chplId + '"]');
   }
 
   rejectListing (chplId) {
-    $('//td[text()="' + chplId + '"]/following-sibling::td[7]/input').scrollAndClick();
+    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
```
andlar, line 93: If we have an ID, can we simplify the locator?
```suggestion
    $('#reject-pending-listing-' + chplId).scrollAndClick();
```

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -77,19 +81,22 @@ class ConfirmPage {
     $('//button[@id="process-pending-listing-' + pendingListingId + '"]').scrollAndClick();
   }
 
+  rejectListingcheckbox (chplId) {
+    $('//input[@id="reject-pending-listing-' + chplId + '"]').scrollAndClick();
```
andlar, line 85: Could we do the same simplification with the other ones too?
```suggestion
    $('#reject-pending-listing-' + chplId).scrollAndClick();
```

e2e/pages/administration/confirm/confirm.po.js
```diff
@@ -60,7 +64,7 @@ class ConfirmPage {
   }
 
   gotoConfirmListingPage (inspectListingId ) {
-    $('//button[@id="pending-listing-inspect-' + inspectListingId + '"]').scrollAndClick();
+    $('//button[@id="process-pending-listing-' + inspectListingId + '"]').scrollAndClick();
```
andlar, line 67: Can this be converted too?

palakpatel311: This one is not working with only #Id.

palakpatel311: Actually I had to revert other locators from id to xpath as it was giving error of invalid or illegal selectors.

e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -4,8 +4,9 @@ import ConfirmPage from './confirm.po';
 import UploadPage from '../upload/upload.po';
 import LoginComponent from '../../../components/login/login.po';
 import Hooks from '../../../utilities/hooks';
+import ToastComponent from '../../../components/toast/toast.po';
 
-let confirmPage, hooks, loginComponent, uploadPage;
+let confirmPage, hooks, loginComponent, toast,uploadPage;
```
kekey1:
```suggestion
let confirmPage, hooks, loginComponent, toast, uploadPage;
```

e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -45,14 +53,19 @@ describe('when user is on confirm listing page', () => {
...
 
+    afterEach(() => {
+      toast.clearAllToast();
+    });
+
     it('should allow user to mass reject multiple listings', () => {
       hooks.open('#/administration/confirm/listings');
-      confirmPage.rejectCheckbox(rejectListingId1).scrollAndClick();
-      confirmPage.rejectCheckbox(rejectListingId2).scrollAndClick();
+      confirmPage.rejectListingcheckbox(rejectListingId1);
```
kekey1:
```suggestion
      confirmPage.rejectListingCheckbox(rejectListingId1);
```

e2e/pages/administration/confirm/confirm.spec.js
```diff
@@ -45,14 +53,19 @@ describe('when user is on confirm listing page', () => {
...
+      confirmPage.rejectListingcheckbox(rejectListingId2);
       confirmPage.rejectButton.waitAndClick();
-      confirmPage.yesConfirmation.waitAndClick();
-      assert.isFalse(confirmPage.findListingtoReject(rejectListingId1).isDisplayed());
-      assert.isFalse(confirmPage.findListingtoReject(rejectListingId2).isDisplayed());
+      hooks.waitForSpinnerToDisappear();
+      expect(toast.toastTitle.getText()).toBe('Success');
+      expect(confirmPage.findListingtoReject(rejectListingId1).isDisplayed()).toBe(false);
```
kekey1:
```suggestion
      expect(confirmPage.findListingToReject(rejectListingId1).isDisplayed()).toBe(false);
```

e2e/workflowTest/upload-confirm-listing-test/bypass-warnings.spec.js
```diff
@@ -32,6 +34,9 @@ describe('listing with no confirm warnings and no errors', () => {
     confirmPage.confirmListing();
     assert.isFalse(confirmPage.warningCheckbox.isDisplayed());
     confirmPage.waitForSuccessfulConfirm();
+    assert.equal(confirmPage.toastContainerTitle.getText(),'Please stand by');
```
kekey1:
```suggestion
    assert.equal(confirmPage.toastContainerTitle.getText(), 'Please stand by');
```

e2e/workflowTest/upload-confirm-listing-test/bypass-warnings.spec.js
```diff
@@ -72,6 +77,9 @@ describe('listing with warnings on confirm and no errors', () => {
     confirmPage.warningCheckbox.click();
     confirmPage.confirmListing();
     confirmPage.waitForSuccessfulConfirm();
+    assert.equal(confirmPage.toastContainerTitle.getText(),'Please stand by');
```
kekey1:
```suggestion
    assert.equal(confirmPage.toastContainerTitle.getText(), 'Please stand by');
```

kekey1 (changes requested):
No functional requests from me, but some small spacing/capitalization style things

kekey1 (dismissed):

e2e/workflowTest/upload-confirm-listing-test/bypass-warnings.spec.js
```diff
@@ -32,6 +34,9 @@ describe('listing with no confirm warnings and no errors', () => {
     confirmPage.confirmListing();
     assert.isFalse(confirmPage.warningCheckbox.isDisplayed());
     confirmPage.waitForSuccessfulConfirm();
+    assert.equal(confirmPage.toastContainerTitle.getText(), 'Please stand by');
+    toast.clearAllToast();
+    hooks.waitForSpinnerToDisappear();
     assert.equal(confirmPage.toastContainerTitle.getText(),'Update processing');
```
andlar:
```suggestion
    assert.equal(confirmPage.toastContainerTitle.getText(), 'Update processing');
```

Similar spacing stuff

e2e/workflowTest/upload-confirm-listing-test/bypass-warnings.spec.js
```diff
@@ -72,6 +77,9 @@ describe('listing with warnings on confirm and no errors', () => {
     confirmPage.warningCheckbox.click();
     confirmPage.confirmListing();
     confirmPage.waitForSuccessfulConfirm();
+    assert.equal(confirmPage.toastContainerTitle.getText(), 'Please stand by');
+    toast.clearAllToast();
+    hooks.waitForSpinnerToDisappear();
     assert.equal(confirmPage.toastContainerTitle.getText(),'Update processing');
```
andlar:
```suggestion
    assert.equal(confirmPage.toastContainerTitle.getText(), 'Update processing');
```

athomas-git (dismissed):

andlar (dismissed):

e2e/workflowTest/upload-confirm-listing-test/bypass-warnings.spec.js
```diff
@@ -37,7 +37,7 @@ describe('listing with no confirm warnings and no errors', () => {
     assert.equal(confirmPage.toastContainerTitle.getText(), 'Please stand by');
     toast.clearAllToast();
     hooks.waitForSpinnerToDisappear();
-    assert.equal(confirmPage.toastContainerTitle.getText(), 'Update processing');
+    assert.equal(confirmPage.toastContainerTitle.getText(), 'Success');
```
andlar:
```suggestion
    expect(confirmPage.toastContainerTitle.getText()).toBe('Success');
```

andlar (commented):
Is the `screenshots/.gitkeep` file needed, or will the tests automatically create that directory during the build?

wdio.conf.js
```diff
@@ -156,7 +156,7 @@ exports.config = {
         return;
       }
       var message = assertion.error.message.replace(/[^a-z0-9]/gi, '_').toLowerCase();
-      var location = `${__dirname}/test_reports/e2e/assertionError_` + message + '.png';
+      var location = `${__dirname}/test_reports/e2e/screenshot/assertionError_` + message + '.png';
```
andlar: This line uses both interpolation and concatenation. Should do only one
```suggestion
      var location = `${__dirname}/test_reports/e2e/screenshot/assertionError_${message}.png`;
```

wdio.conf.js
```diff
@@ -199,6 +199,23 @@ exports.config = {
...
+        var filename = path.join(dir, list[i]);
+        var stat = fs.statSync(filename);
+        if(filename == "allure-results" || filename.includes('wdio') || filename.includes('.png')) {
+          // pass these files- this will be download data and definition files
+        } else if(stat.isDirectory()) {
+          // rmdir recursively
+        } else {
+          // rm filename
+          fs.unlinkSync(filename);
+        }
```
andlar: Why are there three if/else sections, but only one of them has any actions? This should be a single

```
if (something is true) {
   do the unlinkSync action
}
```

wdio.conf.js
```diff
@@ -199,6 +199,22 @@ exports.config = {
       // if it doesn't exist, create it
       fs.mkdirSync(downloadDir);
     }
+    function rmdir(dir) {
+      var list = fs.readdirSync(dir);
+      for(var i = 0; i < list.length; i++) {
+        var filename = path.join(dir, list[i]);
+        var stat = fs.statSync(filename);
+        if(filename == "allure-results" || filename.includes('wdio') || filename.includes('.png') || stat.isDirectory()) {
+          // pass these files- this will be download data, definition files and screenshot, report folder
```
andlar: This block doesn't do anything, so this block shouldn't exist

palakpatel311: I wanted to have it to remember what all is there in this directory as this actually on bamboo.

tmy1313: Basically, since this is a config file we control the use/calls/params of the `rmdir` function.  We should never be calling the function with a dir that we do not want deleted.  This is basically creating a 'whitelist' of dirs that cannot be deleted, but this is not the correct construct for whitelisting.

wdio.conf.js
```diff
@@ -199,6 +199,19 @@ exports.config = {
...
+        var filename = path.join(dir, list[i]);
+        var stat = fs.statSync(filename);
+        if(!stat.isDirectory()) {
+          fs.unlinkSync(filename);
+        }
+      }
+    }
+    rmdir(downloadDir + 'screenshot');
+    rmdir(downloadDir + 'junitreport');  
+    rmdir(downloadDir);
```
andlar: I'm confused by these lines. Are these trying to remove the directories called out in the method? If so, why was the downloadDir created in the lines just above this function? Also, why not just call rmdir(downloadDir) to wipe out both of the screenshot and junitreport directories at the same time? If they're not trying to remove the directories, then the name of the function doesn't make sense

Is the whole point here to remove artifacts created by our tests? But not all the files in the test_reports/e2e directory? If that's the case, then why not put all of the things we want to remove into a single directory (or a few of them) under the test_reports/e2e directory and just remove those?

Maybe the use of the variable "downloadDir" is the problem? I mean, this directory (test_reports/e2e) by directory name doesn't have anything to do with download files, and has a bunch of files that aren't related to download files, so maybe making it the download directory isn't the best way to do it

Do we need a directory structure something like
```
+ test_reports/e2e - for general output files
+----/screenshot - for screenshots
+----/allure-results - for allure reporting
+----/junitreport - for junit reporting
+----/downloads - for temporary storage of downloadable files
```

andlar (commented):
image
I don't think this file should be here. We specifically `.gitignore` the `/test_reports` directory at the root level, so forcing a file nested inside that ignored directory to exist seems wrong

wdio.conf.js
palakpatel311: Yes so all downloads by our tests are downloaded by test_reports/e2e/.
I would like to remove those. hence, rmdir(downloadDir);
e2e/screenshot/ has screenshots from all runs so bamboo artifacts all of the previous runs screenshots as well. so i am removing previous screenshot at the start of run hence rmdir(downloadDir + 'screenshot');. This is needed as i need to remove files from this folder.
Same with junit reports, i need to remove previous reports before the start of run.

I can certainly add +----/downloads - for temporary storage of downloadable files for download files but it would just change rmdir(downloadDir + downloads);

andlar: The variable name is `downloadDir`, but it looks like that directory has things that are not downloadable files. Maybe the `test_reports/e2e` directory needs to be referenced by something like `outputDirectory` and the `downloadDir` variable could be created as `const downloadDir = outputDirectory + '/downloads';` or something similar

A separate option could be to use the `yarn e2e:clean` command, which might need to be updated, but was originally written to remove all of the generated output files before running tests. That's probably closer to the idiomatic way of building and cleaning output files

andlar (approved):

andlar (approved):

kekey1 (approved):

athomas-git (approved):

tmy1313 (approved):

PR #886, opened 2021-04-14
OCD-3634: Fix various issues with removal and update of SED

kekey1:


src/app/pages/reports/listings/listings.component.js
```diff
@@ -407,7 +407,7 @@ export const ReportsListingsComponent = {
       ];
       p = 0;
       c = 0;
-      while (pTask[p] || cTask[c]) {
+      while (pTask[p] && cTask[c]) {
```
andlar, line 410: With this `&&` comparison, if a task is added that alphabetically sorts to the end of the list of tasks, the list of Previous tasks will end before reaching the end of the list of current tasks, which short-cuts the expression and the added, last (alphabetically) task isn't mentioned in the report. A similar thing would happen with the Participants below, though in that case it might need a new participant to be created and added to a task, or to have a listing with an abundance of participants, and adding one of the "higher ID" ones to a task

image

On the bottom one, I added the task with name "Fake SED". On the top one I added 
image

andlar (changes requested):

PR #890, opened 2021-04-30
OCD-3636: Update upload beta test to look for updated errors and warnings

palakpatel311:
[#OCD-3636]

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,3 +81,7 @@ When debugging, taking screenshots can be useful. The command `browser.saveScree
 #### Automation credetials set up
 
 Copy `e2e/config/credentialsEXAMPLE.js` to the file `e2e/config/credentials.js` and set the passwords for the users in the file to be valid users
+
+#### Automation environment Urls set up
+
+Copy `e2e/config/urlsEXAMPLE.js` to the file `e2e/config/urls.js` and set the urls for each environments in the file
--- a/README.md
+++ b/README.md
@@ -63,13 +63,15 @@ On most Yarn commands the CSS Linter, JS Linter and HTML Linters will run. Webpa
 
 #### E2E testing
 
-By default, E2E tests will be executed against http://localhost:3000/. This URL can be configured at runtime with the parameter `baseUrl`. For example: `yarn e2e --baseUrl http://www.example.com/` would set all tests to run against www.example.com
+By default, E2E tests will be executed against http://localhost:3000/. This URL can be configured at runtime with the parameter `baseUrl`. For example: `ENV='Environment you want to run on' yarn e2e --baseUrl http://www.example.com/` would set all tests to run against www.example.com
 
-If a single spec file should be tested, instead of all of them, the command `yarn e2e --spec path/to/file.spec.js` will exercise only that spec file
+ENV options are 'dev', 'qa', 'stage'
 
-To change the loglevel, test could be run as `yarn e2e --l info`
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -63,13 +63,15 @@ On most Yarn commands the CSS Linter, JS Linter and HTML Linters will run. Webpa
 
 #### E2E testing
 
-By default, E2E tests will be executed against http://localhost:3000/. This URL can be configured at runtime with the parameter `baseUrl`. For example: `yarn e2e --baseUrl http://www.example.com/` would set all tests to run against www.example.com
+By default, E2E tests will be executed against http://localhost:3000/. This URL can be configured at runtime with the parameter `baseUrl`. For example: `ENV='Environment you want to run on' yarn e2e --baseUrl http://www.example.com/` would set all tests to run against www.example.com
 
-If a single spec file should be tested, instead of all of them, the command `yarn e2e --spec path/to/file.spec.js` will exercise only that spec file
+ENV options are 'dev', 'qa', 'stage'
```
andlar: Can tests be run locally? How would that happen? What's the default if this variable isn't supplied?

palakpatel311: Tests can be run even without passing ENV. It is more like, if ENV is passed, then it will be used in the spec files where it is needed, if not it will run test without considering ENV. I updated script to reflect about local ENV and added more clarity.

README.md
```diff
@@ -63,13 +63,15 @@ On most Yarn commands the CSS Linter, JS Linter and HTML Linters will run. Webpa
 
 #### E2E testing
 
-By default, E2E tests will be executed against http://localhost:3000/. This URL can be configured at runtime with the parameter `baseUrl`. For example: `yarn e2e --baseUrl http://www.example.com/` would set all tests to run against www.example.com
+By default, E2E tests will be executed against http://localhost:3000/. This URL can be configured at runtime with the parameter `baseUrl`. For example: `ENV='Environment you want to run on' yarn e2e --baseUrl http://www.example.com/` would set all tests to run against www.example.com
```
andlar: It seems weird to have two variables that are both related to the environment. Is there some way to combine these? As it stands, if you want to run tests against DEV, you'll need to both fill out the ENV variable and the baseUrl variable, and similarly for STG and QA. It seems like there should be one variable that handles both cases

palakpatel311: Yes updated.

wdio.conf.js
```diff
@@ -4,7 +4,17 @@ const fs = require('fs');
...
+let baseUrl='http://localhost:3000/';
 
+if (process.env.ENV === 'dev') {
+  baseUrl='https://chpl.ahrqdev.org/'
+}
+else if (process.env.ENV === 'qa') {
+  baseUrl='https://chpl-qa.ahrqdev.org/'
+}
+else if (process.env.ENV === 'stage') {
+  baseUrl='https://chpl-qa.ahrqdev.org/'
```
andlar: I like this functionality, but this URL is incorrect.

The other thought I had is that we generally don't want to publicize the URLs for our not PROD environments. Can we move these baseURL definitions into a gitignored file like we have for our username/password data?

palakpatel311: Good idea

andlar (dismissed):
Suggest updating the Bamboo plans to provide the `e2e/config/urls.js` file immediately, on all of the plans that execute these tests, so when the code moves forward the files are ready to be used

athomas-git (dismissed):

andlar (approved):

athomas-git (approved):

PR #891, opened 2021-05-03
OCD-3505: Add AQA tests for ONC-ACB and ONC-ATL management

ashwinimore:


e2e/pages/organizations/onc-acbs/acbs.po.js
```diff
@@ -0,0 +1,81 @@
...
+  acbEditButton: '#edit-organization',
+  acbName: '#organization-name',
+  acbWebsite: '#organization-website',
+  saveOrganizationButton: '#chpl-organization-save',
+  retireOrganization: '#organization-retired',
+  retirementDate: '#retirement-date',
+  retiredStatus: '//*[@id="chpl-organization-ONC-ACB-2"]/ui-view/div/div',
+  acbCreatetButton: '//*[contains(text(),"Create New ONC-ACB")]',
+  manageUsersPanelHeader: '//*[contains(text(),"Manage Users")]',
+  acbGeneralInformationPanel: '//*[@id="main-content"]/div/div/ui-view/chpl-onc-organizations/div[2]/div[2]/div[1]/div[2]',
```
palakpatel311: Can we try avoiding using absolute xpaths?
for example: this one has Id as chpl-organization-ONC-ACB-3, Looks like dynamic but it has id so we should work with that.

Same comments on others.

e2e/pages/organizations/onc-acbs/acbs.po.js
```diff
@@ -0,0 +1,81 @@
...
+
+class AcbPage {
+  constructor () { }
+
+  acbNameButton (acb) {
+   return $('//*[contains(text(),"' + acb + '")]');
+  }
+
+  acbsList (acb) {
+    return $('//*[@id="main-content"]/div/div/ui-view/chpl-onc-organizations/div[2]/div[1]/div/div/div[' + acb + ']/button');
```
palakpatel311: I am not 100% sure what this is doing but if this is getting list of ACBs then we have unique class name as '.organizations-side-nav' which can be used here

e2e/pages/organizations/onc-acbs/acbs.readonly.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      hooks = new Hooks();
+      address = new AddressComponent();
+      login = new LoginComponent();
+      await hooks.open('#/organizations/onc-acbs');
+    });
+
+    it('should display all ACB organizations', () => {
+	let expectedAcbs = ["ONC-ACBs", "CCHIT", "Drummond Group", "ICSA Labs", "SLI Compliance", "Surescripts LLC", "UL LLC"];
+	let i;
+	for (i=1; i<=6; i++) {
```
palakpatel311: This should have first assertion to check length of actual acbs before comparing them.

e2e/pages/organizations/onc-acbs/acbs.readonly.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    it('should allow user to edit own ACB details', () => {
+      let acb = 'UL LLC';
+      let timestamp = (new Date()).getTime();
+      let newAcbName = acb + ' - ' + timestamp;
+
+      page.acbNameButton(acb).click();
+      page.acbEditButton.click();
+      page.acbName.setValue(newAcbName);
+      page.acbWebsite.clearValue();
+      page.acbWebsite.setValue(websiteUrl);
```
palakpatel311: setValue() command does clear value first and then set it so line #63 is not needed.

e2e/pages/organizations/onc-acbs/acbs.readonly.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      page.saveOrganizationButton.click();
+      hooks.waitForSpinnerToDisappear();
+      expect(page.acbGeneralInformation.getText()).toContain(newAcbName);
+      expect(page.acbGeneralInformation.getText()).toContain(websiteUrl);
+      expect(page.acbGeneralInformation.getText()).toContain(acbAddress.address);
+      expect(page.acbGeneralInformation.getText()).toContain(acbAddress.city);
+      expect(page.acbGeneralInformation.getText()).toContain(acbAddress.state);
+      expect(page.acbGeneralInformation.getText()).toContain(acbAddress.zip);
+      expect(page.acbGeneralInformation.getText()).toContain(acbAddress.country);
+      page.acbEditButton.click();
```
palakpatel311: We shouldn't have any cleaning steps in test after assertions. These steps should be moved to after() function which in this case may need rearranging tests.

e2e/pages/organizations/onc-acbs/acbs.readonly.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      page.acbName.setValue(acb);
+      page.saveOrganizationButton.click();
+    });
+
+    it('should require ACB Name, website and address on edit', () => {
```
palakpatel311: We are checking here error messages displayed on empty fields on required field so this test could be rephrased something like 'should show correct error message when required fields are empty'

e2e/pages/organizations/onc-acbs/acbs.readonly.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      let timestamp = (new Date()).getTime();
+      let newAcbName = acb + ' - ' + timestamp;
+
+      page.acbNameButton(acb).click();
+      page.acbEditButton.click();
+      page.acbName.setValue(newAcbName);
+      page.acbWebsite.clearValue();
+      page.acbWebsite.setValue(websiteUrl);
+      address.set(acbAddress);
+      page.saveOrganizationButton.click();
```
palakpatel311: This tests is saving it so it should be in non readonly

e2e/pages/organizations/onc-acbs/acbs.po.js
```diff
@@ -0,0 +1,81 @@
+const elements = {
+  acbEditButton: '#edit-organization',
+  acbName: '#organization-name',
+  acbWebsite: '#organization-website',
+  saveOrganizationButton: '#chpl-organization-save',
+  retireOrganization: '#organization-retired',
+  retirementDate: '#retirement-date',
+  retiredStatus: '//*[@id="chpl-organization-ONC-ACB-2"]/ui-view/div/div',
+  acbCreatetButton: '//*[contains(text(),"Create New ONC-ACB")]',
```
palakpatel311: Typo. Extra t between create and button

e2e/pages/organizations/onc-atls/atls.po.js
```diff
@@ -0,0 +1,83 @@
+const elements = {
+  atlEditButton: '#edit-organization',
+  atlName: '#organization-name',
+  atlWebsite: '#organization-website',
+  saveOrganizationButton: '#chpl-organization-save',
+  retireOrganization: '#organization-retired',
+  retirementDate: '#retirement-date',
+  retiredStatus: '//*[@id="chpl-organization-ONC-ATL-2"]/ui-view/div',
+  atlCreatetButton: '//*[contains(text(),"Create New ONC-ATL")]',
```
palakpatel311: Same typo

e2e/pages/organizations/onc-atls/atls.po.js
```diff
@@ -0,0 +1,83 @@
...
+  manageUsersPanelHeader: '//*[contains(text(),"Manage Users")]',
+  ulButton: '//*[contains(text(),"UL LLC")]',
```
palakpatel311: I see same button locator for cchitButton and drummondButton. This can be generalize by passing acb to a method.

e2e/pages/organizations/onc-atls/atls.readonly.spec.js
```diff
@@ -0,0 +1,131 @@
...
+      }
+    });
+
+    describe('when impersonating as an ATL', () => {
+      beforeEach(() => {
+      login.logIn('onc');
+      login.logoutButton.waitForDisplayed();
+      hooks.open('#/users');
+      page.impersonateUser.scrollIntoView({block: 'center', inline: 'center'});
+      page.impersonateUser.click();
```
palakpatel311: I see we dont have aqa atl set up right now. I would like to see if we can create one rather than impersonating atl user. Leaving it to @andlar

e2e/pages/organizations/onc-atls/atls.readonly.spec.js
```diff
@@ -0,0 +1,131 @@
...
+      beforeEach(() => {
+        login.logIn('onc');
+        login.logoutButton.waitForDisplayed();
+      });
+
+      afterEach(() => {
+        login.logOut();
+      });
+
+    it('should allow user to unretire and retire existing ATL', () => {
```
palakpatel311: Can we separate these as in ACBs?
This test should be in non readonly

e2e/pages/organizations/onc-atls/atls.readonly.spec.js
```diff
@@ -0,0 +1,131 @@
...
+    it('should allow user to edit own ATL details', () => {
+	  let atl = 'UL LLC';
+      let timestamp = (new Date()).getTime();
+      let newAtlName = atl + ' - ' + timestamp;
+	  page.atlNameButton(atl).click();
+      page.atlEditButton.click();
+      page.atlName.setValue(newAtlName);
+      page.atlWebsite.setValue(websiteUrl);
+      address.set(atlAddress);
+      page.saveOrganizationButton.click();
```
palakpatel311: This test should be in non readonly

palakpatel311 (commented):
Great work on first time doing these UI Tests in this frame work. :)

I have reviewed code related to ACBs edit and some of the comments applies to ATLs edit as well. So not repeating my self there.

e2e/pages/organizations/onc-acbs/acbs.po.js
ashwinimore: I _did_ think about this and thought of parametrizing, then thought it's ok just for this one as I did not see we'd want to change test data on this one as we would in other cases (also thought if/when we do move to react on this one, request better locators then) and that we don't really use acb ids (vs. name & code). I'll update it so it uses acb id   
As for other xpaths, the classname wasn't unique on those so I went ahead and used xpath, i'll look for other possibilities

e2e/pages/organizations/onc-atls/atls.readonly.spec.js
ashwinimore: so....i wrote this one also exactly similar to ATL one (unretire then retire), as discussed with the team last week and put it in readonly but there's slight difference with how ATL save is processed vs. ACBs (e.g. toaster/no toaster). This test didn't work for ACB same way it did here. I wanted to test this out on AQA plan- to see which one does better and then make an update but couldn't do that just yet. I read that specs within a describe block _should_ run in sequence but wanted to test it out first.. more on this after findings (thinking of using feature review for discussion)

e2e/pages/organizations/onc-atls/atls.readonly.spec.js
ashwinimore: I moved it here because i was changing the ATL name (same for ACB) back to original name and although address is changed, it doesn't have impact on anything else. I don't feel strongly about keeping it here, I originally had it in non readonly and moved it after the discussion last week. If no one else has different opinion, I'll move it to non readonly

e2e/pages/organizations/onc-atls/atls.readonly.spec.js
ashwinimore: A while ago, we had a brief discussion during team meeting on adding user accounts for aqa for the ones that we don't have it already and I recall the consensus was that we could use impersonation for those. I added this based on that and would like to keep it until we have the accounts added(if we decide to do so). I like that it covers impersonation test as well

e2e/pages/organizations/onc-atls/atls.po.js
ashwinimore: already implemented using atlNameButton (atl) method below. This was missed in cleanup(leftover from initial write up before parameterizing), will be removed in next commit

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    hooks = new Hooks();
+    address = new AddressComponent();
+    login = new LoginComponent();
+    await hooks.open('#/organizations/onc-acbs');
+  });
+
+  it('should display all ACB organizations', () => {
+    const expectedAcbs = ['ONC-ACBs', 'CCHIT', 'Drummond Group', 'ICSA Labs', 'SLI Compliance', 'Surescripts LLC', 'UL LLC'];
+    let i;
+    for (i = 1; i <= 6; i += 1) {
```
andlar: What this test will miss is if there is an ACB added to the system that shows up in this list after UL LLC. The loop only goes to six, but if there are seven ACBs we won't notice that

e2e/pages/organizations/onc-acbs-atls/acbs.spec.js
```diff
@@ -0,0 +1,117 @@
...
+      expect(page.newOrganizationGeneralInfo.getText()).toContain(newAcbName);
+      expect(page.newOrganizationGeneralInfo.getText()).toContain(websiteUrl);
+      expect(page.newOrganizationGeneralInfo.getText()).toContain(acbAddress.address);
+      expect(page.newOrganizationGeneralInfo.getText()).toContain(acbAddress.city);
+      expect(page.newOrganizationGeneralInfo.getText()).toContain(acbAddress.state);
+      expect(page.newOrganizationGeneralInfo.getText()).toContain(acbAddress.zip);
+      expect(page.newOrganizationGeneralInfo.getText()).toContain(acbAddress.country);
+    });
+
+    it('should allow user to unretire and retire existing ACB', () => {
```
andlar: I thought we were going to put this test in the `readonly` file?

e2e/pages/organizations/onc-acbs-atls/atls.spec.js
```diff
@@ -0,0 +1,126 @@
...
+      expect(page.generalInformation(organizationType, atlId).getText()).toContain(newAtlName);
+      expect(page.generalInformation(organizationType, atlId).getText()).toContain(websiteUrl);
+      expect(page.generalInformation(organizationType, atlId).getText()).toContain(atlAddress.address);
+      expect(page.generalInformation(organizationType, atlId).getText()).toContain(atlAddress.city);
+      expect(page.generalInformation(organizationType, atlId).getText()).toContain(atlAddress.state);
+      expect(page.generalInformation(organizationType, atlId).getText()).toContain(atlAddress.zip);
+      expect(page.generalInformation(organizationType, atlId).getText()).toContain(atlAddress.country);
+    });
+  });
+  describe('when logged in as ONC', () => {
```
andlar:
```suggestion
  });
  
  describe('when logged in as ONC', () => {
```

I think we try to separate `describe` and `it` blocks with blank lines, though I haven't found an eslint rule that enforces that

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
```diff
@@ -0,0 +1,88 @@
...
+  organizationEditButton: '#edit-organization',
+  organizationName: '#organization-name',
+  organizationWebsite: '#organization-website',
+  saveOrganizationButton: '#chpl-organization-save',
+  organizationList: '.organizations-side-nav',
+  retireOrganization: '#organization-retired',
+  retirementDate: '#retirement-date',
+  manageUsersPanelHeader: '//*[contains(text(),"Manage Users")]',
+  errorMessage: '.text-danger.ng-scope',
+  newOrganizationGeneralInfo: '//*[@id="main-content"]/div/div/ui-view/chpl-onc-organizations/div[2]/div[2]/div[1]/div[2]',
```
andlar:
```suggestion
  newOrganizationGeneralInfo: '//chpl-onc-organizations/div[2]/div[2]/div[1]/div[2]',
```

You can probably start from the `chpl-onc-organizations` html element, as on these pages there should only ever be one of them, and anything to cut down on the number of nested elements in an xpath is good with me

Can do something similar with the next two lines, starting with `chpl-address` and `chpl-users` respectively

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
```diff
@@ -0,0 +1,88 @@
...
+  get addressLine1errorMessage() {
+    return $(elements.addressLine1OnEdit);
+  }
+
+  impersonateUser(userId) {
+    return $(`//*[@id="user-component-impersonate-${userId}"]`);
+  }
+}
+
+export default OrganizationPage;
```
andlar: The filename is plural, the class is singular. I don't know if this page toggles between singular and plural, but if the PO data is only one organization, then I'd suggest going with singular in both places

andlar (changes requested):

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,77 @@
+import OrganizationPage from './organizations.po';
+import Hooks from '../../../utilities/hooks';
+import AddressComponent from '../../../components/address/address.po';
+import LoginComponent from '../../../components/login/login.po';
+
+let address; let hooks; let login; let
```
palakpatel311: These can be combined as let address,hooks,......

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    it('should show errors for missing inputs in required fields ACB Name, website and address on ACB edit', () => {
+      const acb = 'UL LLC';
+      const newAcbName = `${acb} - ${timestamp}`;
+
+      page.organizationNameButton(acb).click();
+      page.organizationEditButton.click();
+      page.organizationName.clearValue();
+      expect(page.errorMessage.getText()).toBe('Field is required');
```
palakpatel311, line 85: More appropriate is to check required fields when they click on save.

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,77 @@
...
+      page.organizationWebsite.setValue(websiteUrl);
+      address.editAddress.clearValue();
+      expect(page.addressLine1errorMessage.getText()).toBe('Field is required');
+    });
+
+    it('should not present the option to edit ACB details other than own organization', () => {
+      const acb = 'Drummond Group';
+      page.organizationNameButton(acb).click();
+      expect(page.organizationEditButton.isDisplayed()).toBe(false);
```
palakpatel311, line 68: expect(page.organizationEditButton.isDisplayed()**).toBeFalse;**

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
```diff
@@ -0,0 +1,88 @@
...
+
+  get errorMessage() {
+    return $(elements.errorMessage);
+  }
+
+  get addressLine1errorMessage() {
+    return $(elements.addressLine1OnEdit);
+  }
+
+  impersonateUser(userId) {
```
palakpatel311: This doesnt belong to organizations page. It should be under user management po.

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
palakpatel311: only $('chpl-onc-organization') as locator would work.

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
```diff
@@ -0,0 +1,88 @@
...
+  organizationName: '#organization-name',
+  organizationWebsite: '#organization-website',
+  saveOrganizationButton: '#chpl-organization-save',
+  organizationList: '.organizations-side-nav',
+  retireOrganization: '#organization-retired',
+  retirementDate: '#retirement-date',
+  manageUsersPanelHeader: '//*[contains(text(),"Manage Users")]',
+  errorMessage: '.text-danger.ng-scope',
+  newOrganizationGeneralInfo: '//*[@id="main-content"]/div/div/ui-view/chpl-onc-organizations/div[2]/div[2]/div[1]/div[2]',
+  addressLine1OnEdit: '//*[@id="chpl-organization-edit"]/div/chpl-address/span/span/div[1]/div[1]/div',
```
palakpatel311: This locator looks like being used for error message for address and I do see separate error message locator for these messages.

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
```diff
@@ -0,0 +1,88 @@
...
+  manageUsersPanelBody: '//*[@id="main-content"]/div/div/ui-view/chpl-onc-organizations/div[2]/div[2]/div[2]/div[2]/chpl-users/span/div[2]',
```
palakpatel311: Trying to avoid absolute xpath as much as we can. $('chpl-users') locator would work here.

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
ashwinimore: that's how I had it before doing the lint check, this change was a suggested fix to meet eslint rule. If we have the rule active, then I understand this is expected per practice. If we want to use 'combined' way then I think the rule should be relaxed. I think we should discuss this as a team to have some clarity (I know I need it)

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
ashwinimore: addressed all comments in latest commit

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
ashwinimore: updated in latest commit

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,128 @@
...
+import AddressComponent from '../../../components/address/address.po';
+import LoginComponent from '../../../components/login/login.po';
+
+let address;
+let hooks;
+let login;
+let page;
+
+describe('the ONC-ACB Management page', () => {
+  const timestamp = (new Date()).getTime();
```
andlar: No need to change this, but I discovered this recently, and thought it looked much better to do the same thing:

```suggestion
  const timestamp = Date.now();
```

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
andlar: That might work, but there are some places where clicking the button is impossible if the form is invalid, which means that trying to click the button will fail, but trying to click it is required before the errors appear. Adding that interaction makes the test harder to implement; it could be done, but I'm not sure it's worth the effort

e2e/pages/organizations/onc-acbs-atls/organization.po.js
```diff
@@ -0,0 +1,86 @@
...
+  organizationNameButton(organizationName) {
+    return $(`//*[contains(text(),"${organizationName}")]`);
+  }
+
+  get organizationList() {
+    return $(elements.organizationList);
+  }
+
+  organizationListCount() {
+    return $(elements.organizationList).length;
```
andlar: You already define the `organizationList` above, so here it could leverage that definition in such a way that if the locator for the organization list changes, this count wouldn't have to

I think something like this should work:

```suggestion
    return this.organizationList.length;
```

e2e/pages/organizations/onc-acbs-atls/organization.po.js
```diff
@@ -0,0 +1,86 @@
...
+  get manageUsersPanel() {
+    return $('chpl-users');
+  }
+
+  get newOrganizationGeneralInfo() {
+    return $('chpl-onc-organization');
+  }
+
+  retiredStatus(organizationType, organizationId) {
+    return $(`//*[@id="chpl-organization-ONC-${organizationType}-${organizationId}"]/ui-view/div/div`);
```
andlar: I find locators that aren't xpath easier to understand. I don't think it's required to change it, but something like this should work too, and I think is easier to read:

```suggestion
    return $(`#chpl-organization-ONC-${organizationType}-${organizationId}/ui-view/div/div`);
```

e2e/pages/users/user.po.js
```diff
@@ -9,44 +9,47 @@ const userElements = {
...
   }
 
-  userInformation (name) {
-    return $('//h2[text()="' + name + '"]/parent::div/parent::div/following-sibling::div');
+  userInformation(name) {
+    return $(`//h2[text()="${name}"]/parent::div/parent::div/following-sibling::div`);
   }
 
+  impersonateUser(userId) {
+    return $(`//*[@id="user-component-impersonate-${userId}"]`);
```
andlar: Similar thought here. Also not required:

```suggestion
    return $(`#user-component-impersonate-${userId}`);
```

andlar (dismissed):
I have some suggestions, but none of them are required. I'm okay with this going forward as is

e2e/pages/organizations/onc-acbs/acbs.readonly.spec.js
palakpatel311: This is not addressed. We should check acbCount and expectedACBs length, if its same then only enter for loop

e2e/pages/organizations/onc-acbs-atls/organizations.po.js
palakpatel311: This seems to be missed.

palakpatel311:
There was one comment seems to be missed. I don't want to hold up this PR for that. I can wait for that change to review in final code review.

palakpatel311 (dismissed):

palakpatel311 (dismissed):

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,128 @@
...
+    address = new AddressComponent();
+    login = new LoginComponent();
+    await hooks.open('#/organizations/onc-acbs');
+  });
+
+  it('should display all ACB organizations', () => {
+    const acbCount = page.organizationListCount();
+    const expectedAcbs = ['ONC-ACBs', 'CCHIT', 'Drummond Group', 'ICSA Labs', 'SLI Compliance', 'Surescripts LLC', 'UL LLC'];
+    let i;
+    for (i = 1; i <= acbCount; i += 1) {
```
andlar: I think I figured out what is bothering me about this. What would happen if `UL LLC` was deleted? The `acbCount` variable would be 6, so the loop would check the array of expectedAcbs but for the values of `i` from 1 to 5, but that would not look for the expected `UL LLC`

I know that's an unusual situation to get into, but is that also what we're trying to do with our AQA stuff? Find and test edge cases that we wouldn't normally see?

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,128 @@
...
+      expect(page.organizationList.getText()).toContain(expectedAcbs[i]);
+    }
+  });
+
+  describe('when logged in as an ACB', () => {
```
andlar: If we're logging in as a specific ACB, I think it would make sense to call out the ACB we log in as during the describe block. Probably would suggest we do the same thing for the `it` blocks on lines 52, 60, and 66. That way the tests would read something like

```
when logged in as UL
  it should display registered users under UL
  it should not present the option to edit ACB details for drummond
  it should not display registered users under drummond
```

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
palakpatel311: Now noticing there is 'ONC-ACBs' in the expected ACBs??

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,128 @@
...
+    address = new AddressComponent();
+    login = new LoginComponent();
+    await hooks.open('#/organizations/onc-acbs');
+  });
+
+  it('should display all ACB organizations', () => {
+    const expectedAcbs = ['CCHIT', 'Drummond Group', 'ICSA Labs', 'SLI Compliance', 'Surescripts LLC', 'UL LLC'];
+    const actualAcbs = page.organizationList.getText();
+    let i;
+    for (i = 1; i <= expectedAcbs.length; i += 1) {
```
andlar: This now checks to see if all of the expected ACBs are found, but it won't give an error if there are any new ACBs that have been added

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -30,13 +30,15 @@ describe('the ONC-ACB Management page', () => {
...
-      expect(actualAcbs).toContain(expectedAcbs[i - 1]);
-    }
+  it('should have at least 6 ACB organizations', () => {
+    expect(page.organizationListCount()).toBeGreaterThanOrEqual(6);
+  });
+
+  it('should display ACB organizations in alphabetical order', () => {
+    const acbList = page.organizationList.getText();
+    const currentAcbList = acbList.split();
+    const sortedAcbList = currentAcbList.sort();
```
andlar: The Javascript array sort method sorts the items in the array in place, modifying the original array, so this comparison will _always_ succeed

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  it('should display ACB organizations in alphabetical order', () => {
+    const acbCount = page.organizationListCount();
+    const actualList = [];
+    let i;
+    for (i = 1; i <= acbCount; i += 1) {
+      actualList.push(page.organizationListValue(i - 1).getText());
+    }
+    const expectedList = actualList.sort();
```
andlar: The change didn't actually change anything. Calling `actualList.sort()` sorts the actual list. Setting the expected list to the sorted actual list will always mean that the two lists are equal. I can see two different ways of doing this. Pseudocode:

1)

```
// using an array to walk through the actual list, and string comparison for ordering check
actualList = get list from PO
for (i = 2; i < acbCount; i++) {
expect(actualList[i - 1]) to be less than actualList[i])
}
```

2)
```
// get two lists
actualList = get list from po
expectedList = page.organizationList.sort()
expect(actualList).toEqual(expectedList) // note, using toBe probably won't work
```

ashwinimore: I thought i was doing option 2 except that I didn't realize that the array was modified at sort (thought that the value of actualList was not altered). I like your first option better and going to go with that. Thank you!

e2e/pages/organizations/onc-acbs-atls/acbs.readonly.spec.js
```diff
@@ -0,0 +1,135 @@
...
+
+  it('should have at least 6 ACB organizations', () => {
+    expect(page.organizationListCount()).toBeGreaterThanOrEqual(6);
+  });
+
+  it('should display ACB organizations in alphabetical order', () => {
+    const acbCount = page.organizationListCount();
+    const actualList = [];
+    let i;
+    for (i = 1; i <= acbCount; i += 1) {
```
andlar: There's no reason to loop through this list twice. Either get the full list of ACBs using a page object method and then loop, or, in the one loop where you're adding them to this array (and why is that happening this way?) you could check the alphabetical order at the same time

andlar (approved):

palakpatel311 (approved):

PR #894, opened 2021-05-06
OCD-3643: Display more Listing data in Complaints download file

andlar:


src/app/components/surveillance/complaints/complaints.html
```diff
@@ -107,11 +107,11 @@
...
-                csv-column-order="['id', 'acbName', 'complainantTypeName', 'complainantTypeOther', 'oncComplaintId', 'acbComplaintId', 'csvReceivedDate', 'summary', 'actions', 'complainantContacted', 'developerContacted', 'oncAtlContacted', 'flagForOncReview', 'csvClosedDate', 'csvListings', 'csvSurveillances', 'csvCriteria']">
-          <i class="fa fa-cloud-download"></i> Download {{ $ctrl.displayedComplaints.length }} displayed result{{ $ctrl.displayedComplaints.length !== 1 ? 's' : '' }}
+                ng-csv="$ctrl.csvComplaints" filename="{{ $ctrl.filename }}" add-bom="true"
+                ng-disabled="!$ctrl.csvComplaints"
+                csv-header="['complaint_id', 'ONC-ACB', 'Complainant Type Name', 'complainant_type_other', 'onc_complaint_id', 'acb_complaint_id', 'received_date', 'summary', 'actions', 'complainant_contacted', 'developer_contacted', 'onc_atl_contacted', 'informed_onc_per_170.523(s)', 'closed_date', 'developer_name', 'product_name', 'version', 'associated_listings', 'associated_surveillance', 'associated_criteria']"
```
kekey1, line 112: Do these headers have to be written like this to match something in the data? Or could we make them capitalized and use spaces instead of underscores?

kekey1 (dismissed):

src/app/components/surveillance/complaints/complaints.html
andlar: They could be written however we want, but the only changes here were adding the *_name columns, by the name they're referred to in the requirements. The original column names haven't changed, and though I don't recall if they were required to be that way or not, no one has asked for them to change

tmy1313 (dismissed):

PR #896, opened 2021-05-07
OCD- 3534 AQA test for manage surveillance reporting tab

palakpatel311:


e2e/pages/surveillance/manage/manage.spec.js
```diff
@@ -0,0 +1,61 @@
...
+describe('on manage surveillance page', () => {
+  beforeEach(async () => {
+    loginComponent = new LoginComponent();
+    manage = new ManagePage();
+    hooks = new Hooks();
+    toast = new ToastComponent();
+    await hooks.open('#/surveillance/manage');
+  });
+
+  describe('When admin user is logged in', () => {
```
athomas-git: Can we use ONC user instead of ADMIN if this is the only positive scenario ?

andlar: I was thinking that it would probably be better to have our automated tests use the ROLE_ONC users in general, and only use the ROLE_ADMIN ones for things the ROLE_ONC ones don't do. My thought is that the ROLE_ADMIN users are just us, and if something goes wrong for us it's not good, but it's not as bad as if something goes wrong for ROLE_ONC (and eventually ROLE_ONC_STAFF)

andlar (approved):

athomas-git (approved):

PR #900, opened 2021-05-12
OCD 3495:  AQA tests for collections page

palakpatel311:


e2e/pages/collections/api-documentation/api-documentation.readonly.spec.js
```diff
@@ -0,0 +1,135 @@
...
+      page.downloadApiDocButton.click();
+    });
+
+    it('should download a file', () => {
+      const fileName = 'APIDocData-20193112.xlsx';
+      const filePath = path.join(global.downloadDir, fileName);
+      browser.waitForFileExists(filePath, config.timeout);
+      expect(fs.existsSync(filePath)).toBe(true);
+      const stat = fs.statSync(filePath);
+      console.log(stat.size / 1000);
```
andlar: Probably not required

e2e/pages/collections/collections.po.js
```diff
@@ -1,18 +1,96 @@
 const elements = {
   bodyText: 'ai-body-text',
   listingTable: 'table',
+  searchBar: '#generalFilter',
+  pagination: '.pagination--results-found',
+  clearFilters: '//a[text()="Clear Filters"]',
+  dateFilter: '#filter-button-decertificationDate',
```
andlar: I don't think these three date fields are in every collections page, so should probably be moved to the derived pages that use them

e2e/pages/collections/corrective-action/corrective-action.readonly.spec.js
```diff
@@ -1,31 +1,147 @@
 import CorrectiveActionPage from './corrective-action.po';
 import Hooks from '../../../utilities/hooks';
 
-let hooks, page;
+let hooks; let
+  page;
```
andlar:
```suggestion
let hooks;
let page;
```

I really don't like the way the auto-fix runs these into new lines like this, but don't think I have a way for the linting to enforce it

e2e/pages/collections/corrective-action/corrective-action.readonly.spec.js
```diff
@@ -1,31 +1,147 @@
...
+    describe('using certification status filter to select withdrawn by developer', () => {
+      beforeEach(() => {
+        page.selectFilter('certificationStatus', 'Withdrawn_by_Developer');
+        page.waitForUpdatedListingResultsCount();
+      });
+
+      it('should filter listing results', () => {
+        countAfter = page.listingTotalCount();
+        expect(countAfter).toBeGreaterThan(countBefore);
+      });
```
andlar: When the describe block only has one `it` in it, we probably don't need a `beforeEach`. The code in the before each could just go into the it. It's not a required change, but it would cut down on boilerplate stuff

e2e/pages/collections/developers/developers.readonly.spec.js
```diff
@@ -0,0 +1,81 @@
+import DevelopersPage from './developers.po';
+import Hooks from '../../../utilities/hooks';
+
+let hooks; let page;
+const DEVELOPER_COL_IDX = 1;
+const developerName = 'Rabbit';
```
andlar: These first two make sense to be "globally" defined, as they'll be used in every test. I don't think the developerName will be though, and could be moved down into just the blocks that actually use it

wdio.conf.js
```diff
@@ -297,8 +297,9 @@ exports.config = {
   /**
    * Function to be executed after a test (in Mocha/Jasmine).
    */
-  // afterTest: function(test, context, { error, result, duration, passed, retries }) {
-  // },
+  afterTest: function(test, context, { error, result, duration, passed, retries }) {
```
andlar: My guess is that the commented out stuff is here to show folks using this tool what the function parameters look like. That's useful, and I'd like to keep that around, but as this is written none of those arguments are used in the function. This could be changed to the below, to keep the method signature, but only use the arguments we actually care about

```suggestion
  // afterTest: function(test, context, { error, result, duration, passed, retries }) {
  afterTest: function() {
```

e2e/pages/collections/collections.po.js
palakpatel311: It is used in 3 collections so I think it should be in this PO.

andlar: If these fields need to be in three other POs, then there could be an intermediate page that those three inherit from. By putting these fields in this page, that means all of the other POs that inherit from this one members that are literally invalid on those pages, which isn't right

e2e/pages/collections/corrective-action/corrective-action.po.js
```diff
@@ -1,7 +1,7 @@
 import CollectionsPage from '../collections.po';
 
 class CorrectiveActionPage extends CollectionsPage {
-  constructor () {
+  constructor() {
```
andlar: In this PO, does it provide any value that is not already in the "CollectionsPage"? If it doesn't, then the Corrective Action page could just instanciate one of the POs directly derived from the base CollectionsPage. There's no need for this PO if it doesn't provide any additional value

e2e/pages/collections/collections.po.js
palakpatel311: I do see inheritance problem but I am not so sure about this as what name of this intermediate page will be? This is POM framework so all page objects resides on the page gets stores by it's page name for example collection page.
I can definitely move out date filters to those collections which uses it which would resolve this issue. Let me know your thoughts. I will wait to make any changes until then.

andlar: I'm not sure what the problem is now. I think you could either move the date fields to an intermediate class, and work out a name for it, or put the fields on the individual collection pages. Either way is probably fine, as in both of those options the date fields aren't on the "main / root" page, which is what we want

andlar (dismissed):

athomas-git (dismissed):

andlar (approved):

athomas-git (approved):

PR #901, opened 2021-05-13
OCD-3383: Add developer to version activity report

kekey1:


src/app/pages/reports/versions/versions.html
```diff
@@ -59,6 +59,7 @@ <h3>Version</h3>
             </td>
           </tr>
           <tr>
+            <th class="search-header" st-sort="developerName" st-skip-natural="true">Developer</th>
```
andlar, line 62: When adding a column to the table, the `<thead>` and `<tfoot>` sections need their `colspan` values to increase from the 4 they're set at to 5, so the header/footers span across the entire table

andlar (changes requested):

andlar (approved):

tmy1313 (approved):

PR #908, opened 2021-05-24
OCD-3546 - Handle routing errors/404 stuff

andlar:


ashwinimore (dismissed):

e2e/components/login/login.po.js
```diff
@@ -1,60 +1,62 @@
...
+      this.toggleLoginComponent();
+      un = $(this.elements.username);
+      pw = $(this.elements.password);
+      btn = $(this.elements.login);
+    }
+    un.addValue(credentials[user].email || credentials[user].username);
+    pw.addValue(credentials[user].password);
+    btn.scrollAndClick();
+    if (!($(this.elements.component).isDisplayed())) {
+      this.toggleLoginComponent();
```
palakpatel311, line 43: After login, why do we open login component again?
Because when this is used in files needed for login as below:
login.logIn('admin');    
login.waitToBeLoggedIn();

So code is 

**if (!($(this.elements.component).isDisplayed())) 
{      this.toggleLoginComponent();}

And then again:
if (!($(this.elements.logout).isDisplayed())) 
{      this.toggleLoginComponent();    }**    
$(this.elements.logout).waitForDisplayed()

Isn't Highlighted code same?

e2e/components/login/login.readonly.spec.js
```diff
@@ -1,76 +1,83 @@
...
+    });
 
-  it('should be able to log in as ONC', () => {
-    component.logIn('onc');
-    component.logoutButton.waitForDisplayed();
-    expect(component.toggleLoginComponent.getText()).toBe('AQA ONC');
-  });
+    it('should be able to log in as drummond ACB', () => {
+      component.logIn('drummond');
+      component.waitToBeLoggedIn();
```
palakpatel311: Next step here is to check logged in user's name for that we dont have to open login component.

e2e/components/login/login.po.js
andlar: We don't open it again. If we're using the legacy component, the first call to toggle opens it, and this one closes the window. That needs to happen because if we don't, then any other call that tries to use the legacy login component needs to figure out if that is open or not, and that's hard to do. By closing it we know that we always start in a state of "closed"

e2e/components/login/login.readonly.spec.js
andlar: ~Correct, but waiting to be logged in doesn't require opening the login component either~

This was wrong; the check to see if you're logged in requires the component to be open. Further discussion in a lower comment

e2e/components/login/login.po.js
palakpatel311: But at the end of these calls, Login toggle is OPEN.
1. When we enter username, password and click on login button- login toggle is open
2. with this check if (!($(this.elements.component).isDisplayed())) - it will close
3. with this check if (!($(this.elements.logout).isDisplayed()))- it will open again.

e2e/components/login/login.readonly.spec.js
palakpatel311: I think this is related to logIn method conversation. Either we have logIn() method to keep login component open and close with waitToBeLoggedIn() or we have logIn() method to close login component and no need for waitToBeLoggedIn()

andlar: `waitToBeLoggedIn` is using the logic that was originally in this test, where the check was "is the log out button displayed". For that to work, and for the `waitToBeLoggedIn` function to be as _functional_ and side-effect free as possible, the method needs to assume it's starting from any random page, which means it needs to first open the toggle component, and then look for the "log out" button

The "log in" method needs to be able to work with both the new and old login components, so the "wait to be logged in" one does as well

e2e/components/login/login.po.js
andlar: At the end of which calls is the toggleable component open? Line 49? It's open when it gets to that line, so line 49 closes it, which restores the page to the state it was in, with respect to the login component, except that the user is now logged in. The check on line 48 doesn't _change_ anything; it just figures out if this page has the new login component...

As I type that, I recognize it doesn't actually do that, as if we were on the page with the new component, we wouldn't be at this point, so line 48 will always return true, which we don't want

palakpatel311: No line 47. Our login behavior is that it keeps toggleable component open after login. and as you mentioned line 48 checks new login

e2e/pages/administration/jobs/scheduled/scheduled.readonly.spec.js
```diff
@@ -1,35 +1,37 @@
...
+    loginComponent.waitToBeLoggedIn();
     hooks.waitForSpinnerToDisappear();
   });
 
   it('should see the right set of scheduled jobs', () => {
-    const expected = new Set(['All Broken Surveillance Rules Report','Developer Access Report','Inherited Certification Status Errors Report','Listing Validation Email Report','ONC-ACB Questionable URL Report','Overnight Broken Surveillance Rules Report','Pending "Change Request" Report','Questionable Activity Report','Questionable URL Report','Real World Testing Email Report','Summary Statistics Email','Trigger Developer Ban Notification']);
+    const expected = new Set(['All Broken Surveillance Rules Report', 'Cures Statistics Email', 'Developer Access Report', 'Inherited Certification Status Errors Report', 'Listing Validation Email Report', 'ONC-ACB Questionable URL Report', 'Overnight Broken Surveillance Rules Report', 'Pending "Change Request" Report', 'Questionable Activity Report', 'Questionable URL Report', 'Real World Testing Email Report', 'Summary Statistics Email', 'Trigger Developer Ban Notification']);
     browser.waitUntil(() => scheduled.scheduledJobTable.isDisplayed());
```
palakpatel311: This test is failing frequently when assert runs before table gets filled in. so can we change this to browser.waitUntil(() => scheduled.scheduledJobRows.length > 1);

e2e/components/login/login.po.js
```diff
@@ -1,60 +1,64 @@
...
+    } else {
+      un = $(this.elements.component).$(this.elements.userName);
+      pw = $(this.elements.component).$(this.elements.password);
+      btn = $(this.elements.component).$(this.elements.login);
+    }
+    un.addValue(credentials[user].email || credentials[user].username);
+    pw.addValue(credentials[user].password);
+    btn.scrollAndClick();
+    if (usingLegacy) {
+      this.toggleLoginComponent();
```
palakpatel311, line 43: I don't think we need line 50 and 51 if we call waitToBeLoggedIn() always after logIn(). Which also makes me think of combining both.

palakpatel311 (dismissed):

ashwinimore (dismissed):

tmy1313 (dismissed):

palakpatel311 (dismissed):

kekey1 (dismissed):

ashwinimore (dismissed):

tmy1313 (dismissed):

ashwinimore (dismissed):

palakpatel311 (dismissed):

ashwinimore (approved):

tmy1313 (approved):

kekey1 (approved):

palakpatel311 (approved):

PR #912, opened 2021-06-03
OCD-3637: Add AQA test regarding permissions for surveillance

palakpatel311:
[#OCD-3637]


e2e/components/navigation/navigation.po.js
```diff
@@ -1,17 +1,27 @@
...
+    this.elements = {
+      reportsToggle: '#reports-toggle',
+      reports: '#reports-dropdown-menu>li',
+      surveillanceToggle: '#surveillance-toggle',
+      surveillanceOptions: '#surveillance-dropdown-menu>li',
+    };
+  }
+
+  get reportsToggle() {
+    return $(this.elementselements.reportsToggle);
```
andlar: Looks like a typo. If this is a typo, but the tests are passing, then maybe this element doesn't need to exist at all?

e2e/components/navigation/navigation.readonly.spec.js
```diff
@@ -59,10 +61,42 @@ describe('when logged in', () => {
...
+      expect(surveillanceOptions.size).toBe(expected.length);
+      expected.forEach((exp) => {
+        expect(surveillanceOptions.has(exp)).toBe(true);
+      });
+    });
+  });
+  describe('as ONC-STAFF', () => {
+    beforeEach(() => {
+      login.logIn('oncstaff');
+      login.logoutButton.waitForDisplayed();
```
andlar: Leaving a note here that when the PR for OCD-3546 gets merged, this line will likely need to be removed

e2e/components/upload/upload-surveillance/upload-surveillance.readonly.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    loginComponent.logIn('onc');
+    hooks.waitForSpinnerToDisappear();
+  });
+
+  it('can\'t see surveillance upload section', () => {
+    expect(uploadSurveillanceComponent.root.isDisplayed()).toBe(false);
+  });
+});
+
+describe('when on administration upload page as admin', () => {
```
andlar, line 31: I think our general rule is that we want to test the actions that our users will be taking, and usually that means not testing the ROLE_ADMIN stuff as much, as we want to make sure we catch things before ROLE_ACB runs into them

e2e/components/upload/upload-surveillance/upload-surveillance.spec.js
```diff
@@ -3,7 +3,8 @@ import LoginComponent from '../../login/login.po';
 import Hooks from '../../../utilities/hooks';
 import ToastComponent from '../../toast/toast.po';
 
-let hooks, loginComponent, toast, uploadSurveillanceComponent;
+let hooks; let loginComponent; let toast; let
+  uploadSurveillanceComponent;
```
andlar:
```suggestion
let hooks; 
let loginComponent; 
let toast; 
let uploadSurveillanceComponent;
```

e2e/components/upload/upload-surveillance/upload-surveillance.po.js
```diff
@@ -1,28 +1,31 @@
...
-  uploadButton: '.MuiButton-containedPrimary',
-};
-
 const path = require('path');
 
 class UploadSurveillanceComponent {
-  constructor () { }
+  constructor() {
+    this.elements = {
+      root: 'chpl-upload-surveillance-bridge',
```
andlar, line 6: I'm not sure if there's a better option, but the `*-bridge` elements are temporary things that we hope will go away over time. If it's possible to use something else to scope this element that would be better for long term maintenance purposes

andlar (changes requested):

e2e/components/upload/upload-surveillance/upload-surveillance.readonly.spec.js
palakpatel311: Added a test for acb as well and kept admin test as onc and admin behavior is different in this case.

e2e/components/upload/upload-surveillance/upload-surveillance.po.js
palakpatel311: I am aware, but there is not other element to go by. when we remove *-bridge , hopefully it will be easy to just remove bridge from the locator.

e2e/components/navigation/navigation.po.js
palakpatel311: Tests were failing on these so fixed typo.

e2e/components/navigation/navigation.readonly.spec.js
palakpatel311: Removed

andlar (dismissed):

athomas-git (dismissed):

e2e/components/navigation/navigation.readonly.spec.js
```diff
@@ -40,7 +40,7 @@ describe('when logged in', () => {
     });
   });
 
-  describe('as ONC', () => {
+  describe('as an ONC', () => {
```
andlar: Suggest

```suggestion
  describe('as an ONC User', () => {
```
or

```suggestion
  describe('as ROLE_ONC', () => {
```

The phrase "an ONC" doesn't make sense, as there's no such thing as a single "ONC"

Same comment with the ONC-STAFF

e2e/components/navigation/navigation.readonly.spec.js
```diff
@@ -97,4 +97,25 @@ describe('when logged in', () => {
       });
     });
   });
+  describe('as an Admin', () => {
```
andlar:
```suggestion
  describe('as an ADMIN user', () => {
```

or

```suggestion
  describe('as ROLE_ADMIN', () => {
```

Here the problem is that the word "Admin" is not specific to our ROLE_ADMIN, as it generally just means "an administrative user", which in our system could be any logged in user

andlar (commented):
Mostly grammar stuff, but I think it's reasonable to have text that's supposed to be able to be read and mean something to be grammatically correct

andlar (dismissed):

athomas-git (dismissed):

andlar (approved):

athomas-git (approved):

PR #913, opened 2021-06-04
OCD-3637 - Update ROLE_ONC_STAFF permissions for "Surveillance"

tmy1313:


src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,244 @@
...
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+} from '@material-ui/core';
+
+import ChplCriterionTitle from '../../../../util/criterion-title';
+import { ChplTooltip } from '../../../../util/chpl-tooltip';
+import { getAngularService } from '.';
+import requirementPropType from '../../../../../shared/prop-types/surveillance-requirement';
```
andlar: The `prop-types` types that come in the package don't use the words "PropType" in their name; they are things like `string`, `object`, etc. On the prop-types I added for criteria editing I called things `testStandard` and `selectedTestStandard` for the "data" item and the "data item associated with a certification result", respectively. I feel like these three ought to be `requirement`, `nonconformity` and `surveillance`, or, if we want to be prepared for Direct Review things: `surveillance`, `surveillanceRequirement`, and `surveillanceNonConformity`. Oh, and I think the style guide refers to "Non-Conformity" as a mixed case dashed item, so where we can use `nonConformity` I think we should

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,244 @@
...
+  },
+  nonconformityAccordionSummaryStatus: {
+    textAlign: 'right',
+  },
+  iconSpacing: {
+    marginLeft: '4px',
+  },
+}));
+
+function ChplNonconformityView({ surveillance, requirement, nonconformity }) {
```
andlar:
```suggestion
function ChplNonconformityView(props) {
```

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,244 @@
...
+  const DateUtil = getAngularService('DateUtil');
+  const currentSurveillance = useState(surveillance)[0];
```
andlar:
```suggestion
  const [currentSurveillance] = useState(props.surveillance);
```

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,244 @@
...
+  const currentRequirement = useState(requirement)[0];
+  const currentNonconformity = useState(nonconformity)[0];
+  const dateFormat = 'MMM d, y';
```
andlar: Is this necessary? We're creating a new component, right? So there's no previous thing we're trying to recreate, which makes me wonder if we can just use whatever our default format is for these dates. We did say we had a standard for what dates should look like in the UI, and I hope our `DateUtil` parser function returns that by default

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,244 @@
...
+  const classes = useStyles();
+
+  return (
+    <ThemeProvider theme={theme}>
+      <Accordion className={classes.nonconformityAccordion}>
+        <AccordionSummary
+          expandIcon={<ExpandMoreIcon color="primary" fontSize="large" />}
+          data-testid="nonconformity-accordion-header"
+        >
+          <Grid container spacing={4}>
```
andlar: I thought we were trying to replace the `<Grid>` elements with a css-grid styled thing

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,244 @@
...
+                      style={{ width: '33%' }}
+                    >
+                      Attribute
+                    </TableCell>
+                    <TableCell>Value</TableCell>
+                  </TableRow>
+                </TableHead>
+                <TableBody>
+                  <TableRow>
+                    <TableCell component="th" scope="row">
```
andlar: I like the `scope` attribute. Will need to make sure I have those in my tables

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.test.jsx
```diff
@@ -0,0 +1,327 @@
...
+      it('should display the "Other Requirement" as the non-conformity type', async () => {
+        const tableRow = screen.getByTestId('pass-rate-row');
+
+        await waitFor(() => {
+          expect(tableRow).toBeVisible();
+        });
+      });
+    });
+  });
+  /*
```
andlar: Debug code? FWIW, I agree that we don't need to test the functionality of that function. It might be worth it to check that it's getting called with the correct data. Or at least, that would be a valid test on this component

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,185 @@
+/* eslint-disable import/no-cycle */
```
andlar: Can we disable this rule just for the line that triggers it?

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,185 @@
...
+  nonconformityContainer: {
+    paddingTop: '16px',
+    paddingLeft: '8px',
+    paddingRight: '8px',
+  },
+}));
+
+const getSurveillanceResults = (surv) => {
+  const results = [];
+  surv.requirements.forEach((req) => req.nonconformities.forEach((nc) => {
```
andlar: Is there a reason to use `forEach` and not `map` or possibly `flatMap`? It looks like you're converting a surveillance into a flattened list, and that's the standard use case for map. `forEach` is usually used when we're trying to trigger some sort of side effect

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,185 @@
...
+  const classes = useStyles();
+
+  useEffect(() => {
+    setSurveillanceResults(getSurveillanceResults(currentSurveillance));
+  }, [currentSurveillance]);
+
+  return (
+    <ThemeProvider theme={theme}>
+      <TableContainer component={Paper}>
+        <Table size="small" aria-label="Surveillance Table" data-testid="surveillance-attributes-table">
```
andlar: Can the tests find this table by its label? I think I've found input elements by their aria-label. If so, we wouldn't need to add the data-testid element. I recognize it's a valid way of finding the element, but I believe it's seen as a last resort method, as it requires changing feature code for the test harness, and that's something that should be rare

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,185 @@
...
+                  />
+                </ChplTooltip>
+              </TableCell>
+              <TableCell data-testid="reqs-surveilled-cell">
+                { currentSurveillance.requirements?.length > 0
+                  && (
+                    <ul className={classes.unindentedData}>
+                      { currentSurveillance.requirements.map((req) => (
+                        <li key={req.id}>
+                          { `${req.type.name} ${req.criterion ? ': ' : ''}` }
```
andlar:
```suggestion
                          { `${req.type.name} ${req.criterion && ': '}` }
```

Not sure if this will work, but it would avoid having a ternary that has a `noop` else

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,185 @@
...
+                    <ul className={classes.unindentedData}>
+                      { surveillanceResults.map((result) => (
+                        <li key={result.id}>
+                          { `${result.statusName} Non-Conformity Found for` }
+                          <ChplCriterionTitle criterion={result.criterion} useRemovedClass />
+                        </li>
+                      ))}
+                    </ul>
+                  )}
+                { surveillanceResults.length === 0 && 'No Non-Conformities Found' }
```
andlar, line 149: :+1:

src/app/components/listing/details/surveillance/surveillance-view.test.jsx
```diff
@@ -0,0 +1,277 @@
...
+  requirements: [],
+  authority: 'ROLE_ACB',
+  lastModifiedDate: 1597786978488,
+};
+
+const dateUtilMock = {
+  timestampToString: jest.fn(() => 'June 1, 2020'),
+};
+
+jest.mock('./nonconformity/nonconformity-view', () => () =><div data-testid="non-conformity-component" />);
```
andlar: This is interesting. Is this the way to mock out the sub-component with this DIV instead? I like it

src/app/components/listing/details/surveillance/view-container.html
```diff
@@ -0,0 +1,16 @@
+<div role="modal" aria-labeled-by="edit-surveillance-label">
+  <div class="modal-header">
+    <button type="button"
+            class="close pull-right"
+            aria-label="Cancel"
+            ng-click="$ctrl.dismiss();">
+      <span aria-hidden="true">&times;</span>
+    </button>
+    <h4 class="modal-title" id="edit-surveillance-label">
```
andlar: Is this the right id?

src/app/components/surveillance/reporting/relevant-listing.component.js
```diff
@@ -36,86 +37,74 @@ export const SurveillanceReportRelevantListingComponent = {
...
+    editSurveillance(relevantSurveillance) {
+      const that = this;
       this._fixRequirementOptions();
-      this.networkService.getListing(this.listing.id, true).then(listing => {
-        let surveillance = listing.surveillance.find(s => s.id === relevantSurveillance.id);
+      this.networkService.getListing(this.listing.id, true).then((listing) => {
+        const surveillance = listing.surveillance.find((s) => s.id === relevantSurveillance.id);
         that.uibModalInstance = that.$uibModal.open({
-          component: 'aiSurveillanceEdit',
+          component: 'chplSurveillanceViewContainerComponent',
```
andlar: We're still supposed to be able to edit surveillances from the reporting module, as ROLE_ADMIN, ROLE_ONC, ROLE_ACB, right? I think it's only supposed to be ROLE_ONC_STAFF that gets the view only mode. (Maybe ROLE_ONC too... not sure about that)

src/app/components/util/criterion-title.test.jsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import {
+  cleanup, render, screen, waitFor,
+} from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { within } from '@testing-library/dom';
+//import { when } from 'jest-when';
+//import userEvent from '@testing-library/user-event';
```
andlar: Debug lines?

src/app/shared/prop-types/acb.js
```diff
@@ -11,4 +11,4 @@ const acb = shape({
   website: string,
 });
 
-export { acb };
+export default { acb };
```
andlar: Why did these need to change? I think the pattern is that we should be using 

```
export { item, item2, item3 };
```
when there are more than one export and 

```
export default item1;
```

when there's only one. This looks like a hybrid, and I'm not sure what happened to need it to be this way

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
andlar: I just got down to the definitions of those prop-types, and see them defined without the `PropType` part of the name. So here you're importing them and renaming them at the same time to work in this component, which is perfect

👍

kekey1 (dismissed):

src/app/components/listing/details/surveillance/surveillance-view.test.jsx
tmy1313: Yes - this is how we can mock sub-components.

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
tmy1313: Without using a specific date format the `date-util.timestampToString` will return a string contain both the date and time.  In this case we don't want to display the time portion.

src/app/components/listing/details/surveillance/surveillance-view.jsx
tmy1313: Can't do `map` - it returned a array of arrays (with a single object).  Could look into using `flatMap`...

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
andlar: What about `getDisplayDateFormat()`?

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,180 @@
...
+  TableContainer,
+  TableHead,
+  TableRow,
+  Typography,
+} from '@material-ui/core';
+
+import { ChplTooltip } from '../../../util/chpl-tooltip';
+/* eslint-disable import/no-cycle */
+import { getAngularService } from '.';
+/* eslint-enable import/no-cycle */
```
andlar: Not required, but here if you want it:

```suggestion
import { getAngularService } from '.'; // eslint-disable-line import/no-cycle
```

src/app/shared/prop-types/contact.js
```diff
@@ -8,4 +8,4 @@ const contact = shape({
   title: string,
 });
 
-export { contact };
+export default { contact };
```
andlar:
```suggestion
export default contact;
```

src/app/shared/prop-types/developer.js
```diff
@@ -11,4 +11,4 @@ const developer = shape({
   website: string,
 });
 
-export { developer };
+export default { developer };
```
andlar:
```suggestion
export default developer;
```

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,250 @@
...
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+} from '@material-ui/core';
+
+import ChplCriterionTitle from '../../../../util/criterion-title';
+import { ChplTooltip } from '../../../../util/chpl-tooltip';
+import { getAngularService } from '.';
```
andlar: If this were imported directly, I think we'd be able to remove the "ignore the eslint rule about no-cycle" comment

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,250 @@
...
+  },
+  iconSpacing: {
+    marginLeft: '4px',
+  },
+}));
+
+function ChplNonconformityView(props) {
+  /* eslint-disable react/destructuring-assignment */
+  const DateUtil = getAngularService('DateUtil');
+  const [currentSurveillance] = useState(props.surveillance);
```
andlar: Are you using the `current*` convention for any particular reason? Is there a reason you can't just use `surveillance` without the `current` part?

src/app/components/listing/details/surveillance/nonconformity/nonconformity-view.jsx
```diff
@@ -0,0 +1,250 @@
...
+              <ChplCriterionTitle criterion={currentNonconformity.criterion} />
+              { ` ${currentRequirement.result.name}` }
+            </div>
+            <div className={classes.nonconformityAccordionSummaryStatus}>
+              { currentNonconformity.status.name }
+            </div>
+          </div>
+        </AccordionSummary>
+        <AccordionDetails>
+          <Container>
```
andlar: What happens if this container element is removed? With the table container just below it I wonder if this is just adding a layer of border/padding/something that isn't necessary

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,178 @@
...
+  },
+}));
+
+const getSurveillanceResults = (surv) => surv.requirements.map((req) => req.nonconformities.map((nc) => ({
+  id: req.id,
+  statusName: nc.status.name,
+  criterion: req.criterion,
+  requirement: req.requirement,
+})))
+  .flat();
```
andlar: The `map` and `flat` commands can be combined as a single `flatMap`, that according to the docs is identical to what you have, but is slightly more efficient: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

src/app/components/listing/details/surveillance/surveillance-view.jsx
```diff
@@ -0,0 +1,178 @@
...
+
+function ChplSurveillanceView({ surveillance }) {
```
andlar: I thought we were sticking with `props`?

src/app/components/surveillance/reporting/relevant-listing.component.js
```diff
@@ -36,86 +37,75 @@ export const SurveillanceReportRelevantListingComponent = {
...
-          backdrop: 'static',
-          keyboard: false,
-          size: 'lg',
-          resolve: {
-            surveillance: () => { return surveillance; },
-            surveillanceTypes: () => { return that.surveillanceTypes; },
-            workType: () => { return 'edit'; },
-          },
-        });
-        that.uibModalInstance.result.then(() => {
```
andlar: Without this `.result.then` construction, if ROLE_ADMIN or ROLE_ACB goes into a surveillance and edits it, does the page refresh and show the updates?

kekey1 (dismissed):

src/app/shared/prop-types/criterion.js
```diff
@@ -1,7 +1,11 @@
 /* eslint-disable import/no-extraneous-dependencies */
+<<<<<<< HEAD
 import {
   bool, number, shape, string,
 } from 'prop-types';
+=======
+import { bool, number, shape, string } from 'prop-types';
+>>>>>>> staging
```
andlar: Merge issue

src/app/components/util/index.js
```diff
@@ -1,11 +1,13 @@
+import ChplCriterionTitle from './criterion-title';
 import ChplDialogTitle from './chpl-dialog-title';
 import { ChplEllipsis } from './chpl-ellipsis';
 import ChplHighlightCures from './highlight-cures';
 import ChplLink from './chpl-link';
 import ChplTextField from './chpl-text-field';
-import ChplTooltip from './chpl-tooltip';
+import { ChplTooltip } from './chpl-tooltip';
```
andlar: I think this component is being exported as a default, so shouldn't have these braces around it

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #922, opened 2021-06-22
OCD-3656 - Create visible display of non-production environments

tmy1313:


src/app/components/util/non-prod-indicator.jsx
```diff
@@ -0,0 +1,86 @@
...
+import { Toolbar } from '@material-ui/core';
+import { makeStyles, ThemeProvider } from '@material-ui/core/styles';
+import Typography from '@material-ui/core/Typography';
+
+import theme from '../../themes/theme';
+// eslint-disable-next-line import/no-cycle
+import { getAngularService } from '.';
+
+const useStyles = makeStyles(() => ({
+  NavBarCallout: {
```
andlar: Is there a reason these are PascalCase? These are technically JS objects, and our convention on JS stuff is to camelCase them. I know the MUI components have PascalCase, but all of our internal classes are camelCase

One other thing I think we should take advantage of, is that these styles are scoped to this component, so we can be pretty generic on the names. I've been using things like `container: { stuff }` for the "basic / root" component. Not required to change anything, but it does seem to make it a little simpler (to me)

Oh, and I just noticed this class isn't actually used anywhere. The above comments should apply to the other class though

src/app/components/util/non-prod-indicator.jsx
```diff
@@ -0,0 +1,86 @@
...
+          headerValue = response.headers('environment');
+        }
+        setProduction(headerValue === 'PRODUCTION');
+      });
+  }, []);
+
+  return (
+    <ThemeProvider theme={theme}>
+      <>
+        { !production
```
andlar: If this component is on PROD, it will return an element that has the `ThemeProvider` and then an empty DIV. Then best practice in React is to `return null` if an element shouldn't render (https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering). I suggest putting something either above or below the `useEffect` method that does something like

```
if (production) {
return null;
}
```

That would return null until the network call finished, and then the banner would appear. According to that link above, the lifecycle methods continue running, even if the `return null` happens, so it shouldn't interrupt the `useEffect`

src/app/components/util/non-prod-indicator.jsx
```diff
@@ -0,0 +1,86 @@
...
+          && (
+            <Toolbar className={classes.NavBarTextBox} id="non-prod-indicator">
+              <Typography variant="body2" noWrap>
+                THIS IS NOT THE PRODUCTION SITE
```
andlar: This could all be replaced with something like

```
{
Array.from({length: 15}, () => ('THIS IS NOT THE PRODUCTION SITE')).join(' <b>|</b> ');
}
```

src/app/services/network.service.js
```diff
@@ -849,6 +849,10 @@ export default class NetworkService {
     return this.apiPUT('/versions', versionObject);
   }
 
+  getSystemStatus() {
+    return this.$http.get('/rest/system-status');
```
andlar, line 853: Is there a test for this?

src/app/themes/theme.js
```diff
@@ -56,7 +56,7 @@ const theme = createMuiTheme({
       lineHeight: '1.3em',
     },
     body2: {
-      fontSize: '1em',
+      fontSize: '0.875em',
```
andlar: How many other places are we using body2? Has this change impacted where those are being used in such a way that we need to check them for font size / contrast issues?

src/app/components/util/non-prod-indicator.jsx
```diff
@@ -0,0 +1,86 @@
+import React, { useEffect, useState } from 'react';
+import { Toolbar } from '@material-ui/core';
+import { makeStyles, ThemeProvider } from '@material-ui/core/styles';
+import Typography from '@material-ui/core/Typography';
+
+import theme from '../../themes/theme';
+// eslint-disable-next-line import/no-cycle
+import { getAngularService } from '.';
```
andlar: If you were to make this a direct import from the `../../service/angular-react-helper` directory, and remove the import in the `./index.js` the import/no-cycle warning would go away

src/app/components/util/non-prod-indicator.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+  NavBarTextBox: {
+    minHeight: '25px',
+    backgroundColor: '#c44f65',
+    width: '100%',
+    color: '#ffffff',
+  },
+}));
+
+function ChplNonProdIndicator() {
```
kekey1, line 18: I won't hold up the PR for this since it clearly works and will do what we need, but I guess I wish it was a little more generic? Like if it was called ChplEnvironmentIndicator and used text that came from the Environment header instead of the hard-coded string in the UI. If we ever wanted to have different headers for different environments we could do that with environment properties coming from the server. I know that's not needed for the ticket, so this is just a comment and maybe something to think about in the future if we ever need that.

src/app/themes/theme.js
tmy1313: I saw we are using in the `login.jsx` component.  This is some of the thigs I wish we could get standardized and not have to touch them.  I pulled this updated value from https://codesandbox.io/s/chpl-material-ui-jq7q8?file=/src/NavBarCallout.js.

andlar: If we're using this value in the login.jsx component then it's not a global value. If you need the body2 to be a different font size in a particular component, then we shouldn't be changing the global styles

kekey1 (dismissed):

src/app/components/util/index.js
```diff
@@ -1,7 +1,8 @@
 import ChplDialogTitle from './chpl-dialog-title';
-import { ChplEllipsis } from './chpl-ellipsis';
+import { ChplEllipsis } from './ellipsis';
```
andlar: I don't think this import is right

andlar (approved):

kekey1 (approved):

PR #930, opened 2021-07-01
OCD 3682: AQA tests for complaints

palakpatel311:


e2e/pages/surveillance/complaints/complaints.readonly.spec.js
```diff
@@ -1,13 +1,13 @@
-import ComplaintsComponent from './complaints.po';
-import LoginComponent from '../../login/login.po';
+import ComplaintsPage from '../../surveillance/complaints/complaints.po';
```
andlar: The reason this was a component in the first place is that the same complaints component is used on the main "complaints management" page as is used inside the "relevant complaints" section of the "surveillance -> reporting" section

andlar: I think it's reasonable to add a ComplaintsPage object that maps to the complaints management page, but suggest leaving the complaints component as a component, so it could be used in the reporting section

palakpatel311: Rearranged to have all the complaints, table, buttons from complaints page on ComplaintsPage and moved add/edit complaint form elements to the ComplaintsComponent.

andlar: I'm still not sure why any of the Complaints elements ended up in a dedicated Complaints Page. The only thing that's on the Complaints page that isn't in the Surveillance Reporting page is the "Add new Complaint" button, and the navigation to get there, so why wouldn't we want to keep all of the Complaint elements in a single Complaint component?

e2e/pages/surveillance/complaints/complaints.readonly.spec.js
```diff
@@ -1,13 +1,13 @@
-import ComplaintsComponent from './complaints.po';
-import LoginComponent from '../../login/login.po';
+import ComplaintsComponent from '../../../components/surveillance/complaints/complaints.po';
```
andlar: Should this file be in the `../pages/..` directory tree or should it be back in the `../components/..` directory tree? It's testing just the complaints component, right?

e2e/pages/surveillance/complaints/complaints.readonly.spec.js
```diff
@@ -1,13 +1,13 @@
...
+import LoginComponent from '../../../components/login/login.po';
 import Hooks from '../../../utilities/hooks';
 
-let component;
 let hooks;
 let login;
+let complaints;
```
andlar: Is there a reason to have changed this variable name? When it was `component` I knew it was talking about the component that was under test, but with the variable name `complaints` I would assume this is a variable that contains some sort of array of complaint objects, not the component that holds the entire thing

e2e/workflowTest/complaints/add-new-complaint.spec.js
```diff
@@ -0,0 +1,58 @@
...
+beforeEach(async () => {
+  login = new LoginComponent();
+  hooks = new Hooks();
+  page = new ComplaintsPage();
+  complaints = new ComplaintsComponent();
+  hooks.open('#/surveillance/complaints');
+  await hooks.waitForSpinnerToDisappear();
+});
+
+describe('As ROLE_ACB user', () => {
```
andlar: The grammar here seems weird

```suggestion
describe('as a ROLE_ACB user', () => {
```
or

```suggestion
describe('as ROLE_ACB', () => {
```

e2e/workflowTest/complaints/add-new-complaint.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    expect(complaints.fieldError('complainant-type')).toBe('Complainant Type is required');
+    expect(complaints.fieldError('acb-complaint-id')).toBe('ONC-ACB Complaint ID is required');
+    expect(complaints.fieldError('summary')).toBe('Complaint Summary is required');
+  });
+
+  it('should be able to add new complaint with required fields', () => {
+    const timestamp = (new Date()).getTime();
+    const fields = {
+      body: 'Drummond Group',
+      receivedDate: '06/23/2021',
```
andlar, line 43: The "receivedDate" is the value that the table sorts on first, right? So this is set to a few days ago, as the PR is in place, but in a month or two there will be new complaints in the system, and this complaint will be 5, 10, or more rows down in the table. With the table as is, showing only 5 rows, this complaint won't show up at all. Could this date be set as something like "yesterday" instead? That way we'd be much more likely to not have some other complaint come in and mess up the ordering

e2e/workflowTest/complaints/edit-complaint.spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+  afterEach(() => {
+    login.logOut();
+  });
+
+  it('should display correct error messages regarding received and closed date', () => {
+    const timestamp = (new Date()).getTime();
+    const fields = {
+      body: 'Drummond Group',
+      receivedDate: '06/23/2021',
```
andlar, line 30: The "add new complaint" test creates a complaint with the same receivedDate as this one. What happens if this one runs first, or second, compared to that other one? If the point of this test is to create one, then edit the one that was created, there's no guarantee that this test is editing the newly created one

Would it make sense to put all three of these add/edit/delete tests into a single workflow test? Where the `it` block is something like:

```
it('should add/edit/delete complaints`, () => {
  try to create a complaint
  expect errors to exist
  create a complaint
  expect the complaint be created
  try to edit the complaint
  expect to get edit errors
  edit the complaint
  expect to see successful edits
  delete the complaint
  expect to see the complaint gone
```

e2e/components/surveillance/complaints/complaints.po.js
```diff
@@ -1,21 +1,68 @@
...
+    return $(this.elements.downloadResultsButton);
+  }
+
+  get filter() {
+    return $(this.elements.filter);
+  }
 
-  getComplaintsTableHeaders () {
-    return $(elements.complaintsTable).$('thead').$$('th');
+  getcellValue(row, col) {
```
andlar: This doesn't look like the standard camelCase we use

```suggestion
  getCellValue(row, col) {
```

I'm also not sure it should be in this PO file at all. It looks like such a generic method that maybe it belongs in the hooks file. There's nothing about this method that cares what component/page it's on

e2e/workflowTest/complaints/add-new-complaint.spec.js
```diff
@@ -0,0 +1,58 @@
...
+      summary: `Test Summary - ${timestamp}`,
+    };
+    page.addNewComplaint();
+    hooks.waitForSpinnerToDisappear();
+    complaints.set(fields);
+    complaints.saveComplaint();
+    hooks.waitForSpinnerToAppear();
+    hooks.waitForSpinnerToDisappear();
+    complaints.filter.addValue(fields.acbId);
+    expect(complaints.getcellValue(1, 4)).toBe(fields.acbId);
```
andlar: Both the 1 and the 4 here are magic numbers; they don't mean anything to a reader, and there's no way of telling what they should be. I think the 4 could be replaced with something like:

```
const ACB_ID_IDX = 4;
```

though I'm not sure what to replace the 1 with; maybe `FIRST_ROW`?

e2e/pages/surveillance/complaints/complaints.readonly.spec.js
palakpatel311: No, It is checking complaints table and download button.

e2e/workflowTest/complaints/edit-complaint.spec.js
palakpatel311: I am using unique acbId with timestamp so it will be all different complaints. More over keeping them separate will make it easier to add more test/validation specifics to add/edit/delete

e2e/components/surveillance/complaints/complaints.po.js
palakpatel311: Moved table methods to hooks

e2e/workflowTest/complaints/add-new-complaint.spec.js
palakpatel311: That's why I am searching complaint using acbId before edit/delete action.

e2e/pages/surveillance/complaints/complaints.readonly.spec.js
andlar: The table, download button, filter, and add new complaint button are all part of the complaints component, right? The only thing on the complaints page that's not part of the complaints component is the navigation that gets there, and the navigation that goes somewhere else

There isn't even a page object for the complaints page, so why is the spec file in this directory at all? 
image

ashwinimore (dismissed):

andlar (dismissed):

ashwinimore (dismissed):

andlar (dismissed):

andlar (approved):

ashwinimore (approved):

PR #932, opened 2021-07-06
OCD-3708: AQA test to add RWT data on listing edit

ashwinimore:


e2e/components/listing/listing.po.js
```diff
@@ -10,6 +10,10 @@ const listingEditElements = {
   saveCertifiedProduct: 'button=Save Certification Criteria',
   closeListingEditButton: 'button.close.pull-right.ng-isolate-scope',
   yesConfirmation: '//button[text()="Yes"]',
+  rwtPlansUrl: '#rwt-plans-url',
```
palakpatel311, line 13: Can you move all of these new added changed to e2e/pages/listing/listing.po.js?
I was planning to remove this file once we update upload specs to new UI so going forward we will be using listing page class.

e2e/workflowTest/edit-listing-test/edit-basic-listing-data.spec.js
```diff
@@ -0,0 +1,61 @@
+import ListingPage from '../../pages/listing/listing.po';
+import Hooks from '../../utilities/hooks';
+import loginComponent from '../../components/login/login.po';
```
palakpatel311: loginComponent L needs to be capital

e2e/workflowTest/edit-listing-test/edit-basic-listing-data.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    it('should be able to add RWT data to the listing and save edits', () => {
+      listingEdit.rwtPlansUrl.setValue(testPlansUrl);
+      listingEdit.rwtPlansCheckDate.setValue(plansDateInput);
+      listingEdit.rwtResultsUrl.setValue(testResultsUrl);
+      listingEdit.rwtResultsCheckDate.setValue(resultsDateInput);
+      action.save();
+      hooks.waitForSpinnerToDisappear();
+      browser.waitUntil(() => toast.toastTitle.isDisplayed());
```
palakpatel311, line 54: May be good idea to check toaster title to be success or update processing (not sure which one it shows)

palakpatel311 (changes requested):

e2e/components/listing/listing.po.js
ashwinimore: I was thinking about keeping this component and rename it to listing-edit (?) and use listing.po.js under pages for page objects on listing details page(not edit). Listing edit component can be used on both before and after listing confirm vs. listing page is post confirm. What are your thoughts on that?

e2e/workflowTest/edit-listing-test/edit-basic-listing-data.spec.js
ashwinimore: I think this is good for this test as the actual data from update is verified after this step and test won't pass unless update was successful. There are failures on some existing tests that expect success and not find the text due to either the toast is missing or has other text, so not adding that check would avoid false failures of those kind

e2e/components/listing/listing.po.js
palakpatel311: Yes renaming sounds good. I was checking src code and to be consistent with that, can we add this edit component under e2e/components/listing/edit(new folder)?
This will affect existing some of the tests

palakpatel311 (dismissed):

palakpatel311 (approved):

athomas-git (approved):

andlar (approved):

PR #933, opened 2021-07-12
OCD-3659: Update MUU/PI AQA tests based on flag

palakpatel311:


e2e/utilities/hooks.js
```diff
@@ -14,6 +15,13 @@ class Hooks {
   waitForSpinnerToAppear () {
     browser.waitUntil( () => $('#loading-bar-spinner').isDisplayed());
   }
+
+  getFlagState(flagName){
+    browser.url(browser.options.baseUrl + urls.ff4jURL);
+    let jsonObj= JSON.parse($('body').getText())
+    const foundFlag = jsonObj.find(flag => flag.key === flagName);
+    return foundFlag?.active;
+  }
```
andlar: Could we put this function in one of the preparation hooks that's defined in the wdio.conf file? As written, it'll require a network call for each test that uses a flag, which could become a lot of tests as we implement this. If it were put in some sort of preparation step (maybe `before`, `beforeSession`, or `onPrepare`) to get the entire flag object and save it in the browser (or maybe config) then this method could just reference that globally stored object

e2e/utilities/hooks.js
```diff
@@ -1,3 +1,4 @@
+const urls = require('../config/urls');
```
andlar: Is this being used for anything?

wdio.conf.js
```diff
@@ -231,7 +231,12 @@ exports.config = {
    */
   before: function (capabilities, specs) {
     assert = require('chai').assert;
-
+    browser.addCommand('getFlag', function () {
```
andlar, line 237: I like this solution, and I'm glad that it works, but now I'm wondering why we're adding a command to the browser object that we only call once. Would it make sense to just load the object using an anonymous function? Maybe something like:

```suggestion
    global.flagObj = () => {
```

palakpatel311: It doesn't open URL this way. So to assign value to global variable this is how i have to call.

andlar (dismissed):

ashwinimore (dismissed):

andlar (approved):

ashwinimore (approved):

PR #963, opened 2021-09-10
OCD-3662: Add AQA tests for CQM errors on flexible upload and correct tools errors

palakpatel311:
[#OCD-3662]


e2e/workflowTest/upload-confirm-listing-test/upload-listing-error-warning-validation.spec.js
```diff
@@ -9,38 +9,39 @@ let hooks;
...
-beforeEach(() => {
+beforeAll(() => {
   confirmPage = new ConfirmPage();
   uploadListingComponent = new UploadListingComponent();
   loginComponent = new LoginComponent();
   hooks = new Hooks();
   toast = new ToastComponent();
   hooks.open('#/administration/upload');
   loginComponent.logIn('admin');
+  uploadListingComponent.uploadListingBeta('../../../resources/upload-listing-beta/2015_InvalidData.csv');
```
andlar: These were previously only uploaded if the environment was not `STG`. With this change, will these get uploaded on STG? Or try to? STG doesn't have this functionality, right?

palakpatel311: No When we pass ENV as stage, it will not even run before all hook as there will be no tests to be run

andlar: If that's the case then the `if` check should be moved higher up. I didn't know the beforeAll won't execute if there aren't any tests under it, and it's not obvious that is the case. If the if check is outside of everything that won't execute it'll make it much more clear

e2e/workflowTest/upload-confirm-listing-test/upload-listing-error-warning-validation.spec.js
```diff
@@ -9,38 +9,39 @@ let hooks;
...
 let toast;
 let uploadListingComponent;
-const inputs = require('./dataProviders/chplId-data-validation-dp');
+const inputs = require('./dataProviders/upload-listing-error-warning-validation-dp');
 
 const validListingId = '15.04.04.1722.AQA4.03.01.1.200620';
 
-beforeEach(() => {
+if (process.env.ENV !== 'stage') {
+beforeAll(() => {
```
andlar: The indentation here looks like it's incorrect, but I think the functionality is right

andlar (dismissed):

ashwinimore (dismissed):

andlar (dismissed):

ashwinimore (approved):

andlar (approved):

PR #967, opened 2021-09-16
Ocd 3480: AQA testing for Surveillance Management - Reporting

palakpatel311:


e2e/components/action-bar/action-bar.po.js
```diff
@@ -1,22 +1,40 @@
...
+  get saveButton () {
+    return $(this.elements.save);
+  }
+
+  yes (){
+    $(this.elements.yes).click();
+  }
+
+  delete () {
+    $(this.elements.delete).click();
```
andlar: The alignment of the parenthesis in these functions doesn't match up. Does this pass linting?

e2e/pages/surveillance/reporting/annual/annual.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    reportingPage.expandAcb('Drummond Group');
+    reportingPage.editAnnualReport('Drummond Group', 2019).click();
+  });
+
+  afterEach(() => {
+    if (toast.toastTitle.isExisting()) {
+      toast.clearAllToast();
+    }
+    loginComponent.logOut();
+  });
```
andlar: This `afterEach` is the same for both ROLE ONC and ROLE ACB. Can it be lifted up so it's only in one place?

e2e/pages/surveillance/reporting/annual/annual.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+  xit('can cancel initiating of annual report', () => {
```
andlar: Is this the one that's ignored because it's an existing bug?

andlar (dismissed):

andlar (dismissed):

athomas-git (dismissed):

e2e/pages/surveillance/reporting/quarterly/quarterly.spec.js
```diff
@@ -115,7 +115,7 @@ describe('when logged in as a ROLE_ACB', () => {
     hooks.waitForSpinnerToDisappear();
     action.cancel();
     action.yes();
-    hooks.waitForSpinnerToDisappear();
+    browser.waitUntil(() => reportingPage.editQuarterlyReport('Drummond Group', 2022, 'Q4').isDisplayed());
```
andlar: This doesn't make sense to me. Why are we waiting for a thing and then expecting that thing to be true? If it was going to be false then it'd never pass the "waitUntil" line, so would fail the test with a timeout instead of with a missed expectation.

andlar (dismissed):

athomas-git (dismissed):

andlar (approved):

athomas-git (approved):

PR #985, opened 2021-11-03
OCD-3771: Add RWT collections page

andlar:


src/app/api/collections.jsx
```diff
@@ -0,0 +1,22 @@
...
+const useFetchRealWorldTestingCollection = ({
+  orderBy,
+  pageNumber,
+  pageSize,
+  sortDescending,
+  query,
+}) => {
+  const axios = useAxios();
+  return useQuery(['search/beta', orderBy, pageNumber, pageSize, sortDescending, query], async () => {
+    // const response = await axios.get(`/search/beta?rwtOptions=is_eligible,has_plans_url,has_results_url&${query}&pageNumber=${pageNumber}&pageSize=${pageSize}&orderBy=${orderBy}&sortDescending=${sortDescending}`); // option with no rwt filter
```
tmy1313: Commented code.

tmy1313 (dismissed):

tmy1313 (dismissed):

e2e/pages/collections/real-world-testing/real-world-testing.readonly.spec.js
```diff
@@ -0,0 +1,103 @@
...
+      const searchTerm = 'Veracity';
+      page.searchForText(searchTerm);
+      expect(page.getTableCellText(page.results[0], columnIndex)).toContain(searchTerm);
+    });
+  });
+
+  xdescribe('when clicking on all RWT download button', () => {
+    it('should download a file', () => {
+      page.downloadAllRwtButton.click();
+      const apiFileName = 'APIDocData';
```
ashwinimore, line 93: I understand that this test is marked to skip until the download is fully functional and the filename used here is just a placeholder until the filename is finalized in the implementation ticket, is that correct? if the idea of adding the test now is to enable the test by simply removing x, without making any other changes, then I think adding a filename that is relevant to RWT term is a good idea. I am fine with keeping this as-is and changing it in OCD-3804. I just had that thought when i saw APIDocData as sample text in RWT test

ashwinimore (dismissed):

kekey1 (dismissed):

e2e/pages/collections/real-world-testing/real-world-testing.readonly.spec.js
andlar: I don't think this button is going to exist at all once we get to that other ticket. I'd still like to leave it here as a placeholder for whatever we end up changing this to, but if that's a concern I can certainly remove it instead

athomas-git (dismissed):

e2e/pages/collections/real-world-testing/real-world-testing.readonly.spec.js
ashwinimore: not a concern at all, leaving the test here until it's all finalized sounds good to me.

athomas-git (dismissed):

ashwinimore (dismissed):

kekey1 (dismissed):

ashwinimore (dismissed):

athomas-git (dismissed):

kekey1 (dismissed):

kekey1 (approved):

athomas-git (approved):

tmy1313 (approved):

ashwinimore (approved):

PR #990, opened 2021-11-19
OCD-3809: user management UI tweaks

andlar:


src/app/pages/users/management.component.js
```diff
@@ -66,7 +67,8 @@ const UserManagementComponent = {
         case 'refresh':
           this.networkService.getUsers()
             .then((response) => {
-              that.users = response.users;
+              that.users = response.users
+                .filter((user) => !['ROLE_ACB', 'ROLE_ATL', 'ROLE_DEVELOPER'].includes(user.role));
```
kekey1, line 71: Could both of the filters in this file be "if the user doesn't have any organization" then include them rather than enumerating the roles. Just thinking if that is the intention and if we ever have a future role/organization or change the names of any existing roles then it would still work.

kekey1 (approved):

tmy1313 (approved):

src/app/pages/users/management.component.js
andlar: I didn't want to do that because we have specific pages to look for users in the ROLEs in the list, but we don't have a specific page for anyone else. If we did add some other users to the system who had organizations I'd want them to appear on this page until/unless we created that organization management page

athomas-git (approved):

PR #1003, opened 2022-01-05
OCD-3481 : AQA tests for complaints reporting

palakpatel311:


e2e/components/surveillance/complaints/complaints.readonly.spec.js
```diff
@@ -73,6 +74,70 @@ describe('when logged in as an ACB', () => {
...
+          complaintsComponent.advanceFilterOptions('Anonymous');
+          complaintsComponent.advanceFilterOptions('Government_Entity');
+          complaintsComponent.advanceFilterOptions('Provider');
+          complaintsComponent.advanceFilterOptions('Developer');
+          complaintsComponent.advanceFilterOptions('Third__Party_Organization');
+          complaintsComponent.chooseAdvanceSearchOption('Status')
+          complaintsComponent.advanceFilterOptions('Open');
+          complaintsComponent.waitForUpdatedTableRowCount();
+          for(let i=1; i< hooks.getTableRows().length; i++){
+            expect(hooks.getCellValue(i,6)).toBe('Other - [Please Describe]');
```
ashwinimore: will this test work when the data changes so that there are more closed complaints of Complainant Type 'Other - [Please Describe]' or 'Patient'? I ask this because there are closed complaints for Patient Complainant Type which is one of the applied filter, and when there are more complaints added, the value in row 6 may change

ashwinimore (changes requested):

e2e/components/surveillance/complaints/complaints.readonly.spec.js
palakpatel311: You are right. Updated assertion

ashwinimore (dismissed):

athomas-git (dismissed):

e2e/components/surveillance/complaints/complaints.po.js
```diff
@@ -12,13 +12,41 @@ class ComplaintsComponent {
...
+      informedOnc: '#flag-for-onc-review'
     };
   }
 
   get editButton() {
     return $(this.elements.editButton);
   }
 
+  complaintsBody () {
+    return $('chpl-complaint-view-bridge').$('div').getText();
```
andlar: Is there any other locator that we can use instead of the `-bridge` type? These components are temporary ones that will go away as we transition to React, and I'd rather not rely on them now if we can avoid it

e2e/components/surveillance/complaints/complaints.po.js
```diff
@@ -70,6 +124,33 @@ class ComplaintsComponent {
     $('//span[text()="Delete"]/parent::button').click();
     $('//span[text()="Yes"]/parent::button').click();
   }
+
+  waitForUpdatedTableRowCount () {
+    let start;
+    let next;
+    do {
+      start = $('table').$('tbody').$$('tr').length;
+      browser.pause(2000);
```
andlar: Is there any other way of doing this pause? Using the `browser.waitUntil` function or something? This could get stuck in an infinite loop otherwise

e2e/components/surveillance/complaints/complaints.readonly.spec.js
```diff
@@ -73,6 +74,70 @@ describe('when logged in as an ACB', () => {
         const button = complaintsComponent.downloadResultsButton;
         expect(button.isExisting()).toBe(false);
       });
+
+      describe('when searching complaints by ONC-ACB Complaint ID, or Associated Criteria', () => {
+        const oncAcbComplaintID = 'SC-000093';
+        const oncAcbComplaintID_IDX = '4';
```
andlar: There's not much point in putting these `const`s outside of the `it` they're used in, as these aren't used in more than one it, as opposed to the `FIRST_ROW` constant above

e2e/components/surveillance/complaints/complaints.po.js
```diff
@@ -30,6 +58,22 @@ class ComplaintsComponent {
     $(`//li[text()="${fields.type}"]`).click();
   }
 
+  setOptional(fields) {
```
andlar: I had to go look at where this function is used to figure out what it does. Can we name it something that is more descriptive of what it does? Maybe `setOptionalFields`?

e2e/pages/organizations/onc-acbs-atls/atls.readonly.spec.js
```diff
@@ -46,7 +46,7 @@ describe('the ONC-ATL Management page', () => {
     }
   });
 
-  describe('when impersonating as UL', () => {
+  xdescribe('when impersonating as UL', () => {
```
andlar: Does this need to be ignored? Can we impersonate a different user to keep these tests going?

andlar (changes requested):

e2e/pages/organizations/onc-acbs-atls/atls.readonly.spec.js
palakpatel311: I had to change bunch of things on this test to make it run as we dont show ROLE_ATLs under user management anymore. But it works now and removed skip label.

e2e/pages/organizations/onc-acbs-atls/atls.spec.js
```diff
@@ -3,7 +3,7 @@ import Hooks from '../../../utilities/hooks';
 import AddressComponent from '../../../components/address/address.po';
 import LoginComponent from '../../../components/login/login.po';
 import ToastComponent from '../../../components/toast/toast.po';
-import UsersPage from '../../users/user.po';
+import UsersPage from '../../users/users.po';
```
andlar: If we're not on the Users Page, then we should be using the Users component, not the Users Page. This import should be to the components/user/user.po.js file. The component that's being imported has a "pageTitle" element that doesn't exist on the organizations pages, which is an indication the wrong page object is being used

e2e/pages/users/users.po.js
```diff
@@ -13,6 +13,10 @@ class UsersPage {
   getUsers() {
     return $$(this.elements.users);
   }
+
+  impersonateUser(name) {
+    $(`//*[contains(@title, 'Impersonate ${name}')]`).click();
```
andlar: This one _could_ be here, because we _can_ impersonate from the Users page, but it also needs to be on the users component for use when the tests are on an organization page (ACB, ATL, Developer)

palakpatel311: I am hesitant of duplicating same method in both and leaning towards only keeping in UserComponent.

andlar (dismissed):

athomas-git (dismissed):

ashwinimore (dismissed):

athomas-git (approved):

ashwinimore (approved):

andlar (approved):

PR #1027, opened 2022-03-09
OCD-3881 - Add Attestation "Requirement" and "Non-Conformity Type" values

tmy1313:


kekey1 (dismissed):

src/app/components/listing/details/surveillance/requirement/edit.html
```diff
@@ -50,6 +50,16 @@ <h4 class="modal-title" id="edit-requirement-label"><span ng-if="$ctrl.workType
       </div>
     </div>
 
+    <div class="form-group" ng-if="$ctrl.requirement.type.name === 'Attestations Submission'">
+      <label for="requirement-transparency" class="col-sm-3">Requirement <span class="text-danger">*</span></label>
+      <div class="col-sm-9">
+        <select class="input-sm form-control" ng-model="$ctrl.requirement.requirement" id="requirement-rwt" name="requirement"
```
andlar: This "id" is "requirement-rwt", which doesn't match the name, nor does it match the "for" in the label two lines up. These should all be the same, though I believe "name" is usually camelCase, while the ID is snake-case

andlar (changes requested):

src/app/components/listing/details/surveillance/requirement/edit.html
tmy1313: The "id" is definitely wrong.  I think the "name" is correct based on the existing pattern - we only show one of these label/select combos.  There is a block displayed based on what "requirementType" selected (capability, transparency, rwt, attestations).  By all of them using the name "requirement", the selected value maps correctly to the JS code.

andlar (approved):

e2e/config/urlsEXAMPLE.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
-  devURL: 'DEV URL',
-  qaURL: 'QA URL',
-  stageURL: 'STAGE URL',
-  ff4jURL: 'URL'
+  devURL: 'http://www.example.com/',
+  qaURL: 'http://www.example.com/',
+  stageURL: 'http://www.example.com/',
+  ff4jURL: 'rest/endpoint?api_key=1234abcd'
```
ashwinimore, line 5: As line-78 suggests in README above, should this be in a copy e2e/config/urls.js (which is in gitignore, to exclude from commit)?

kekey1 (approved):

e2e/config/urlsEXAMPLE.js
andlar: This can't be in the urls.js file, because that file is where our _real_ data goes. This is the example file, and the change here was made to show what the actual format of these values should be. The previous version made it hard to tell if a given URL should have a `/` at the end, or have the `http` or `https` or something at the beginning. For the FF4j one, there wasn't anything that made it clear there should / shouldn't be a `/` at the beginning either

ashwinimore (approved):

athomas-git (approved):

PR #1036, opened 2022-04-05
OCD-3855: add aqa tests for Announcements

ashwinimore:


andlar (approved):
It's not a lot of tests, but it is at least one, and assuming it works it's a perfect start for adding more, if we choose to. Thanks!

athomas-git (approved):

PR #1048, opened 2022-05-17
OCD-3941: update test to add new report job in the list of expected jobs for ONC Staff users

ashwinimore:


e2e/pages/administration/jobs/scheduled.readonly.spec.js
```diff
@@ -30,6 +30,7 @@ describe('the Scheduled Jobs page', () => {
         'All Broken Surveillance Rules Report',
         'Cures Statistics Email',
         'Developer Access Report',
+        'Developer Attestations Check-in Report Email',
```
andlar, line 33: This was a job that changed an existing report, right? So should one be removed too? Or something else?

andlar (dismissed):

e2e/pages/listing/listing.po.js
```diff
@@ -60,6 +60,7 @@ class ListingPage {
   listingHistoryModalRows() {
     return $(this.elements.listingHistoryTitle)
       .parentElement()
+      .$('div.MuiDialogContent-root.MuiDialogContent-dividers')
```
andlar, line 63: Why is this here?

I generally try to avoid locating based on MUI stuff, because we have no control over the class names, and I'd expect any MUI version upgrade to break a lot of things

ashwinimore: without that the test fails to locate the table element in product history modal. The test has been failing for a long time. If you strongly recommend not using it, then I can look for other way to locate it. The table does not have unique id defined.

ashwinimore: or I can revert the change for now and maybe rewrite the test in next PR...open to suggestions

andlar: I don't like this, but if you say it works then I'll call it a win. If the structure changes or we change the content that shows up in this thing at some point, maybe we can revisit, but for now, and the sake of time, It's okay

andlar (approved):

athomas-git (approved):

PR #1055, opened 2022-06-16
update to chrome driver version

athomas-git:
updated chrome driver version to 101.0.0


andlar (changes requested):
When the package.json file is updated, the yarn.lock file needs to be updated as well, and the changes to that file should be saved in git as well

ashwinimore (approved):

andlar (approved):

PR #1065, opened 2022-07-12
OCD-3965 : test to add new user job for ONC STAFF user

athomas-git:
OCD-3965 : test to add new user job for ONC STAFF user

Fix for existing aqa failure

### DIFF
--- a/e2e/components/surveillance/complaints/complaints.readonly.spec.js
+++ b/e2e/components/surveillance/complaints/complaints.readonly.spec.js
@@ -40,7 +40,7 @@ describe('when logged in as an ADMIN', () => {
 
       it('should have a button to download results', () => {
         const button = complaintsComponent.downloadResultsButton;
-        expect(button.getText()).toBe('Download all complaints');
+        expect(button.getText()).toBe('DOWNLOAD ALL COMPLAINTS');
         expect(button).toBeClickable();
       });
     });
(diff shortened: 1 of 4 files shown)


kekey1 (dismissed):

andlar (dismissed):

ashwinimore (dismissed):

e2e/components/surveillance/complaints/complaints.readonly.spec.js
```diff
@@ -40,7 +40,7 @@ describe('when logged in as an ADMIN', () => {
 
       it('should have a button to download results', () => {
         const button = complaintsComponent.downloadResultsButton;
-        expect(button.getText()).toBe('Download all complaints');
+        expect(button.getText()).toBe('DOWNLOAD ALL COMPLAINTS');
```
andlar: I don't like these tests to expect the text to be in all caps. They're not all caps in the code, and they're not all caps in a screen-reader mode. They only look all caps due to css, so if a test were to execute on a css-less page it wouldn't work. That said, I understand why this change was made. What I've done in other places where the css capitalization gets in the way of the text matching is to use RegExps to do a match in a case-insensitive way. 

The `e2e/components/login/login.readonly.spec.js` tests have some examples

```suggestion
        expect(/Download all complaints/i.test(button.getText())).toBe(true);
```

andlar (dismissed):

kekey1 (approved):

ashwinimore (approved):

andlar (approved):

PR #1113, opened 2022-11-03
OCD-3955 - Allow admin to cancel a scheduled system job

tmy1313:


src/app/pages/administration/jobs/system-triggers-view.jsx
```diff
@@ -41,84 +45,139 @@ const useStyles = makeStyles({
...
+          aria-label={`Delete Job ${item.name}`}
+        >
+          <DeleteIcon />
+        </IconButton>
+      );
+    }
+    return null;
+  };
+
   useEffect(() => {
```
andlar: There's a convention I've been using of keeping the "useEffect" blocks immediately after the "initialization" lines, where the useState, useContext, const { stuff } = props, and other "start" stuff goes. It's not iron-clad, but it helps me see what things are happening when the component "loads". In this case, if you did that, you'd get an eslinting error about using a function before it's defined (for the "getAction" function), so what I usually do is `let getAction;` at the end of the initialization block is (right before the useStyles()) call, and then define the function definition later in the file, after this useEffect block

src/app/pages/administration/jobs/system-triggers-view.jsx
```diff
@@ -41,84 +45,139 @@ const useStyles = makeStyles({
...
-                        </TableRow>
-                      ))}
-                  </TableBody>
-                </Table>
-              </TableContainer>
-            )}
-        </>
-      </CardContent>
-    </Card>
+    <div>
```
andlar: This should be a React.Fragment instead of a <div> element. The reason is that DIVs have layout meaning, while the fragment things are just "grouping" elements

```suggestion
    <>
```

src/app/pages/administration/jobs/system-triggers-view.jsx
```diff
@@ -23,6 +26,7 @@ const headers = [
   { property: 'description', text: 'Description' },
   { property: 'nextRunDate', text: 'Next Run Date' },
   { property: 'triggerScheduleType', text: 'Trigger Schedule Type' },
+  { property: 'actions', text: 'Actions' },
```
andlar: The "User Scheduled Jobs" table has this same Actions column, but I believe has the "invisible = true" annotation. We need the text to exist for 508 purposes, but it doesn't have to be visible on the page, and generally we don't show it

```suggestion
  { property: 'actions', text: 'Actions', invisible: true },
```

src/app/pages/administration/jobs/system-triggers-view.jsx
```diff
@@ -41,84 +45,139 @@ const useStyles = makeStyles({
...
+  const [pendingMessage, setPendingMessage] = useState('');
   const classes = useStyles();
 
+  const confirmDelete = (item) => {
+    setIsConfirming(true);
+    setPendingAction({
+      action: 'delete',
+      payload: {
+        name: item.triggerName,
+        group: item.triggerGroup,
```
andlar, line 72: Recommend using the spread syntax here, as that's the syntax you're using in the other payload object on the user triggers job. The alternative could be to switch the other one to be explicit, but I've had that come back to bite me in the past when the object structure changed, or an element I needed hadn't been explicitly included
```suggestion
   ...item,
```

andlar (commented):
This is a great PR. There are a couple of minor things I'd like you to look at, and think about my suggestions, but I don't have any show-stopper change requests

src/app/pages/administration/jobs/system-triggers-view.jsx
tmy1313: I had to add an ESLint disable statement to get this to work.  What am I missing?

andlar: That's odd. I've seen that error before, but only in my .test files, where I'm mocking some elements I don't care about. I know I've created React Nodes via functions before; maybe move the function outside the component definition? As in, usually just below the block of "const useStyles = makeStyles" stuff I'll add functions that can generate Nodes without depending on anything other than what's passed in.

The other option (and I don't know if this would work or not, but might be worth trying) would be to reverse the logic of the `if` in this function. Basically, `return null` immediately if the trigger isn't the right type. I wonder if, by having the `return (<IconButton` stuff at the "function's root level" it might take the function name as the display-name the way we'd expect. So, in pseudo code:

```
getAction = (item) => {
if (item.triggerType !== 'ONE_TIME") { return null;}
return (
<Icon Button
...
);
};
```

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1167, opened 2023-03-30
OCD-3673: Add aqa tests for system maintenance - standards 

ashwinimore:


e2e/components/standards/qms-standard/qms-standard.spec.js
```diff
@@ -34,14 +34,25 @@ describe('the qms standard component', () => {
...
       await expect(await (await component.dataTable).getText()).toContain(name);
+      await component.editQmsStandard(name);
+      await (await component.name).clearValue();
+      await (await component.name).addValue(newName);
+      await action.save();
+      await expect(await (await component.dataTable).getText()).toContain(newName);
+      await component.editQmsStandard(newName);
+      await action.delete();
+      await action.clickYesToConfirm();
+      await (browser.waitUntil(async () => component.qmsDataAvailable()));
```
andlar: Can we add, or change, this to use a similar action as the line above where we wait for the number of rows in the table to have changed back to the initial count? I worry that just waiting for the table to exist might get us to a state where the table's there, but the data hasn't refreshed yet, which would make the next line fail

e2e/components/standards/qms-standard/qms-standard.po.js
```diff
@@ -28,6 +32,17 @@ class QmsStandardComponent {
            ).$$('tr');
   }
   /* eslint-enable indent */
-}
 
+  async editQmsStandard(standardName) {
+    const rows = await this.getData();
+    await rows.forEach(async (row) => {
```
andlar: While I can see this working, the code has to iterate through all of the rows to find the one it needs, then it picks the cell with the button to edit, right? There's a one-line way of doing that I used in the Developer Join AQA. The line I used was:

```
 $(`td=${developerName}`).parentElement().$('button').click();
```

This is not the async version, but my guess is you could wrap that in nested awaits, like I've done in a bunch of the async files.

The other advantage to a path like this is that if the order of the columns changes, or a new one is added, or whatever, the locator doesn't have to move. For that matter, this entire `editSomething` function might be able to be part of the system-maintenance PO, reducing the need to duplicate this function

e2e/components/standards/qms-standard/qms-standard.po.js
```diff
@@ -29,5 +29,4 @@ class QmsStandardComponent {
   }
   /* eslint-enable indent */
 }
-
```
andlar, line 32: Not sure why this line was removed. All the other PO files have a blank line before the final export line

e2e/components/standards/svap/svap.spec.js
```diff
@@ -45,6 +47,16 @@ describe('the svap component', () => {
       await action.save();
       await browser.waitUntil(async () => (await component.getSvaps()).length > initialCount);
       await expect(await (await component.svapTable).getText()).toContain(version);
+      await page.editStandards(version);
+      await (await component.version).clearValue();
+      await (await component.version).setValue(newVersion);
+      await action.save();
+      await expect(await (await component.svapTable).getText()).toContain(newVersion);
```
andlar: I think we need a "wait" in here before expecting our results. I agree that this probably works, but as a general practice we should wait for something to change, then expect our new something to happen. In the add/delete cases we wait for the table length to change. I recognize that won't work here, but maybe we could wait for the old name to not be found in the table? 

Maybe something like:

```suggestion
      await action.save();
      await browser.waitUntil(async () => (await component.svapTable).getText().not.toContain(version);
      await expect(await (await component.svapTable).getText()).toContain(newVersion);
```

e2e/pages/administration/system-maintenance/system-maintenance.po.js
```diff
@@ -24,6 +24,10 @@ class SystemMaintenancePage {
   async canNavigate(target) {
     return (await $(this.elements.navigationButton(target))).isClickable();
   }
+
+  async editStandards(identifier) {
```
andlar: I love this genericizing action, and we need to take it one step farther, as not all the things in system-maintenance are standards. Maybe:

```suggestion
  async editItem(identifier) {
```

andlar (dismissed):

andlar (dismissed):

athomas-git (dismissed):

athomas-git (dismissed):

e2e/pages/administration/system-maintenance/system-maintenance.po.js
```diff
@@ -21,9 +24,46 @@ class SystemMaintenancePage {
     await (await $(this.elements.navigationButton(target))).click();
   }
 
+  get addButton() {
+    return $(this.elements.addButton);
```
andlar, line 29: I know the "addButton" has to be different on each of the descendent pages, so whatever "addButton" locator this page uses probably won't work, but it still needs to be in the elements object. Right now there is no "addButton" in `this.elements`, which would mean that this would be undefined. I could maybe see an argument for leaving it that way, but that opens up a lot of other topics, and the simple fix would be to add a locator to this file that covers it

e2e/pages/administration/system-maintenance/system-maintenance.po.js
```diff
@@ -21,9 +24,46 @@ class SystemMaintenancePage {
...
+  async getData() {
+    return (await
+            (await
+             this.dataTable
+            ).$('tbody')
+           ).$$('tr');
+  }
+  /* eslint-enable indent */
+
+  async editItem(identifier) {
```
andlar: There are two of these functions in this file

andlar (changes requested):
We are really close. Two small things to address and I might be done with comments

e2e/pages/administration/system-maintenance/system-maintenance.po.js
ashwinimore: agree. I missed that, pushed update for this

e2e/pages/administration/system-maintenance/system-maintenance.po.js
ashwinimore: agree, that happened in error while consolidating files. Removed duplicate in new commit

andlar (approved):

athomas-git (approved):

PR #1176, opened 2023-04-27
OCD 4139 : AQA tests for new SVAP Information collection page

athomas-git:
OCD 4139 : AQA tests for new SVAP Information collection page


e2e/pages/collections/svap-information/svap-information.po.js
```diff
@@ -0,0 +1,24 @@
+import { open as openPage } from '../../../utilities/hooks.async';
+import CollectionPage from '../collection.po';
+
+class SvapInformationPage extends CollectionPage {
+  constructor() {
+    super();
+    this.elements = {
+      ...this.elements,
+      header: 'h1=SVAP Information',
+      downloadSVAPInformation: '#download-listings',
```
andlar: I don't think this element is needed. If this is the main download button it should be in the CollectionPage PO, but I also don't think we're doing any tests with that button. The API Info page has a page specific button, but there isn't one for SVAP

e2e/pages/collections/svap-information/svap-information.readonly.spec.js
```diff
@@ -0,0 +1,98 @@
...
+        await expect(countAfter).toBeLessThan(countBefore);
+      });
+
+      it('should filter on Certification Status', async () => {
+        await page.removeFilter('Certification Status', 'Active');
+        countAfter = await page.getTotalResultCount();
+        await expect(countAfter).toBeLessThan(countBefore);
+      });
+
+      it('should filter on has any SVAP', async () => {
```
andlar: I would suggest using one of the other SVAP filters as a test. I don't know if we're going to keep this one as an available filter at all, and even if we do, the actual SVAP filters are more likely something end users will use, so that's what we should test against

andlar (changes requested):

e2e/pages/collections/svap-information/svap-information.readonly.spec.js
```diff
@@ -0,0 +1,94 @@
+import SvapInformationPage from './svap-information.po';
+
+const path = require('path');
+const fs = require('fs');
+
+const config = require('../../../config/mainConfig');
```
andlar: Are these doing anything? I think they're usually imported when there's a file download test, but I don't think we need those in this spec

andlar (dismissed):

andlar (dismissed):

andlar (dismissed):

andlar (dismissed):

e2e/pages/collections/svap-information/svap-information.readonly.spec.js
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  //below text will be updated, once PO confirmed exact text on svap information page
+  it('should have body text', async () => {
+    await expect(await page.getBodyText()).toContain('This list includes all 2015 Edition, including Cures Update, health IT products that have been INSERT TEXT HERE');
+    await expect(await page.getBodyText()).toContain('Please note that by default, only listings that are active or suspended are shown in the search results.');
+  });
+
+  it('should have table headers in a defined order', async () => {
+    const expectedHeaders = ['CHPL ID', 'Certification Edition', 'Developer', 'Product', 'Version', 'Status', 'SVAP Information', 'SVAP Notice URL', 'Actions'];
```
andlar:
```suggestion
    const expectedHeaders = ['CHPL ID', 'Certification Edition', 'Developer', 'Product', 'Version', 'Status', 'SVAP Information', 'SVAP Notice', 'Actions'];
```

Changed the header after talking with the PO. This change should match the updates

andlar (dismissed):

ashwinimore (dismissed):

andlar (approved):

ashwinimore (approved):

PR #1218, opened 2023-09-01
OCD-4348: No longer need editionless file with XX code

kekey1:
The '15' code will work in all the other existing AQA files and whether it is used or not when parsing the listing is based on the state of the editionless flag.

[#OCD-4348]


andlar (dismissed):

tmy1313 (dismissed):

src/app/pages/listing/edit.component.js
```diff
@@ -58,7 +58,7 @@ const ListingEditPageComponent = {
     }
 
     prepareResources() {
-      this.resources.testStandards.data = this.resources.testStandards.data.filter((item) => !item.year || item.year === this.listingDetails.edition.name);
+      this.resources.testStandards.data = this.resources.testStandards.data.filter((item) => !item.year || item.year === this.listingDetails.edition?.name);
```
andlar, line 61: If there is a Test Standard that has a year, and we're on a Listing that's editionless, that TS won't be allowed on the Listing. I don't know if that's right. I would assume that any 2015 TS should be valid for editionless Listings. I don't know what should be the case with TS from 2011 or 2014, assuming there are any.

andlar (approved):

tmy1313 (approved):

PR #1222, opened 2023-09-20
OCD-4340: convert charts AQA to be async

ashwinimore:


e2e/pages/charts/charts.readonly.spec.js
```diff
@@ -11,119 +11,121 @@ describe('the charts page', () => {
...
+    it('should have the right options in the "view certification criteria" dropdown', async () => {
       const expected = new Set(['All', '2015', '2015 Cures Update']);
-      expect(page.programTypeDropdownOptions.length).toBe(expected.size);
-      const options = [...new Set(page.programTypeDropdownOptions.map((item) => item.getText()))];
-      options.forEach((option) => {
-        expect(expected.has(option)).toBe(true, `did not find expected option: "${option}"`);
-      });
+      await expect(page.programTypeDropdownOptions.length).toBe(expected.size);
+      const options = [...new Set(await Promise.all(page.programTypeDropdownOptions.map(async item => await item.getText())))];
+      for (const option of options) {
```
andlar: Does the `options.forEach` construct not work here? I've been trying to get away from `for` loops, but if there's a reason that can't happen I'm fine with it

andlar (dismissed):

e2e/pages/charts/charts.readonly.spec.js
ashwinimore: No it didn't work. I am working on getting it to work in some other way, suggestions are welcome

ashwinimore: what I meant was that the above snippet (in addition to the options.forEach construct) did not work. I am going to have to change it with something that works if that part of the test is to stay

athomas-git (approved):

andlar (approved):

PR #1259, opened 2023-12-12
[#OCD-4294] ui!: Remove references to XML/XSD files from API Page

andlar:


tmy1313 (dismissed):

src/app/pages/resources/api/api.jsx
```diff
@@ -101,28 +95,24 @@ function ChplResourcesApi() {
...
+      '2011 edition products': { data: `${API}/download/2011?api_key=${getApiKey()}&format=json`, label: '2011' },
     };
     setFiles(data);
     setDownloadOptions(() => allOptions);
   }, [API, getApiKey, getToken]);
 
   const downloadFile = (type) => {
     if (selectedOption) {
-      $analytics.eventTrack(`Download CHPL${type === 'definition' ? ' Definition' : ''}`, { category: 'Download CHPL', label: files[selectedOption].label });
+      $analytics.eventTrack('Download CHPL', { category: 'Download CHPL', label: files[selectedOption].label });
```
kekey1, line 115: Since we aren't offering definition files anymore, I feel like there should be an update to the text on the left side of this page to direct the user to the "Schemas" area at the waaaaayyyy bottom of the Swagger UI.

src/app/pages/resources/api/api.jsx
```diff
@@ -161,23 +144,7 @@ function ChplResourcesApi() {
...
-                      The Inactive products summary file is updated nightly.
-                    </li>
-                    <li>
-                      The 2014 Edition Products file and the 2011 Edition Products file are updated quarterly.
-                    </li>
-                  </ul>
-                </li>
-                <li>
-                  <Typography gutterBottom><strong>Certified Products (XML):</strong></Typography>
+                  <Typography gutterBottom><strong>Certified Products:</strong></Typography>
```
kekey1: I don't have a good place to leave this comment but I think you also need to update text from "Download the CHPL" where it says "JSON and XML files have been moved to the [CHPL API page](https://chpl-dev.healthit.gov/#/resources/api)." Maybe that whole alert can be removed. As a person who totally forgot that we only moved the JSON and XML resources, I was really confused when looking at the CHPL API page. I wanted to download a CSV and I couldn't, but forgot that I had to do that from a different page. The functionality seems really oddly broken up. Downloading the JSON is also "Downloading the CHPL" and I wonder if the "Download the CHPL" page could let you download ALL formats (JSON and CSV) whereas the API page could let you download just the JSON since we guessed developers care most about that. I don't know. I couldn't find anywhere on the "Download the CHPL" page where it told me what format of file I would be getting and that also was confusing.

kekey1 (changes requested):

src/app/pages/resources/api/api.jsx
kekey1: It's equally confusing for the updated page to no longer tell you anywhere what file format you are downloading. I agree that not every item in the dropdown needs "(JSON)" next to it, but it should probably be in some/one place on the page. What about the button? "Download JSON" or "Download Data as JSON" or even on the left side in the description.

kekey1 (dismissed):

tmy1313 (dismissed):

kekey1 (approved):

tmy1313 (approved):

PR #1268, opened 2023-12-18
OCD-4317 - Update Charts to account for Editionless

tmy1313:


kekey1 (dismissed):

src/app/pages/charts/charts.scss
```diff
@@ -1,5 +1,5 @@
 .bar-chart {
-    height: 1800px;
+    height: 800px;
```
andlar: I think this class is unused. It should be removed, if possible.

Probably the .pie-chart class as well

andlar (dismissed):

kekey1 (approved):

andlar (approved):

PR #1275, opened 2024-01-12
OCD-4449

mattystank:
- Add CTA with editionLessIsOn on home page
- Change icon color to darker blue to match mockup

### DIFF
--- a/src/app/pages/collections/search/landing-page.jsx
+++ b/src/app/pages/collections/search/landing-page.jsx
@@ -180,7 +180,8 @@ function ChplLandingPage() {
               <ChplFilterSearchTerm />
               <ChplFilterQuickFilters />
             </Box>
-            { editionlessIsOn ? (
+            { editionlessIsOn
+            && (
               <>
                 <Box pb={4}>
                   <Typography className={classes.subHeaders} align="left" component="h2" variant="h2" gutterBottom>
@@ -202,7 +203,7 @@ function ChplLandingPage() {
                   </CardContent>
                 </Card>
               </>
-            ) : null}
+            )}
             <Box pt={6} pb={4}>
               <Typography className={classes.subHeaders} align="left" component="h3" variant="h2" gutterBottom>
                 Use our collections pages to help find a particular category of listings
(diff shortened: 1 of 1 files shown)


src/app/pages/collections/search/landing-page.jsx
```diff
@@ -191,7 +214,7 @@ function ChplLandingPage() {
                   <Card className={classes.collectionsCards}>
                     <CardContent>
                       <Box display="flex" flexDirection="row" gridGap={8}>
-                        <CodeIcon color="primary" />
+                        <CodeIcon style={{ color: theme.palette.primary.dark }} />
```
andlar, line 218: Does this actually work? The color in the palette file is named `primaryDark`

src/app/pages/collections/search/landing-page.jsx
```diff
@@ -180,8 +180,31 @@ function ChplLandingPage() {
               <ChplFilterSearchTerm />
               <ChplFilterQuickFilters />
             </Box>
-            <Box pb={4}>
-              <Typography className={classes.subHeaders} align="left" variant="h2" gutterBottom>
+            { editionlessIsOn ? (
```
andlar: This shouldn't be a ternary. The way this is set up you'll display one thing if the flag is on, and another if it's off, but the way you have it there's nothing that gets displayed when it's off, so this should just be a conditional. Something like this:
```suggestion
            { editionlessIsOn 
            && (
```

src/app/pages/collections/search/landing-page.jsx
```diff
@@ -180,8 +180,31 @@ function ChplLandingPage() {
...
+                        </Typography>
+                        <Typography>
+                          To simplify the Certification Program and support more modular and extensible future updates, the HTI-1 final rule discontinues year-themed editions of certification criteria. This change also supports broader use of certification criteria and standards adopted by ONC for other federal agencies and programs.
+                        </Typography>
+                      </Box>
+                    </Box>
+                  </CardContent>
+                </Card>
+              </>
+            ) : null}
```
andlar: Changing it to a conditional makes this something like:
```suggestion
            )}
```

andlar (changes requested):

andlar (approved):

kekey1 (approved):

PR #1297, opened 2024-03-14
OCD-4532 - Create/invite new ADMIN Cognito users that can login to CHPL

tmy1313:


src/app/pages/registration/register-user.jsx
```diff
@@ -82,6 +92,28 @@ function ChplRegisterUser(props) {
...
+          hash,
+          user: data,
+        };
+        networkService.createInvitedCognitoUser(packet)
+          .then((resp) => {
+            setMessage('Your account has been created.  A one-time password has been emailed to you.')
+            setState('success');
+          }, (error) => {
+            if (error.data.errorMessages) {
+              error.data.errorMessages.forEach((message) => {
```
andlar: If there are more than three messages, sending each one individually as a toast would mean that only the last three would be visible to the user, as the earlier ones get removed once we hit some number. I'd suggest joining these messages with something like `error.data.errorMessages.join('; ')` and then message something like "You have errors: `insert joined messages here`"

src/app/pages/organizations/onc-organizations/onc-organizations.jsx
```diff
@@ -139,7 +139,7 @@ function ChplOncOrganizations() {
         }
         break;
       case 'invite':
-        invite({ role: 'ROLE_ACB', emailAddress: payload.email, permissionObjectId: activeId }, {
+        invite({ role: 'chpl-onc-acb', emailAddress: payload.email, permissionObjectId: activeId }, {
```
andlar: Does this work? Will it work without turning the SSO system on?

src/app/components/user/users.jsx
```diff
@@ -58,16 +60,22 @@ function ChplUsers({ dispatch, roles, users: initialUsers }) {
   const $rootScope = getAngularService('$rootScope');
   const authService = getAngularService('authService');
   const networkService = getAngularService('networkService');
-  const { mutate } = usePutUser();
+  const { mutate } = usePutUser();7
```
andlar: Looks like a typo

src/app/components/user/cognito-user-invite.jsx
```diff
@@ -0,0 +1,133 @@
...
+    onSubmit: () => {
+      invite();
+    },
+    validationSchema,
+    validateOnChange: false,
+    validateOnMount: true,
+  });
+
+  return (
+    <ThemeProvider theme={theme}>
```
andlar: Shouldn't need this ThemeProvider here. This component is loaded in the users.jsx one, which is wrapped in the user-wrapper.jsx, that has the `<AppWrapper>` as it's root, which includes the theme

src/app/components/user/cognito-user-invite.jsx
```diff
@@ -0,0 +1,133 @@
...
+  formik = useFormik({
+    initialValues: {
+      email: '',
+    },
+    onSubmit: () => {
+      invite();
+    },
+    validationSchema,
+    validateOnChange: false,
+    validateOnMount: true,
```
andlar: I've been removing both of these defaults in the newer components. I don't think they provide a good user experience

src/app/components/user/cognito-user-invite.jsx
```diff
@@ -0,0 +1,133 @@
...
+import PersonAddIcon from '@material-ui/icons/PersonAdd';
+import EmailIcon from '@material-ui/icons/Email';
+import { arrayOf, func, string } from 'prop-types';
+import { useFormik } from 'formik';
+import * as yup from 'yup';
+
+import theme from '../../themes/theme';
+import { ChplDialogTitle, ChplTooltip, ChplTextField } from '../util';
+
+const useStyles = makeStyles(() => ({
```
andlar: I don't know why this is the format that we started with, but it's unnecessary. This has less code and still works:

```suggestion
const useStyles = makeStyles({
```

I think one of the parentheses on line 28 will need to be removed too

src/app/components/registration/cognito-user-create.jsx
```diff
@@ -0,0 +1,160 @@
...
+    onSubmit: () => {
+      create();
+    },
+    validationSchema,
+    validateOnChange: false,
+    validateOnMount: true,
+  });
+
+  return (
+    <ThemeProvider theme={theme}>
```
andlar: Same thing with ThemeProvider. This component is in the registerUser.jsx file, which already applies the themes

src/app/components/registration/cognito-user-create.jsx
```diff
@@ -0,0 +1,160 @@
...
+  Paper,
+  ThemeProvider,
+  Typography,
+  makeStyles,
+} from '@material-ui/core';
+import { useFormik } from 'formik';
+import * as yup from 'yup';
+
+import theme from '../../themes/theme';
+import { ChplTextField } from '../util';
```
andlar: We have an updated way of doing local (project specific) invites that removes the "relative directory" issue. This line can be replaced with:

```suggestion
import { ChplTextField } from 'components/util';
```

And if you take out the ThemeProvider I don't think you'll need the theme import either, but if you do:

```
import { theme } from 'themes';
```

kekey1 (dismissed):

src/app/components/registration/cognito-user-create.jsx
```diff
@@ -0,0 +1,159 @@
...
+    onSubmit: () => {
+      create();
+    },
+    validationSchema,
+    validateOnChange: false,
+    validateOnMount: true,
+  });
+
+  return (
+    <>
```
andlar: Shouldn't need this line. We use these when we need to group elements, but the only element inside this is a `<Paper>`, so this line can be removed

src/app/components/registration/cognito-user-create.jsx
```diff
@@ -0,0 +1,159 @@
+import React, { useContext, useState } from 'react';
+import {
+  func,
+} from 'prop-types';
+import {
+  Button,
+  Paper,
+  ThemeProvider,
```
andlar: I don't think this is being used

src/app/components/user/cognito-user-invite.jsx
```diff
@@ -0,0 +1,128 @@
...
+  DialogActions,
+  DialogContent,
+  makeStyles,
+} from '@material-ui/core';
+import PersonAddIcon from '@material-ui/icons/PersonAdd';
+import EmailIcon from '@material-ui/icons/Email';
+import { useFormik } from 'formik';
+import * as yup from 'yup';
+
+import theme from '../../themes/theme';
```
andlar: I don't think this is being used

src/app/components/user/cognito-user-invite.jsx
```diff
@@ -0,0 +1,128 @@
...
+import { ChplDialogTitle, ChplTooltip, ChplTextField } from '../util';
```
andlar: Should use the absolute import
```suggestion
import { ChplDialogTitle, ChplTooltip, ChplTextField } from 'components/util';
```

src/app/components/user/cognito-user-invite.jsx
```diff
@@ -0,0 +1,128 @@
...
+});
+
+const validationSchema = yup.object({
+  email: yup.string()
+    .required('Email is required')
+    .email('Enter a valid Email'),
+});
+
+function ChplCognitoUserInvite(props) {
+  const [open, setOpen] = React.useState(false);
```
andlar:
```suggestion
  const [open, setOpen] = useState(false);
```

You're already importing the `useState`. No need to prefix it with `React`, and we don't do that anywhere else

andlar (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1310, opened 2024-04-18
OCD-4494 - View Cognito user information

tmy1313:


src/app/components/user/cognito-user-view.jsx
```diff
@@ -0,0 +1,127 @@
...
+  CardHeader,
+  ThemeProvider,
+  Typography,
+  makeStyles,
+} from '@material-ui/core';
+
+import theme from '../../themes/theme';
+import {
+  user as userPropType,
+} from '../../shared/prop-types';
```
andlar: A while ago I figured out how to avoid relative path imports, which helps when moving components around, as the relativity would have to change. I've made those changes to all the files I've found, and don't use relative path imports in new ones. Also, I tend to alphabetize the CHPL imports based on the "from" value, so this block would be:
```suggestion
import { user as userPropType } from 'shared/prop-types';
import { theme} from 'themes';
```

src/app/components/user/cognito-user-view.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+const useStyles = makeStyles(() => ({
```
andlar: The extra function here isn't necessary. It started out that way, but I've been trying to remove it everywhere I find it:

```suggestion
const useStyles = makeStyles({
```

Then end of the makeStyles will need a `)` removed, I believe

src/app/components/user/cognito-user-view.jsx
```diff
@@ -0,0 +1,127 @@
...
+  content: {
+    //display: 'grid',
```
andlar: Debug code?

src/app/components/user/cognito-user-view.jsx
```diff
@@ -0,0 +1,127 @@
...
+    //},
+  },
+  userCard: {
+    display: 'grid',
+    gridTemplateRows: '64px auto 50px',
+  },
+}));
+
+function ChplCognitoUserView(props) {
+  /* eslint-disable react/destructuring-assignment */
```
andlar: We aren't doing any destructuring in this block; the eslint commands don't need to be here

src/app/components/user/users.jsx
```diff
@@ -140,6 +141,26 @@ function ChplUsers({ dispatch, roles, groupNames, users: initialUsers }) {
     }
   };
 
+  const displayUser = (user) => {
+    if (!isOn) { return null; }
```
andlar: Is this line still needed? The API is returning the right list of users, so I don't know why the UI would need to know what the flag state is, especially with the checks you're making in the next few lines

src/app/pages/organizations/onc-organizations/onc-organizations.jsx
```diff
@@ -153,7 +153,6 @@ function ChplOncOrganizations() {
             });
           },
         });
-        break;
```
andlar: Removing this would mean a cognito-invite also invites for a chpl-user, right? Is that the desired behavior?

src/app/components/user/cognito-user-view.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+export default ChplCognitoUserView;
+
+ChplCognitoUserView.propTypes = {
+  user: userPropType.isRequired,
+  dispatch: func,
+};
+
+//ChplCognitoUserView.defaultProps = {
+//  dispatch: () => {},
```
andlar: I thought I already commented on this, but can't find it, so must have missed it. If the `dispatch` function is not required in the propTypes definition, there needs to be a defaultProps value for it. I _thought_ our linter was set up to flag that, but could be misremembering

andlar (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1311, opened 2024-04-23
OCD-4560

mattystank:
Update Compare Widget Category Analytics to ReactGA4

### DIFF
--- a/src/app/components/compare-widget/compare-display.jsx
+++ b/src/app/components/compare-widget/compare-display.jsx
@@ -54,7 +54,7 @@ function ChplCompareDisplay() {
     ReactGA.event({
       category: 'Compare Widget',
       action: 'Compare Listings',
-      label: listings.map((listing) => listing.id).join('&'),
+      label: 'test',
     });
     $location.url(`/compare/${listings.map((listing) => listing.id).join('&')}`);
     $rootScope.$broadcast('HideCompareWidget');
@@ -65,7 +65,8 @@ function ChplCompareDisplay() {
     ReactGA.event({
       category: 'Compare Widget',
       action: 'Remove all Listings',
-    }); $rootScope.$broadcast('compare.removeAll');
+    }); 
+    $rootScope.$broadcast('compare.removeAll');
   };
 
   if (!listings || listings.length === 0) {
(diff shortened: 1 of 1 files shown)


src/app/components/compare-widget/compare-display.jsx
```diff
@@ -44,22 +45,27 @@ const useStyles = makeStyles({
...
   const $rootScope = getAngularService('$rootScope');
   const { listings, removeListing } = useContext(CompareContext);
   const classes = useStyles();
 
   const compareAll = () => {
-    $analytics.eventTrack('Compare Listings', { category: 'Compare Widget' });
+    ReactGA.event({
+      category: 'Compare Widget',
+      action: 'Compare Listings',
+      label: listings.map((listing) => listing.id).join('&'),
```
andlar: Why is this label here? We didn't have it before, and no one was asking for it

src/app/components/compare-widget/compare-display.jsx
```diff
@@ -44,22 +45,27 @@ const useStyles = makeStyles({
...
     $rootScope.$digest();
   };
 
   const removeAll = () => {
-    $analytics.eventTrack('Remove all Listings', { category: 'Compare Widget' });
-    $rootScope.$broadcast('compare.removeAll');
+    ReactGA.event({
+      category: 'Compare Widget',
+      action: 'Remove all Listings',
+    }); $rootScope.$broadcast('compare.removeAll');
```
andlar: The "rootscope.broadcast" part should be on a separate line

andlar (changes requested):

src/app/components/compare-widget/compare-display.jsx
mattystank: I can take it out... This was more of a test to see I got the label in a different category/action

andlar: If you want to keep it as a test, then mark it specifically as a `label: 'test'` or something similar. I don't want to get something in here that we forget to take out, especially if it's something that we don't have a use for

mattystank: Will do !

andlar (approved):

kekey1 (approved):

tmy1313 (approved):

PR #1314, opened 2024-04-26
OCD-4562: Remove sync aqa tests

ashwinimore:


e2e/pages/charts/charts.readonly.spec.js
```diff
@@ -1,18 +1,15 @@
-import Hooks from '../../utilities/hooks';
+import { open } from '../../utilities/hooks.async';
```
andlar: If we're going to remove the synchronous hooks file, we probably don't need the `.async` on the file that remains

e2e/workflowTest/edit-listing-test/edit-basic-listing-data.spec.js
```diff
@@ -1,13 +1,12 @@
 import ListingPage from '../../pages/listing/listing.po';
-import Hooks from '../../utilities/hooks';
+import { open } from '../../utilities/hooks.async';
 import LoginComponent from '../../components/login/login.po';
 import ActionBarComponent from '../../components/action-bar/action-bar.async.po';
-import ToastComponent from '../../components/toast/toast.po';
+import ToastComponent from '../../components/toast/toast.async.po';
```
andlar: Same comment here. If the synchronous toast is gone, then we don't need to call out the fact that the remaining one is async. Also true with the action-bar file

andlar (approved):

athomas-git (approved):

PR #1339, opened 2024-07-02
OCD-4518 - Build "Unique products" report in PowerBI

tmy1313:
OCD-4518

### DIFF
--- a/src/app/pages/charts/charts.component.js
+++ b/src/app/pages/charts/charts.component.js
@@ -14,8 +14,8 @@ export const ChartsComponent = {
       this.chartState = {
         tab: 'product',
       };
-      this.loadCriterionProductCountChart();
       this.loadNonconformityCountChart();
+      this.loadUniquProductsReportUrl();
     }
 
     changeTab (target) {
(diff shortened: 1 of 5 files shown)


kekey1 (dismissed):

src/app/pages/charts/product/product.component.js
```diff
@@ -1,77 +1,26 @@
...
   },
   controller: class ChartsProductComponent {
-    constructor($analytics, $log, utilService) {
+    constructor($analytics, $log, $sce, utilService) {
       'ngInject';
 
       this.$analytics = $analytics;
       this.$log = $log;
+      this.$sce = $sce;
       this.utilService = utilService;
```
andlar: The only one of these four injects being used is the $sce one. The others should be removed. I sometimes keep $log around for ease of debugging, but I don't think that's necessary any more. The `constructor(...){` line will need to change too

src/app/pages/charts/charts.component.js
```diff
@@ -30,10 +30,10 @@ export const ChartsComponent = {
     }
 
     ////////////////////////////////////////////////////////////////////
-
-    loadCriterionProductCountChart () {
+    loadUniquProductsReportUrl () {
       let that = this;
-      this.networkService.getCriterionProductStatistics().then(data => that.criterionProduct = data);
```
andlar, line 36: If this networkService call isn't being used anywhere else, it should be removed from the network.service.js file

andlar (changes requested):

kekey1 (approved):

andlar (approved):

PR #1342, opened 2024-07-15
OCD-4545 = Edit Cognito user information

tmy1313:


src/app/api/users.jsx
```diff
@@ -22,8 +22,19 @@ const usePutUser = () => {
...
+    },
+  });
+};
+
 export {
+  usePutUser, // eslint-disable-line import/prefer-default-export
+  usePutCognitoUser, // eslint-disable-line import/prefer-default-export
   usePostCreateCognitoInvitedUser,
   usePostCreateInvitedUser,
-  usePutUser,
```
andlar: These line don't need the eslint-disable. That's only if there's only a single export. Also, the standard in the API files is to sort the export alphabetically (as well as sorting the functions, but there's no change needed there)

src/app/components/user/cognito-user-edit.jsx
```diff
@@ -0,0 +1,169 @@
...
+} from 'prop-types';
+import {
+  Card,
+  CardHeader,
+  CardContent,
+  FormControlLabel,
+  Switch,
+  ThemeProvider,
+  makeStyles,
+  Typography,
```
andlar: These should be alphabetized

src/app/components/user/cognito-user-edit.jsx
```diff
@@ -0,0 +1,169 @@
...
+} from '@material-ui/core';
+import { useFormik } from 'formik';
+import * as yup from 'yup';
+
+import theme from '../../themes/theme';
+import { ChplTextField } from '../util';
+import { ChplActionBar } from '../action-bar';
+import {
+  user as userPropType,
+} from '../../shared/prop-types';
```
andlar:
```suggestion
import theme from 'themes/theme';
import { ChplTextField } from 'components/util';
import { ChplActionBar } from 'components/action-bar';
import {
  user as userPropType,
} from 'shared/prop-types';
```

We don't do relative imports, except for "local relative". I usually sort imports by the values after the `from` keyword as well

src/app/components/user/cognito-user-edit.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+const useStyles = makeStyles(() => ({
```
andlar:
```suggestion
const useStyles = makeStyles({
```

Making it a function was necessary when we had to pass some sort of value into the makeStyles call. We don't, so having it a function is overkill

src/app/components/user/cognito-user-edit.jsx
```diff
@@ -0,0 +1,169 @@
...
+    onSubmit: () => {
+      save();
+    },
+    validationSchema,
+    validateOnChange: false,
+    validateOnMount: true,
+  });
+
+  return (
+    <ThemeProvider theme={theme}>
```
andlar: If this component is used as a sub-component of any of our React ones, then somewhere up the tree there's already a ThemeProvider that provides the theme. Putting another one here is extraneous

src/app/components/user/cognito-user-view.jsx
```diff
@@ -1,15 +1,23 @@
...
   Card,
+  CardActions,
   CardContent,
   CardHeader,
   ThemeProvider,
   Typography,
   makeStyles,
 } from '@material-ui/core';
 
+import { ChplTooltip } from '../util';
```
andlar:
```suggestion
import { ChplTooltip } from 'components/util';
```

src/app/components/user/users.jsx
```diff
@@ -148,10 +168,10 @@ function ChplUsers({
         <ChplCognitoUserView
           key={userToDisplay.cognitoId}
           user={userToDisplay}
+          dispatch={handleDispatch}
         />
       );
-    }
-    if (userToDisplay.userId) {
+    } if (userToDisplay.userId) {
```
andlar:
```suggestion
    } 
    if (userToDisplay.userId) {
```

src/app/components/user/users.jsx
```diff
@@ -163,6 +183,27 @@ function ChplUsers({
...
+  const displayUserEdit = (userToEdit) => {
+    if (userToEdit.cognitoId) {
+      return (
+        <ChplCognitoUserEdit
+          user={userToEdit}
+          errors={errors}
+          dispatch={handleDispatch}
+        />
+      );
+    } if (userToEdit.userId) {
```
andlar:
```suggestion
    } 
    if (userToEdit.userId) {
```

Either a newline, or an `else if`. Putting the if on the same line as the end of the previous block is odd

kekey1 (dismissed):

andlar (dismissed):

src/app/components/user/cognito-user-view.jsx
```diff
@@ -87,6 +99,21 @@ function ChplCognitoUserView({ user: initialUser }) {
...
+                aria-label={`Edit ${user.fullName}`}
+                onClick={edit}
+              >
+                <EditOutlinedIcon />
+              </Button>
+            </ChplTooltip>
+          </ButtonGroup>
+        </CardActions>
       </Card>
     </ThemeProvider>
```
andlar: This doesn't need to have the `ThemeProvider` tag, as some parent of the component should have it. That should mean that the import of the `theme` can be removed too. I know this isn't specifically related to the changes for this ticket, but could be considered part of the gradual clean-up we're supposed to keep doing

src/app/components/user/cognito-user-view.jsx
```diff
@@ -1,15 +1,22 @@
...
+  CardActions,
   CardContent,
   CardHeader,
-  ThemeProvider,
   Typography,
   makeStyles,
 } from '@material-ui/core';
 
+import { ChplTooltip } from 'components/util';
 import theme from 'themes/theme';
```
andlar: Without the `<ThemeProvider>`, unless the `theme` is used in the styles thing (usually for page size issues) this import probably isn't needed

kekey1 (approved):

andlar (approved):

PR #1355, opened 2024-08-07
OCD-4378 - Implement keep-alive for Cognito users

tmy1313:


src/app/api/axios.jsx
```diff
@@ -1,10 +1,18 @@
...
+  getAccessToken }
+  from 'axios-jwt';
 import { element } from 'prop-types';
 
+import { getAngularService } from 'services/angular-react-helper';
+
 const AxiosContext = createContext();
 
 function AxiosProvider({ children }) {
+  const ngLocalStorage = getAngularService('$localStorage');
```
andlar: In other places where we pull in an angular service, we continue to use the same name for that service, so this line would be:

```suggestion
  const $localStorage = getAngularService('$localStorage');
```

src/app/navigation/navigation.module.js
```diff
@@ -58,21 +60,50 @@ function authInterceptor($log, API, authService, toaster) {
...
+  function getAuthorizationHeader() {
+    const accessToken = $localStorage.jwtToken;
+    if (accessToken && authService.parseJwt(accessToken).exp > moment(new Date().getTime()).unix()) {
+      return $q.when(accessToken);
+    }
+    const { refreshToken } = $localStorage;
+
+    if ($localStorage.currentUser && $localStorage.refreshToken) {
+      const { cognitoId } = $localStorage.currentUser;
+      const nghttp = $injector.get('$http');
```
andlar: Similar comment about naming. I think the real reason I care is that it might make it easier to remove all the AngularJS stuff later if we can easily recognize services that are from there

src/app/pages/administration/administration.module.js
```diff
@@ -31,8 +31,9 @@ angular
      * .interval is how frequently the system refreshes the token
      * values are defined in the webpack.config.js file
      */
-    IdleProvider.idle(60 * MINUTES_UNTIL_IDLE);
-    IdleProvider.timeout(60 * MINUTES_UNTIL_LOGOUT);
+    IdleProvider.idle(60 * 5);
```
andlar: Not sure why the global values were replaced here. Debugging step maybe?

src/app/services/login.service.js
```diff
@@ -139,8 +142,13 @@
     }
 
     function logout() {
+      $injector.get('networkService').logout({
```
andlar: Shouldn't need to do this injection. If you put `networkService` in the function definition on line 9 it will automatically get the service and let this line just be:

```suggestion
      networkService.logout({
```

tmy1313: There is a circular reference that this allows me to get around.  
image

kekey1 (dismissed):

kekey1 (dismissed):

src/app/components/login/cognito-login.jsx
```diff
@@ -28,12 +28,16 @@ const useStyles = makeStyles({
...
+function ChplCognitoLogin({ dispatch, setState, state, uuid }) {
   const [sessionId, setSessionId] = useState('');
-  const [state, setState] = useState(initialState);
   const [userName, setUserName] = useState('');
   const classes = useStyles();
 
+  console.log('state');
+  console.log(state);
+  console.log('setState');
+  console.log(setState);
```
andlar: Probably don't need these

src/app/components/login/cognito-toggle.jsx
```diff
@@ -6,11 +6,14 @@ import {
...
 import PersonIcon from '@material-ui/icons/Person';
 import { func } from 'prop-types';
+import { getAccessToken } from 'axios-jwt';
+import regeneratorRuntime from "regenerator-runtime";
 
 import ChplCognitoLogin from './cognito-login';
 
 import { FlagContext, UserContext } from 'shared/contexts';
 import theme from 'themes/theme';
+import { getIn } from 'yup/lib/util/reach';
```
andlar: Can you move this up to line 11? This is an import from an external library, and all of our external library imports should be at the top

kekey1 (dismissed):

src/app/index-wrapper.jsx
```diff
@@ -12,22 +12,29 @@ function IndexWrapper() {
   const Keepalive = getAngularService('Keepalive');
   const authService = getAngularService('authService');
   const networkService = getAngularService('networkService');
+  const ngLocalStorage = getAngularService('$localStorage');
```
andlar: Our convention has been to use the `$` to indicate AngularJS services that we use in our react stuff. My hope is that it will make it easier to find and remove them once they're no longer necessary

```suggestion
  const $localStorage = getAngularService('$localStorage');
```

kekey1 (approved):

andlar (approved):

PR #1359, opened 2024-08-21
OCD-4653 - Create Developer widgets

tmy1313:


src/app/navigation/navigation-top.html
```diff
@@ -89,6 +89,7 @@
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc'])"><a ui-sref="administration.api-keys"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.api-keys')"></i> API Keys</a></li>
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc', 'chpl-onc-acb'])"><a ui-sref="administration.change-requests"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.change-requests')"></i> Change Requests</a></li>
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc'])"><a ui-sref="administration.system-maintenance"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.system-maintenance')"></i> System Maintenance</a></li>
+              <li ng-if="vm.hasAnyRole(['chpl-admin'])"><a ui-sref="power-bi"><i class="fa fa-caret-right" ng-if="vm.isActive('sys-admin.power-bi')"></i>Power BI Report</a></li>
```
kekey1: This is just a nitpicky thing but I don't like this being called "Power BI" anything. I would like "CHPL By the Numbers" or "Statistics Dashboard" or just "Dashboard" or something. It seems secondary (from a UI point of view) that we are using Power BI to display this dashboard.

kekey1 (dismissed):

src/app/navigation/navigation-top.html
andlar: I like this idea. We don't know what we're actually going to be calling this thing, but why not try to guess. Even just "chplDashboard" or something would match our patterns and could be what we eventually use

kekey1 (dismissed):

src/app/api/reports.jsx
```diff
@@ -0,0 +1,22 @@
+import { useQuery } from 'react-query';
+
+import { useAxios } from './axios';
+
+const useFetchReportUrl = (reportName) => {
+  const axios = useAxios();
+  return useQuery(['report-url', reportName], async () => {
+    if (reportName) {
+      const response = await axios.get(`report-data/${reportName}/url`);
+      console.log('reports.jsx');
```
andlar: `report-url` and `report-data` should match. I don't know if I care which one is used, but they should be the same

src/app/api/reports.jsx
```diff
@@ -0,0 +1,22 @@
...
+      console.log(response.data);
```
andlar: Should remove debug lines

src/app/navigation/navigation-top.html
```diff
@@ -89,6 +89,7 @@
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc'])"><a ui-sref="administration.api-keys"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.api-keys')"></i> API Keys</a></li>
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc', 'chpl-onc-acb'])"><a ui-sref="administration.change-requests"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.change-requests')"></i> Change Requests</a></li>
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc'])"><a ui-sref="administration.system-maintenance"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.system-maintenance')"></i> System Maintenance</a></li>
+              <li ng-if="vm.hasAnyRole(['chpl-admin'])"><a ui-sref="dashboard"><i class="fa fa-caret-right" ng-if="vm.isActive('sys-admin.dashboard')"></i>Dashboard</a></li>
```
andlar, line 92: Issue #1 on the ticket says ONC should have access to these reports, and the ticket description says that as well

src/app/pages/resources/dashboard/dashboard-wrapper.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+import Dashboard from './dashboard';
+
+import AppWrapper from 'app-wrapper';
+
+function DashboardWrapper() {
```
andlar:
```suggestion
function ChplDashboardWrapper() {
```

Our naming convention is that our components all start with Chpl

src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -0,0 +1,155 @@
...
+} from '@material-ui/core';
+import { useFetchReportUrl } from 'api/reports';
+import { palette } from 'themes';
+
+import AssessmentOutlinedIcon from '@material-ui/icons/AssessmentOutlined';
+import DashboardOutlinedIcon from '@material-ui/icons/DashboardOutlined';
+import GroupOutlinedIcon from '@material-ui/icons/GroupOutlined';
+import HighlightOutlinedIcon from '@material-ui/icons/HighlightOutlined';
+
+const useStyles = makeStyles((theme) => ({
```
andlar:
```suggestion
const useStyles = makeStyles({
```

Passing a function into the makeStyles was what we did early on, but is not the standard going forward

src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -0,0 +1,155 @@
...
+  container: {
+    height: '1200px',
+    padding: theme.spacing(8),
```
andlar, line 22: The `theme` variable is not imported anywhere in this file, so this value can't exist. It should be imported in line 12 with the `import { palette, theme }` line

src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -0,0 +1,155 @@
...
+  Button,
+  makeStyles,
+} from '@material-ui/core';
+import { useFetchReportUrl } from 'api/reports';
+import { palette } from 'themes';
+
+import AssessmentOutlinedIcon from '@material-ui/icons/AssessmentOutlined';
+import DashboardOutlinedIcon from '@material-ui/icons/DashboardOutlined';
+import GroupOutlinedIcon from '@material-ui/icons/GroupOutlined';
+import HighlightOutlinedIcon from '@material-ui/icons/HighlightOutlined';
```
andlar, line 14: These are external imports, and should be part of the first set of imports, immediately after the line with `@material-ui/core`

src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -0,0 +1,155 @@
...
+  Box,
+  Card,
+  CardContent,
+  Typography,
+  Container,
+  Button,
+  makeStyles,
+} from '@material-ui/core';
+import { useFetchReportUrl } from 'api/reports';
+import { palette } from 'themes';
```
andlar: These are internal imports (from our project) and should be separated from the external imports by a new line. I _think_ our js-linter should report on that as an issue, but I don't know if you have it set up

src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -0,0 +1,155 @@
...
+                  <CardContent>
+                    <Typography gutterBottom variant="h6">
+                      <b>Welcome to CHPL By The Numbers Dashboard</b>
+                    </Typography>
+                    <Typography gutterBottom>
+                      A dynamic reporting suite powered by PowerBI, providing detailed insights and analytics derived from CHPL data. This tool offers interactive reports with robust click-through capabilities, allowing users to explore and analyze data seamlessly. Each report is designed to be user-friendly, enabling in-depth exploration of key metrics and trends, with the flexibility to dive deeper into the numbers that matter most.
+                    </Typography>
+                    <Box mt={8} mb={4} display="flex" flexDirection="row" flexWrap="wrap" gridGap={32}>
+                      {[
+                        { text: 'Developer Reporting', page: 'developerReport' },
```
andlar: It feels weird having two arrays of the same data defined in different places. Can this set of data be included in the array above that has the icons, and then this `.map` can just be something like `pages.map((item) =>`?

src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -0,0 +1,155 @@
...
+                      width="100%"
+                      height="900px"
+                      src={developerStatisticsReportUrl?.data?.reportUrl}
+                      frameBorder="0"
+                      allowFullScreen
+                    />
+                  </CardContent>
+                </Card>
+              )}
+              {/* Add more conditionally rendered cards here based on `activeCard` value */}
```
andlar, line 141: Do we have a menu with four icons on it, but only one of them works?

src/app/pages/resources/dashboard/index.js
```diff
@@ -0,0 +1,3 @@
+import Dashboard from './dashboard';
```
andlar: We started with using these index.js files, but they just get in the way and don't provide any value. Even in this code you're not actually using this file anywhere

andlar (changes requested):
I noticed the PR is not up to date with staging. Not sure how much that'll change things, but it'll need to be updated before it can be merged, at least

andlar (dismissed):
Leaving some comments open, as they're likely going to be dependent on PO input. Would be okay moving forward in the meantime

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1360, opened 2024-08-26
OCD-4637: Include button to download current CSV of active listings

kekey1:


andlar (dismissed):
Two things that don't affect the PR, really: 1) this PR is not up to date with staging and 2) should probably mention, to the team, that this button will be in more than one place; I think it will also show up on the search results pages, as well as the intended listing details pages. Not sure if that'll be a problem, so should be mentioned at a FR or something

andlar (approved):

PR #1365, opened 2024-09-11
feat: completely remove the keep alive functionality

tmy1313:
feat: completely remove the keep alive functionality


src/app/index-wrapper.jsx
```diff
@@ -9,34 +9,15 @@ import { getAngularService } from 'services/angular-react-helper';
...
-              authService.saveToken(response.token);
-            });
-        } else {
-          Idle.unwatch();
-        }
-      });
-    }
-    return deregisterKeepalive;
-  }, [$rootScope, Idle, Keepalive, authService, networkService]);
+  }, [$rootScope, Idle, authService, networkService]);
```
andlar:
```suggestion
  }, [Idle, authService]);
```

These look like the only services that are still in this function. The other ones should be removed as well

src/app/services/network.service.js
```diff
@@ -478,10 +478,6 @@ export default class NetworkService {
     return this.apiPOST('/cognito/users/invite', invitationObject);
   }
 
-  keepalive() {
-    return this.apiGET('/auth/keep-alive', { ignoreLoadingBar: true });
```
andlar, line 476: If we're truly removing keepalive, we should probably deprecate this endpoint in the API as well

kekey1 (dismissed):

andlar (dismissed):

kekey1 (approved):

PR #1370, opened 2024-09-17
OCD-4668 - Create Surveillance widgets based on Summary Stats

tmy1313:
OCD-4668


src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -51,19 +51,39 @@ const useStyles = makeStyles({
...
-  { text: 'Listing Reporting', page: 'listingReport', icon: <DashboardOutlinedIcon fontSize="large" color="primary" />},
-  { text: 'User Reporting', page: 'userReport', icon: <GroupOutlinedIcon fontSize="large" color="primary" />},
-  { text: 'More to Come', page: '', icon: <HighlightOutlinedIcon fontSize="large" color="primary" />},
-];
-
 function ChplDashboard() {
   const classes = useStyles();
-  const developerStatisticsReportUrl = useFetchReportUrl('DeveloperStatistics');
-  const [activeCard, setActiveCard] = useState('home');
+  const [activeReport, setActiveReport] = useState();
```
andlar: `useState()` should have something inside the parentheses. In this case, I suggest `undefined`, and later, where you're setting the state to `'home'` just change that to be `undefined` as well. When the user is on the dashboard, the active report is actually undefined

src/app/pages/resources/dashboard/dashboard.jsx
```diff
@@ -82,23 +102,26 @@ function ChplDashboard() {
...
+                        { report.text }
+                      </Button>
+                    ))}
                   </Box>
                 </CardContent>
               </Card>
             </Box>
             <Box width="100%">
-              {activeCard === 'home' && (
+              {activeReport == null && (
```
andlar:
```suggestion
              {!activeReport && (
```

Two things: 
1) `undefined !== null`, so I don't think this check works
2) In JS, `==` and `===` are different, but `===` is more strict, and what we should be using everywhere

That said, the easy answer is to just use the `!activeReport` check as in the code suggestion, and that would handle both `null` and `undefined`

andlar (changes requested):
Also, this PR is not up to date with staging

andlar (approved):

kekey1 (approved):

PR #1451, opened 2025-05-12
OCD-4872

mattystank:


src/app/pages/listing/listing.jsx
```diff
@@ -5,21 +5,25 @@ import {
   CircularProgress,
   Container,
   Typography,
+  IconButton,
   makeStyles,
+  Fade,
```
andlar: These ought to be alphabetized

src/app/pages/listing/listing.jsx
```diff
@@ -5,21 +5,25 @@ import {
...
 } from '@material-ui/core';
 import CloudDownloadIcon from '@material-ui/icons/CloudDownload';
 import EditIcon from '@material-ui/icons/Edit';
+import { Star, StarOutline } from '@material-ui/icons';
```
andlar: Is there a reason to import these icons in a different style than every other icon import we have in the system? If so, that's fine, and we could migrate to that choice, but if that's the case then the imports on lines 12-13 should be updated to match the new format

src/app/pages/listing/listing.jsx
```diff
@@ -5,21 +5,25 @@ import {
...
 import { number, oneOfType, string } from 'prop-types';
 
 import ChplListingHistory from './history/listing-history';
-
+import ChplTooltip from 'components/util/chpl-tooltip';
```
andlar: The "application imports" (things that are imports from our own codebase, but don't start with `./`) are alphabetized by the "from" value in all of our other files. This line should be between the ChplListingView and the getAngularService import

src/app/pages/listing/listing.jsx
```diff
@@ -136,6 +145,18 @@ function ChplListingPage({ id }) {
...
+      const isFavorited = prevFavorites.some((fav) => fav && fav.id === listing?.id); // Add null/undefined checks
+      if (isFavorited) {
+        return prevFavorites.filter((fav) => fav && fav.id !== listing?.id); // Add null/undefined checks
+      }
+      return [...prevFavorites, listing]; // Add listing to favorites
+    });
+  };
+
+  const isFavorited = favorites.some((fav) => fav && fav.id === listing?.id); // Add null/undefined checks
+
```
andlar: Lines 150 and 158 look like duplicate code. Is there any way to reduce that duplication? Could you use the function you generated in line 158 inside of the toggleFavorite function?

Also, I'm not sure about the comments. I don't know if you still want to "add null/undefined checks" or not, so I'm not sure if these functions are "complete" yet

src/app/pages/listing/listing.jsx
```diff
@@ -180,12 +201,31 @@ function ChplListingPage({ id }) {
...
+                  >
+                    <IconButton
+                      onClick={toggleFavorite}
+                      style={{ color: '#e3bf00' }}
+                      aria-label={isFavorited ? 'Unfavorite' : 'Favorite'}
+                    >
+                      {isFavorited ? <Star /> : <StarOutline />}
+                    </IconButton>
+                  </ChplTooltip>
+                  <Fade in={isFavorited} timeout={{ enter: 500, exit: 500 }}>
```
andlar, line 226: We don't have a lot of animation in CHPL. Is this something we're going to be adding more generally, or is this a one-off kind of thing?

src/app/pages/search/listings/listings-view.jsx
```diff
@@ -32,7 +32,7 @@ import {
 import { eventTrack } from 'services/analytics.service';
 import { getStatusIcon } from 'services/listing.service';
 import { getDisplayDateFormat } from 'services/date-util';
-import { useSessionStorage as useStorage } from 'services/storage.service';
+import { useSessionStorage as useStorage, useLocalStorage } from 'services/storage.service'; // Keep the useLocalStorage import
```
andlar: What's this comment for? Actually, just generally, comments should be added when they are necessary to explain something about the code that's not immediately obvious from the line. If they're added to re-explain something obvious, then they tend to just become obsolete, or noise

src/app/pages/search/listings/listings.jsx
```diff
@@ -144,7 +146,18 @@ function ChplListingsPage() {
...
+    if (value.value === 'Previously Compared') {
+      return `${value.value} (${getPreviouslyCompared().length})`;
+    }
+    if (value.value === 'Previously Viewed') {
+      return `${value.value} (${getPreviouslyViewed().length})`;
+    }
+    if (value.value === 'Favorites') {
+      return `${value.value} (${favorites.length})`; // Display count of favorites
+    }
+    return value.value;
```
andlar: This should be a `switch` statement. If all you're doing is checking on a string value, and then returning something based on that, that's what `switch`es are meant for

andlar (changes requested):
Generally, this is good work, but I have some style / formatting / consistency concerns. Happy to talk about them if needed

src/app/pages/listing/listing.jsx
mattystank: This is a one-time adjustment to make the "Favorite" text appear smoother. I made this decision because we're not using a snack bar, which would normally include a in-box animation.

andlar (dismissed):
I'm okay with this going forward, as it looks functionally correct. There are a couple of comments that I would like to see addressed, and a lot of actual code comments that are generally unnecessary, but we can revisit that before it's merged

kekey1 (dismissed):

kekey1 (dismissed):

src/app/pages/listing/listing.jsx
```diff
@@ -4,22 +4,26 @@ import {
...
+  Fade,
+  IconButton,
   Typography,
   makeStyles,
 } from '@material-ui/core';
-import CloudDownloadIcon from '@material-ui/icons/CloudDownload';
-import EditIcon from '@material-ui/icons/Edit';
+import {
+  Star, StarOutline, Edit, CloudDownload,
+} from '@material-ui/icons';
```
andlar: If this line can't be all on one line, and I'm assuming it can't due to our linter, then the icons should be spread out on multiple lines, and alphabetized. The advantage of that "spread" is that when/if we need to add/remove/change one of them (for whatever reason) it's a single thing on a single line that changes

src/app/pages/listing/listing.jsx
```diff
@@ -4,22 +4,26 @@ import {
...
 import { number, oneOfType, string } from 'prop-types';
 
-import ChplListingHistory from './history/listing-history';
```
andlar, line 14: Why not keep the relative import? The ListingHistory component isn't used anywhere else, which is why it's not in the `components` directory, and is instead locally related to this file

src/app/pages/search/listings/listings-view.jsx
```diff
@@ -134,15 +134,17 @@ function ChplListingsView() {
   const { analytics } = useAnalyticsContext();
   const [directReviewsAvailable, setDirectReviewsAvailable] = useState(true);
   const [listings, setListings] = useState([]);
+  const [setFilteredListings] = useState([]);
```
andlar: Did you test this? The `useState` function returns an array of two things, and the first one is the actual object that holds the state, while the second is the function to update that state. I don't understand how this could work

src/app/pages/search/listings/listings-view.test.jsx
```diff
@@ -22,11 +22,15 @@ const mockContext = {
...
   hasSearched: true,
   queryString: jest.fn(() => 'queryString'),
+  filters: { quickFilters: 'Favorites' },
 };
 
 const mockApi = {
   isLoading: true,
-  data: {},
+  data: {
+    results: [], // Mock empty results to avoid undefined errors
```
andlar: Is this comment necessary?

src/app/pages/search/listings/listings.jsx
```diff
@@ -49,6 +50,7 @@ function ChplListingsPage() {
   const { getPreviouslyCompared, getPreviouslyViewed } = useContext(BrowserContext);
   const { analytics } = useAnalyticsContext();
   const [filters, setFilters] = useState(staticFilters);
+  const [favorites] = useLocalStorage('favorites', []); // Access favorites from local storage
```
andlar: Is this necessary?

src/app/pages/search/listings/listings.jsx
```diff
@@ -66,7 +68,7 @@ function ChplListingsPage() {
         getValueDisplay,
         getLongValueDisplay: getValueDisplay,
       }));
-  }, [getPreviouslyCompared, getPreviouslyViewed]);
+  }, [getPreviouslyCompared, getPreviouslyViewed, favorites]); // Add favorites to dependencies
```
andlar: Is this necessary?

src/app/pages/listing/listing.jsx
mattystank: Wanted to double check if this what your looking for 

image

andlar: That is what I wanted for the icons, but the `ChplListingHistory` component should be in a separate section, above the section it's in currently

andlar (approved):

kekey1 (approved):

PR #1466, opened 2025-08-06
OCD-4956 Developer Dashboard

mattystank:
Create baseline setup for compliance dashboards
Create Developer dashboard with skeletons and reusable data components


src/app/components/util/avatar.jsx
```diff
@@ -8,9 +8,9 @@ import { string } from 'prop-types';
 const colors = ['color_bdc3c7', 'color_6f7b87', 'color_2c3e50', 'color_2f3193', 'color_662d91', 'color_922790', 'color_ec2176', 'color_ed1c24', 'color_f36622', 'color_f8941e', 'color_fab70f', 'color_fdde00', 'color_d1d219', 'color_8ec73f', 'color_00a650', 'color_00aa9c', 'color_00adef', 'color_0081cd', 'color_005bab'];
 
 const useStyles = makeStyles((theme) => ({
-  color_bdc3c7: { backgroundColor: '#bdc3c7', color: theme.palette.getContrastText('#bdc3c7') },
-  color_6f7b87: { backgroundColor: '#6f7b87', color: theme.palette.getContrastText('#6f7b87') },
-  color_2c3e50: { backgroundColor: '#2c3e50', color: theme.palette.getContrastText('#2c3e50') },
+  color_bdc3c7: { backgroundColor: '#9efff5ff', color: theme.palette.getContrastText('#bdc3c7') },
+  color_6f7b87: { backgroundColor: '#490404ff', color: theme.palette.getContrastText('#6f7b87') },
+  color_2c3e50: { backgroundColor: '#00666dff', color: theme.palette.getContrastText('#2c3e50') },
```
andlar: I don't understand this. It's possible the naming convention is terrible, but at least it was consistent to the backgroundColor. Now the function to getContrastText isn't getting contrast text against the named color, but instead against some other color. What's going on here?

src/app/navigation/navigation-top.html
```diff
@@ -91,7 +91,7 @@
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc', 'chpl-cms-staff'])"><a ui-sref="administration.cms"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.cms')"></i> CMS</a></li>
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc', 'chpl-onc-acb'])"><a ui-sref="administration.change-requests"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.change-requests')"></i> Change Requests</a></li>
               <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc', 'chpl-onc-acb'])"><a ui-sref="administration.system-maintenance"><i class="fa fa-caret-right" ng-if="vm.isActive('administration.system-maintenance')"></i> System Maintenance</a></li>
-              <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc'])"><a ui-sref="dashboard"><i class="fa fa-caret-right" ng-if="vm.isActive('sys-admin.dashboard')"></i>Dashboard</a></li>
+              <li ng-if="vm.hasAnyRole(['chpl-admin', 'chpl-onc'])"><a ui-sref="reports-dashboard"><i class="fa fa-caret-right" ng-if="vm.isActive('sys-admin.dashboard')"></i>Dashboard</a></li>
```
andlar: Does the dashboard navigation item get the "isActive" class? I'm not sure if it would have before, but the "vm.isActive" call matches the "ui-sref" label in most of the other navigation elements. As long as it shows the "isActive" indicator then that's probably fine, but if it doesn't it'll need an update

src/app/pages/compliance-dashboard/developer-dashboard.jsx
```diff
@@ -0,0 +1,1165 @@
...
+    });
+
+    setSurveillanceActivities(surveillanceData);
+  }, [developer]);
+
+  if (isLoading || insightsLoading || usersLoading || directReviewsLoading || rwtPlansLoading || rwtResultsLoading || attestationsLoading) {
+    return <ChplDashboardSkeleton />;
+  }
+
+  const filteredUsers = displayedUsers.filter((userData) => userData.fullName?.toLowerCase().includes(localSearchTerm.toLowerCase())
```
andlar: This should be a function. Something like `getFilteredUsers` that does the same filtering, but setting it as a const here implies it doesn't change, which is probably wrong if you're actually doing a search/filter

src/app/pages/compliance-dashboard/developer-dashboard.jsx
```diff
@@ -0,0 +1,1165 @@
...
+                        ) : (
+                          <Typography variant="body2" className={classes.textItalic}>
+                            No attestation data available
+                          </Typography>
+                        )}
+
+                        <Typography className={classes.textSpacingTop} variant="body2">
+                          For more information, please visit the
+                          {' '}
+                          <a href="https://www.astp.hhs.gov/sites/default/files/2022-08/Attestations-Condition-Resource-Guide.pdf" target="_blank" rel="noopener noreferrer">Attestations Resource Guide</a>
```
andlar: Should be a `ChplLink` element so we can track analytics

src/app/pages/compliance-dashboard/developer-dashboard.jsx
```diff
@@ -0,0 +1,1165 @@
...
+                            <Typography variant="body2" color="textSecondary">
+                              (
+                              {rwtPlans?.length || 0}
+                              {' '}
+                              Found)
+                            </Typography>
+                          </Box>
+                          {rwtPlans && rwtPlans.length > 0 ? (
+                            rwtPlans.slice(0, 3).map((plan, index) => (
+                              <a
```
andlar: Also a ChplLink

src/app/pages/resources/resources.module.js
```diff
@@ -1,6 +1,7 @@
 import { reactToAngularComponent } from '../../services/angular-react-helper';
 
 import ChplCmsLookupWrapper from './cms-lookup/cms-lookup-wrapper';
+import ChplComplianceDashboardWrapper from '../compliance-dashboard/compliance-dashboard-wrapper';
```
andlar: I thought we were trying to go away from using the term "compliance dashboard". I know this isn't visible to the end user, but it makes it harder for the team to track what file is what when the term isn't used on the page anywhere

andlar (changes requested):

src/app/pages/dashboard/developer-dashboard.jsx
```diff
@@ -592,8 +592,11 @@ function ChplDeveloperDashboard() {
                         <Typography className={classes.textSpacingTop} variant="body2">
                           For more information, please visit the
                           {' '}
-                          <a href="https://www.astp.hhs.gov/sites/default/files/2022-08/Attestations-Condition-Resource-Guide.pdf" target="_blank" rel="noopener noreferrer">Attestations Resource Guide</a>
-                          .
+                          <ChplLink
+                            href="https://www.astp.hhs.gov/sites/default/files/2022-08/Attestations-Condition-Resource-Guide.pdf"
+                            text="Attestations Resource Guide"
+                            external
```
andlar: This is probably incorrect. The default is that `external` is true; it's only when we don't want it to be external that we need to put something like `external="false"`, and in this case, I think it should be counted as an internal link, as we're connecting to another hhs.gov site

andlar (approved):

PR #1486, opened 2026-01-07
OCD-5103 System Maintenance Layout UI

mattystank:


src/app/components/system-maintenance/accessibility-standard/accessibility-standards-view.jsx
```diff
@@ -79,7 +79,7 @@ function ChplAccessibilityStandardsView(props) {
             onTableSort={handleTableSort}
             orderBy={orderBy}
             order={order}
-            stickyHeader
+            stickyHeader={props.stickyHeader}
```
andlar: Two things:

1) I see you're reading a props value here, and in a bunch of other files, but I can't find where that value is being passed in via props. Am I missing something?

2) Reading props directly (like `props.stickyHeader`) is not our convention. Our convention is to resolve the props and deconstruct the arguments to the component function with things like `function ChplApiKeysView({ apiKeys: initialApiKeys, dispatch }) {`

src/app/components/system-maintenance/api-key/api-keys-view.jsx
```diff
@@ -39,15 +39,16 @@ const useStyles = makeStyles({
...
+function ChplApiKeysView(props) {
+  const { dispatch } = props;
   const [apiKeys, setApiKeys] = useState([]);
   const [order, setOrder] = useState('desc');
   const [orderBy, setOrderBy] = useState('lastUsedDate', true);
   const classes = useStyles();
 
   useEffect(() => {
-    setApiKeys(initialApiKeys.sort(sortComparator('lastUsedDate')));
-  }, [initialApiKeys]);
```
andlar, line 50: We use the convention of using the props value, renamed as `initalSomething`, in order to avoid re-rendering the component every time a parent component updates the source data from the automatic network refreshing

src/app/components/system-maintenance/api-key/api-keys-view.jsx
```diff
@@ -66,14 +67,14 @@ function ChplApiKeysView({ apiKeys: initialApiKeys, dispatch }) {
         />
       </div>
       <Paper className={classes.container}>
-        <TableContainer>
+      <TableContainer className={classes.container} component={Paper} style={{ maxHeight: 'calc(100vh - 300px)', overflow: 'auto' }}>
```
andlar: Indentation seems off here

src/app/components/system-maintenance/functionality-tested/functionalities-tested-view.jsx
```diff
@@ -2,18 +2,21 @@ import React, { useContext, useEffect, useState } from 'react';
...
+  MenuItem,
   Typography,
   makeStyles,
 } from '@material-ui/core';
 import { arrayOf, func } from 'prop-types';
 import AddIcon from '@material-ui/icons/Add';
 import EditOutlinedIcon from '@material-ui/icons/EditOutlined';
+import ArrowUpwardIcon from '@material-ui/icons/ArrowUpward';
+import ArrowDownwardIcon from '@material-ui/icons/ArrowDownward';
+import SortIcon from '@material-ui/icons/Sort';
```
andlar: These should be alphabetized

src/app/components/system-maintenance/functionality-tested/functionalities-tested-view.jsx
```diff
@@ -23,50 +26,39 @@ import {
...
-  { text: 'Action', invisible: true },
-];
-
 const useStyles = makeStyles({
   ...utilStyles,
-  tableResultsHeaderContainer:{
-      display: 'flex',
-      justifyContent: 'flex-end',
-  }
+  tableResultsHeaderContainer: {
```
andlar: The table is gone, right? If so, the class shouldn't reference a thing that doesn't exist

src/app/components/system-maintenance/functionality-tested/functionalities-tested-view.jsx
```diff
@@ -23,50 +26,39 @@ import {
...
   const [functionalitiesTested, setFunctionalitiesTested] = useState([]);
   const [order, setOrder] = useState('asc');
   const [orderBy, setOrderBy] = useState('value');
+  const [sortMenuAnchor, setSortMenuAnchor] = useState(null);
   const filterContext = useFilterContext();
   const classes = useStyles();
 
   useEffect(() => {
-    setFunctionalitiesTested(initialFunctionalitiesTested
+    setFunctionalitiesTested(propsFunctionalitiesTested
```
andlar: The convention has been to use `initialSomething` here

src/app/components/system-maintenance/functionality-tested/functionalities-tested-view.jsx
```diff
@@ -121,72 +168,80 @@ function ChplFunctionalitiesTestedView({ dispatch, functionalitiesTested: initia
...
+                  </Box>
+                </Grid>
+                <Grid container spacing={2} alignItems="center">
+                  <Grid item xs={2} sm={2}>
+                    <Typography variant="caption" display="block" color="primary">Regulatory Text Citation</Typography>
+                    <Typography variant="body2">{item.regulatoryTextCitation || 'N/A'}</Typography>
+                  </Grid>
+                  <Grid item xs={2} sm={2}>
+                    <Typography variant="caption" display="block" color="primary">Rule</Typography>
+                    <Typography variant="body2">{item.rule?.name ?? 'N/A'}</Typography>
```
andlar: There's a mix of `||` and `??` to handle blank values. I can't remember what we use as a convention, but we definitely shouldn't be using two different versions in the same file

src/app/components/system-maintenance/functionality-tested/functionalities-tested-view.jsx
```diff
@@ -121,72 +168,80 @@ function ChplFunctionalitiesTestedView({ dispatch, functionalitiesTested: initia
...
+                    <Typography variant="caption" display="block" color="primary">End Date</Typography>
+                    <Typography variant="body2">{getDisplayDateFormat(item.endDay) || 'N/A'}</Typography>
+                  </Grid>
+                  <Grid item xs={6} sm={2}>
+                    <Typography variant="caption" display="block" color="primary">Required Date</Typography>
+                    <Typography variant="body2">{getDisplayDateFormat(item.requiredDay) || 'N/A'}</Typography>
+                  </Grid>
+                  <Grid item xs={6} sm={2}>
+                    <Typography variant="caption" display="block" color="primary">Extension End Date</Typography>
+                    <Typography variant="body2">{getDisplayDateFormat(item.extensionEndDay) || 'N/A'}</Typography>
```
andlar: I'm pretty sure the `getDisplayDateFormat` function has a fallback that returns "N/A" or something like it if the date isn't present. If that's the case, then the `|| 'N/A'` part of this will never execute

src/app/pages/administration/system-maintenance/system-maintenance.jsx
```diff
@@ -114,109 +124,118 @@ const useStyles = makeStyles({
...
   },
-  
+  toggleButton: {
+    maxWidth: 'min-content',
+    padding: '4px',
+  },
 });
 
 const maintenanceItems = [{
+  id: '',
```
andlar: ID should have a value, even if it's a made up one

src/app/pages/administration/system-maintenance/system-maintenance.jsx
```diff
@@ -330,14 +354,25 @@ function ChplSystemMaintenance() {
   return (
     <AnalyticsContext.Provider value={data}>
       <div className={classes.container}>
-        <div className={classes.navigation}>
+        <div className={`${classes.navigation} ${navOpen ? classes.navOpen : classes.navClosed}`}>
           <Card className={classes.navigationFlex}>
-            { maintenanceItems
+            <ChplToolTip title={navOpen ? 'Collapse Navigation' : 'Expand Navigation'}>
+              <Button
+                onClick={() => setNavOpen(!navOpen)}
```
andlar: When using the `setState` returned setting functions, you should use the form of `setNavOpen((previousValue) => "change the previous value")`. Otherwise, in this case, someone doubleclicking this button would set the value to be false and then false, or true and then true, if they click before React updates the state. Using the (p) => change p form queues up the changes so they take place in order

src/app/pages/administration/system-maintenance/system-maintenance.jsx
```diff
@@ -361,24 +396,24 @@ function ChplSystemMaintenance() {
...
-          { active === 'optionalStandards' && <ChplOptionalStandards /> }
-          { active === 'qmsStandards' && <ChplQmsStandards /> }
-          { active === 'standards' && <ChplStandards /> }
-          { active === 'subscriptions' && <ChplManageSubscriptions /> }
-          { active === 'svaps' && <ChplSvaps /> }
-          { active === 'systemJobs' && <ChplSystemJobs /> }
-          { active === 'testData' && <ChplTestData /> }
-          { active === 'testTools' && <ChplTestTools /> }
-          { active === 'ucdProcesses' && <ChplUcdProcesses /> }
+          { active === 'accessibilityStandards' && <ChplAccessibilityStandards onEditModeEnter={closeNavigation} /> }
```
andlar: What is `onEditModeEnter`? I don't see this anywhere else in the PR, and don't think it's something we already had in these components

andlar (changes requested):

src/app/components/system-maintenance/announcement/announcements.jsx
```diff
@@ -1,4 +1,5 @@
 import React, { useContext, useEffect, useState } from 'react';
+import AnnouncementOutlinedIcon from '@material-ui/icons/AnnouncementOutlined';
```
andlar: Is this import being used anywhere?

src/app/components/system-maintenance/certification-criterion/certification-criteria-view.jsx
```diff
@@ -92,14 +83,13 @@ function ChplCertificationCriteriaView({ certificationCriteria: initialCertifica
           .join('; '),
       }))
       .sort(sortComparator('value')));
-  }, [initialCertificationCriteria, filterContext.filters, filterContext.searchTerm]);
+  }, [props.certificationCriteria, filterContext.filters, filterContext.searchTerm]);
```
andlar: Why change from the convention we've been using of `initialVariableHere`?

src/app/components/system-maintenance/code-set/code-sets-view.jsx
```diff
@@ -1,129 +1,146 @@
...
-  { text: 'Applicable Criteria' },
-];
-
 const useStyles = makeStyles({
   ...utilStyles,
-  tableResultsHeaderContainer:{
-      display: 'flex',
-      justifyContent: 'flex-end',
-  }
+  tableResultsHeaderContainer: {
```
andlar: Is this class still used? If it is, it shouldn't have "table" in the name, as there aren't any tables on the page any more

src/app/components/system-maintenance/code-set/code-sets-view.jsx
```diff
@@ -1,129 +1,146 @@
...
+    setCodeSets(propsCodeSets
       .map((item) => ({
         ...item,
         criteriaDisplay: item.criteria
           .sort(sortCriteria)
           .map((c) => `${c.status === 'REMOVED' ? 'Removed | ' : ''}${c.number}`)
           .join(', '),
       })));
-  }, [initialCodeSets]);
+  }, [propsCodeSets]);
```
andlar: Why change the name?

src/app/components/util/chpl-search-result-card.jsx
```diff
@@ -0,0 +1,133 @@
...
+          <Grid
+            key={group.map((f) => f.label).join('-')} // eslint-disable-line react/no-array-index-key
+            container
+            spacing={2}
+            alignItems="flex-start"
+          >
+            {group.map((field) => ( // eslint-disable-line
+              <Grid
+                key={field.label} // eslint-disable-line react/no-array-index-key
+                item
```
andlar: What are all the eslint-disable commands for? I think the array-index-key ones are not valid, and I don't know what the other two do

src/app/components/util/chpl-search-result-card.jsx
```diff
@@ -0,0 +1,133 @@
...
+        sm: number,
+        style: shape({}),
+        iconButton: node,
+      }),
+    ),
+  ),
+  actions: node,
+};
+
+ChplSearchResultCard.defaultProps = {
```
andlar: With the upgrade to React we're going to have to get rid of defaultProps entirely. You could do that now, or wait until the React upgrade ticket is merged, but this will need to change before it can be merged

src/app/pages/administration/system-maintenance/system-maintenance-wrapper.jsx
```diff
@@ -1,16 +1,43 @@
...
+  },
+});
 
 function ChplSystemMaintenanceWrapper() {
+  const classes = useStyles();
+
   return (
     <AppWrapper>
-      <BreadcrumbWrapper title="System Maintenance">
+      <Box className={classes.pageHeader}>
```
andlar: All of these changes should be in the system-maintenance.jsx file. The wrapper files are only around to bridge angularJS and React, and all of them are destined to be deleted, so should not have any styling / functionality / etc.

andlar (approved):

PR #1501, opened 2026-04-17
OCD-5234 Code Set Indicator

mattystank:
This pull request enhances the display of code set information in the `ChplCriterionDetailsView` component, especially for users without admin roles. It introduces a visual indicator for code set compliance, improves the code set status calculation, and refines the UI for different user roles.

**Code Set Status Display Improvements:**

* Added a visual status indicator using `Chip` components with "Up-to-date" and "Not up-to-date" labels, including icons and explanatory text, for non-admin users viewing code set requirements.
* Implemented the `getCodeSetStatus` function to determine if the listing's code sets are current with requirements, factoring in required dates and applicable code sets.
* Applied custom styles for the code set status indicators to visually distinguish up-to-date and not-up-to-date states.

**Role-Based UI Adjustments:**

* Refactored the code set display so that only admin roles see the detailed list of code sets, while other users see the new status indicator and message. [[1]](diffhunk://#diff-88c3971258571ddd1d73fb8c7f5f73be13f7c977e45fed5efb8e27d4bbfb4375L131-R171) [[2]](diffhunk://#diff-88c3971258571ddd1d73fb8c7f5f73be13f7c977e45fed5efb8e27d4bbfb4375R185-R186)

**Dependency and Utility Updates:**

* Added necessary imports for new Material-UI components, icons, and utility functions to support the updated UI and status logic. [[1]](diffhunk://#diff-88c3971258571ddd1d73fb8c7f5f73be13f7c977e45fed5efb8e27d4bbfb4375R5) [[2]](diffhunk://#diff-88c3971258571ddd1d73fb8c7f5f73be13f7c977e45fed5efb8e27d4bbfb4375R20-R26)

These changes improve the clarity and user experience around code set compliance for different types of users.[#OCD-5234]

### DIFF
--- a/src/app/components/listing/details/criteria/criterion-details-view.jsx
+++ b/src/app/components/listing/details/criteria/criterion-details-view.jsx
@@ -2,6 +2,7 @@ import React, { useContext } from 'react';
 import {
   Box,
   Card,
+  Chip,
   IconButton,
   List,
   ListItem,
@@ -16,10 +17,13 @@ import {
   makeStyles,
 } from '@material-ui/core';
 import InfoIcon from '@material-ui/icons/Info';
+import RadioButtonCheckedIcon from '@material-ui/icons/RadioButtonChecked';
+import RadioButtonUncheckedIcon from '@material-ui/icons/RadioButtonUnchecked';
 import { arrayOf } from 'prop-types';
 
 import ChplReliedUponSoftwareView from './relied-upon-software/relied-upon-software-view';
 
+import { useFetchCodeSets } from 'api/standards';
 import {
   ChplEllipsis,
   ChplLink,
@@ -32,9 +36,24 @@ import {
   certificationResult,
   qmsStandard,
 } from 'shared/prop-types';
+import { getDisplayDateFormat } from 'services/date-util';
 import { palette } from 'themes';
 
 const useStyles = makeStyles({
+  upToDate: {
+    backgroundColor: palette.active,
+    color: palette.white,
+    '& .MuiChip-icon': {
+      color: palette.white,
+    },
+  },
+  notUpToDate: {
+    backgroundColor: palette.error,
+    color: palette.white,
+    '& .MuiChip-icon': {
+      color: palette.white,
+    },
+  },
   infoIcon: {
     float: 'right',
   },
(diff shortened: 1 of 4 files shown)


src/app/components/listing/details/criteria/criterion-details-view.jsx
```diff
@@ -57,8 +76,29 @@ function ChplCriterionDetailsView({
 }) {
   const { hasAnyRole, user } = useContext(UserContext);
   const { listing } = useContext(ListingContext);
+  const { data: allCodeSets } = useFetchCodeSets();
```
andlar: Should also use the `isError` and `isLoading` elements to either leave a blank space or a processing indicator while the network call is being made

src/app/components/listing/details/criteria/criterion-details-view.jsx
```diff
@@ -142,14 +182,64 @@ function ChplCriterionDetailsView({
...
+                            )}
+                            { codeSetStatus.status === 'not-up-to-date' && (
+                              <>
+                                <Chip
+                                  icon={<RadioButtonUncheckedIcon />}
+                                  label="Not up-to-date"
+                                  className={classes.notUpToDate}
+                                />
+                                <Typography variant="body2">
+                                  {'This listing\u2019s code set does not meet the requirement that took effect'}
```
andlar: We generally use the HTML encoding instead of the `\u` encodings. My guess is this should be `&apos;` instead of `\u2019\`

src/app/components/listing/details/criteria/criterion-details-view.jsx
```diff
@@ -142,14 +182,64 @@ function ChplCriterionDetailsView({
...
+                            { criterion.codeSets.map((cs) => (
+                              <ListItem key={cs.id}>
+                                { cs.codeSet.name }
+                              </ListItem>
+                            ))}
+                          </List>
+                          { criterion.codeSets?.length === 0 && 'None' }
+                        </>
+                      )}
+                      { !hasAnyRole(['chpl-admin', 'chpl-onc', 'chpl-onc-acb']) && (() => {
```
andlar: I would extract both of these out of the JSX portion of the component and make a function called something like `getDisplayValue` that has this same logic. We usually try to more complicated logic out of the "html-ish structure" stuff so it can be reasoned about a little more cleanly

src/app/components/listing/details/criteria/criterion-details-view.jsx
```diff
@@ -57,8 +76,29 @@ function ChplCriterionDetailsView({
 }) {
   const { hasAnyRole, user } = useContext(UserContext);
   const { listing } = useContext(ListingContext);
+  const { data: allCodeSets } = useFetchCodeSets();
   const classes = useStyles();
 
+  const getCodeSetStatus = () => {
+    if (!allCodeSets || !criterion.codeSets) return null;
+    const criterionId = criterion.criterion.id;
+    const today = new Date().toISOString().split('T')[0];
```
andlar: I think we need to use `moment` or `LocalDay` for this. Our mandate has been to base all date values on Eastern Time, and with this code I believe there'd be a window of time where the code set is not expired yet in ET, but expired in GMT, and so might show the tag/chip/text/whatever incorrectly. I could be wrong, but it's worth looking into

andlar (dismissed):

PR #1503, opened 2026-04-27
OCD-4307: remove flaky tests after UI updates

ashwinimore:


andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

andlar (dismissed):

test/specs/inactive-certificates-search.readonly.e2e.js
```diff
@@ -6,6 +6,7 @@ describe('the Inactive Certificates Search page', () => {
   beforeEach(async () => {
     page = new InactiveCertificatesSearchPage();
     await page.open();
+    await page.isTableVisible();
```
andlar: If this check is useful for the inactive-certificates page, then it would probably be useful for _all_ of the search pages that are like this, and if that's the case, then the check ought to be part of the base pageobject's `await page.open()` command

ashwinimore: I wanted to test this on the failed test in the AQA plan to see if it helps resolve the issue. I'll update it so that it can be used across all search pages.

andlar (approved):

PR #1514, opened 2026-06-16
OCD-5298: remove a variety of AngularJS pieces

andlar:


kekey1 (approved):

mattystank (approved):

src/app/api/axios.jsx
```diff
@@ -62,14 +64,34 @@ function AxiosProvider({ children }) {
       return updated;
     });
 
-    ax.interceptors.response.use((response) => response,
+    ax.interceptors.response.use(
```
mattystank, line 67: Possible stale closure / hook dependency

enqueueSnackbar and setLoginWidgetState are used inside useMemo. Should these be included in the dependency array (or otherwise stabilized) to avoid stale references if context changes?

src/app/api/axios.jsx
```diff
@@ -62,14 +64,34 @@ function AxiosProvider({ children }) {
...
+        }
+        if (response.headers['cache-cleared']) {
+          enqueueSnackbar('Update processing. Your changes may not be reflected immediately in the search results and shortcuts pages. Please contact CHPL admin if you have any concerns', {
+            variant: 'warning',
+          });
+        }
+        return response;
+      },
       (error) => {
         if (error.response.data && error.response.data === 'Invalid authentication token.' && authService.hasAnyRole(['chpl-admin', 'chpl-onc', 'chpl-onc-acb', 'chpl-cms-staff', 'chpl-developer'])) {
```
mattystank, line 88: Could we guard this with optional chaining (error?.response?.data) before access? In network failures, error.response can be undefined.

src/app/components/api-key/api-key-confirm.jsx
```diff
@@ -21,66 +19,64 @@ const useStyles = makeStyles(() => ({
   centeredGridItem: {
     gridColumn: '2 / 4',
   },
-}));
+});
 
-function ChplApiKeyConfirm(props) {
-  const networkService = getAngularService('networkService');
+function ChplApiKeyConfirm({ hash }) {
```
mattystank, line 24: Should hash and mutate be included in the useEffect dependency array? As written, this only runs once and may violate hook dependency expectations.

src/app/components/cms-widget/cms-wrapper.jsx
```diff
@@ -1,58 +1,40 @@
...
 function CmsWrapper({ children }) {
-  const $localStorage = getAngularService('$localStorage');
-  const $rootScope = getAngularService('$rootScope');
-  const [listings, setListings] = useState([]);
+  const [isOpen, setIsOpen] = useState(false);
+  const [listings, setListings] = useLocalStorage('cms', []);
   const queryClient = useQueryClient();
 
-  useEffect(() => {
-    setListings($localStorage?.cmsWidget?.products ?? []);
```
mattystank, line 15: Previous implementation deduped by id when adding. Should we preserve that to prevent duplicate entries in local storage and widget UI?

src/app/components/compare-widget/compare-wrapper.jsx
```diff
@@ -1,49 +1,33 @@
...
-import { getAngularService } from 'services/angular-react-helper';
+import { useLocalStorage } from 'services/storage.service';
 import { CompareContext } from 'shared/contexts';
 
 function CompareWrapper({ children }) {
-  const $localStorage = getAngularService('$localStorage');
-  const $rootScope = getAngularService('$rootScope');
-  const [listings, setListings] = useState([]);
-
-  useEffect(() => {
```
mattystank, line 12: Similar question as CMS widget: should this add path dedupe existing listing.id to avoid duplicates after repeated add actions?

src/app/pages/listing/listing-edit.jsx
```diff
@@ -140,8 +135,6 @@ function ChplListingEditPage() {
     analytics: {
       ...analytics,
       category: 'Edit Listing',
-      label: listing.chplProductNumber,
```
mattystank, line 143: We removed label / aggregationName from edit analytics. Is this intentional from a reporting standpoint, or should we preserve those fields for dashboard continuity?

mattystank (changes requested):

src/app/api/axios.jsx
andlar: The dependency array is empty to indicate that this useMemo should only run once. At that time, it will set up whatever is going on in this memo, and that won't change going forward anyway

src/app/api/axios.jsx
andlar: The first check makes sure there's a value; the second checks the value. If the first one fails due to a chaining issue it doesn't matter, because we wouldn't want to execute the `true` path anyway

src/app/components/api-key/api-key-confirm.jsx
andlar: Why would we try to run the mutate twice? It's a POST or a PUT, and there's no way trying to do a second one would be a good idea

src/app/pages/listing/listing-edit.jsx
andlar: We didn't remove it. It's being provided by a higher order analytics provider

src/app/components/cms-widget/cms-wrapper.jsx
andlar: There's no way to have a state where the button to add a Listing to this widget (or the other one) can exist while the button is enabled to "add"

mattystank (approved):

PR #1527, opened 2026-07-31
OCD-5395 - Reduce build dependency on Bamboo artifacts

tmy1313:


docker/apache2/sites-available/000-default.conf
```diff
@@ -0,0 +1,57 @@
+<VirtualHost *:80>
+	# The ServerName directive sets the request scheme, hostname and port that
+	# the server uses to identify itself. This is used when creating
+	# redirection URLs. In the context of virtual hosts, the ServerName
+	# specifies what hostname must appear in the request's Host: header to
+	# match this virtual host. For the default virtual host (this file) this
+	# value is not decisive as it is used as a last resort host regardless.
+	# However, you must set it for any further virtual host explicitly.
+
+	ServerName chpl.ahrqdev.org
```
andlar, line 10: If we're committing this, for all time, we should probably use the names that are actually correct. AHRQ has been gone for a while, and while it's probably not hurt anything, this would be a good time to update it

src/app/api/api-key.js
```diff
@@ -0,0 +1,10 @@
+const HARDCODED_API_KEY = '12909a978483dfb8ecd0596c98ae9094';
```
andlar: Is this something that can be externalized?

docker/apache2/sites-available/000-default.conf
andlar: Perhaps make the server name (and/or ServerAdmin below) a parameter as well

docker/apache2/apache2.conf
```diff
@@ -0,0 +1,230 @@
...
+# see README.Debian for details.
+
+# Include generic snippets of statements
+IncludeOptional conf-enabled/*.conf
+
+# Include the virtual host configurations:
+IncludeOptional sites-enabled/*.conf
+
+# vim: syntax=apache ts=4 sw=4 sts=4 sr noet
+ServerName 127.0.0.1
```
kekey1, line 230: I don't know where this comment belongs exactly, but we do get an error `AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 10.0.1.199. Set the 'ServerName' directive globally to suppress this message` whenever the development environment is built.

docker/apache2/sites-available/000-default.conf
tmy1313: Good catch on both counts — went with the parameter route, in fc08fbcf.

`ServerName`/`ServerAdmin` in the maintenance vhosts are now `${SERVER_NAME}`/`${SERVER_ADMIN}`, rendered at container start by a new `docker/docker-entrypoint-maint.sh` (defaults: `localhost` and `chpl@ainq.com`), which mirrors how the live image already templates `BACKEND_URLS`/`API_KEY`. So the stale `chpl.ahrqdev.org` is gone and the same maintenance image can serve any environment's hostname without a rebuild.

Two details worth flagging:
- `001-error.conf` had the identical stale pair, and since the image does `a2dissite 000-default.conf` / `a2ensite 001-error.conf`, that is the vhost actually in effect — both are parameterized.
- The entrypoint renders from pristine copies stashed at `/etc/apache2/conf-templates/` rather than substituting in place, so a container restart re-renders instead of finding the placeholders already consumed. `envsubst` is given an explicit variable list so Apache's own `${APACHE_LOG_DIR}` survives untouched.

I verified the rendering logic locally (defaults, override, repeat runs, `${APACHE_LOG_DIR}` preserved) but could not build the image — no Docker daemon on my machine. Also left `chpl@ainq.com` as the default admin address; say the word if that one is stale too.

docker/env-config.template.js
```diff
@@ -0,0 +1,2 @@
+window.__env = window.__env || {};
+window.__env.API_KEY = "${API_KEY}";
```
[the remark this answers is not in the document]
tmy1313: Fixed in 824c5276 � done the way you suggested.

`docker-entrypoint.sh` now builds a complete, pre-quoted JSON string literal in `API_KEY_JSON`, and the template renders it unquoted:

```js
window.__env.API_KEY = ${API_KEY_JSON};
```

`httpd:2.4-alpine` has no `jq` or `python`, so the escaping is busybox `sed` plus `awk`: backslashes first (so escapes added afterwards aren't re-escaped), then quotes, tabs and carriage returns, with embedded newlines folded to \`\\n`\ by awk.

Verified by rendering a key containing a quote, a backslash, a tab, a newline and `</script>`, then parsing the result with node � it parses cleanly and the value round-trips byte-for-byte. The unset-API_KEY branch (which deliberately emits only the namespace so `browserInfo.slice.js`'s `??` default still applies) and ordinary alphanumeric keys are both unchanged.

Worth noting for anyone reading this later: the exposure here was never user-reachable � `API_KEY` is supplied by ops at `docker run` time � so this is hardening rather than a live hole. The more likely real-world failure was a stray quote in a pasted key silently breaking the app.

.dockerignore
```diff
@@ -0,0 +1,6 @@
+.git
+node_modules
+dist
+.yarn/cache
+.yarn/install-state.gz
+.yarn/unplugged
```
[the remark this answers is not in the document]
tmy1313: Added in the latest commit — `docker/www` is now in `.dockerignore`.

Checked the reasoning and it holds: the `Stage maintenance web roots` step in `docker-publish.yml` copies `dist/` into both `docker/www/chpl` and `docker/www/error`, so it is roughly twice the build output, and the live image builds with `context: .`.

Two clarifications for the record: in CI the live and maint jobs run on separate runners, so in practice this was slowing down *local* builds rather than the pipeline; and the maintenance build uses `context: docker`, which makes Docker read `docker/.dockerignore` instead of this file — so the exclusion can't accidentally starve the maint image of the web roots it needs.

docker/apache2/mods-enabled/status.conf
```diff
@@ -0,0 +1,27 @@
+<IfModule mod_status.c>
+        # Allow server status reports generated by mod_status,
+        # with the URL of http://servername/server-status
+        # Uncomment and change the "192.0.2.0/24" to allow access from other hosts.
+
+        <Location /server-status>
+                SetHandler server-status
+                # Require local
+                Require ip 10.0.1.0/24
```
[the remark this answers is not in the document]
tmy1313: Agreed, and fixed in c43abf29 � though I went a step further than the commented-example suggestion.

Rather than hardcoding `Require local`, both `/server-status` blocks now render from `${STATUS_ALLOWED_IPS}` via the entrypoint this PR just added to the maintenance image, reusing the same variable name and the same `127.0.0.1` default the live image already uses:

```apache
<Location /server-status>
    SetHandler server-status
    Require ip ${STATUS_ALLOWED_IPS}
</Location>
```

That gets the safety property you asked for � an unconfigured build is localhost-only, never open to whatever `10.0.1.0/24` happens to mean in the environment it lands in � while keeping the range changeable at `docker run` time instead of requiring a rebuild to ever reopen it. It also means one variable and one mental model across both images; an environment passes the same value to both containers.

I also covered the duplicate `/server-status` block in `sites-available/000-default.conf`, which carried the identical hardcoded CIDR. That vhost is disabled in this image (`a2dissite 000-default.conf`), so it was latent rather than live, but it would have been a trap for whoever re-enabled it.

Verified the rendering against the real config files: defaults give `Require ip 127.0.0.1` in both places, `STATUS_ALLOWED_IPS=10.0.1.0/24` reproduces today's behaviour, repeat runs are idempotent, and Apache's own `${APACHE_LOG_DIR}` survives untouched. I could not build the image itself � no Docker daemon on my machine.

Deployment note for whoever picks this up: DEV needs `STATUS_ALLOWED_IPS` passed to the maintenance container, not just the live one, or `/server-status` there will be localhost-only.

docker/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+RUN corepack enable
+
+# .yarnrc.yml (sets nodeLinker: node-modules) and the tracked .yarn/ subdirs must be
+# present for this first install, or it silently runs under Yarn's default PnP linker
+# instead - then `yarn build` fails later looking for a node-modules install-state
+# file that was never created.
+COPY package.json yarn.lock .yarnrc.yml ./
+COPY .yarn ./.yarn
+RUN yarn install
```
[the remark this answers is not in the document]
tmy1313: Good catch � added in the latest commit.

One detail that makes this more relevant than it looks: `.yarnrc.yml` sets `enableImmutableInstalls: false`, so Yarn's usual CI auto-detection is disabled here and the bare `yarn install` was non-immutable even in CI. The explicit CLI flag is what actually closes the gap. (That setting predates this PR � it came in with `6df1af1cd`, "ci: ensure UI AQA tests can run" � so I left it alone rather than flipping a global that something else may depend on.)

Verified rather than assumed: the lockfile at this branch's HEAD is byte-identical to `origin/development`'s, and the workflow's `build` job � which already runs `yarn install --immutable` � has succeeded on the last 8 runs there, most recently 2026-09-11. And since `build-and-push` declares `needs: [build]`, any lockfile divergence fails that gate before the image build starts, so this flag can't introduce a new CI failure mode; it only stops a local image build from quietly papering over one.

The scenario where it would bite, for the record: `checksumBehavior: update` is also set, so a registry re-publishing a tarball under a new checksum would make Yarn want to rewrite `yarn.lock`, and `--immutable` would then fail. That's already true of the `build` job today, so it isn't new exposure.

docker/apache2/apache2.conf
tmy1313: Thanks for reporting this � and your instinct that the comment was in the wrong place was right. I think I can explain why you saw it.

Line 230 here *does* set `ServerName 127.0.0.1`, but this file only ever reaches the **maintenance** image (`Dockerfile.maint` does `ADD apache2 /etc/apache2`). The live image is built from `httpd:2.4-alpine` and assembles its `httpd.conf` from the base image's stock config, where `ServerName` ships commented out � and nothing in `docker/Dockerfile` or the entrypoint set it afterwards. That's why the message named a real VPC address (`10.0.1.199`) instead of the `127.0.0.1` this file would have produced: the warning was coming from the live container, not the maintenance one.

Fixed on the live image in the latest commit, via a new `servername.conf` template rendered at container start from `SERVER_NAME` (default `localhost`) � following the existing `proxy.conf`/`status.conf` pattern, and reusing the same variable name the maintenance image now takes, so one value can be passed to both containers.

Two honest caveats: `AH00558` is a warning rather than an error � Apache starts and serves fine either way, so the real cost was log noise, which matters here mainly because this image pipes access logs to Datadog. And I verified the template rendering locally (defaults plus DEV-style values across all four variables) but could not build or boot the image to watch the warning disappear, as there's no Docker daemon on my machine. Worth a glance at the container logs on the next DEV build to confirm.

kekey1 (approved):

andlar (approved):

